PR #34, opened 2022-10-05
Require Python 3.9

### LINKED ISSUE #33: Syntax error when importing clouddrift with Python 3.8
```
$ ipython
Python 3.8.10 (default, Jun 22 2022, 20:18:18) 
Type 'copyright', 'credits' or 'license' for more information
IPython 8.5.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import clouddrift
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In [1], line 1
----> 1 import clouddrift

File ~/Work/clouddrift/venv/lib/python3.8/site-packages/clouddrift/__init__.py:1
----> 1 from clouddrift.dataformat import *

File ~/Work/clouddrift/venv/lib/python3.8/site-packages/clouddrift/dataformat.py:9
      5 from typing import Tuple, Optional
      6 from tqdm import tqdm
----> 9 class ragged_array:
     10     def __init__(
     11         self,
     12         coords: dict,
   (...)
     16         attrs_variables: Optional[dict] = {},
     17     ):
     18         self.coords = coords

File ~/Work/clouddrift/venv/lib/python3.8/site-packages/clouddrift/dataformat.py:29, in ragged_array()
     22     self.attrs_variables = attrs_variables
     23     self.validate_attributes()
     25 @classmethod
     26 def from_files(
     27     cls,
     28     indices: list,
---> 29     preprocess_func: Callable[[int], xr.Dataset],
     30     vars_coords: dict,
     31     vars_meta: list = [],
     32     vars_data: list = [],
     33     rowsize_func: Optional[Callable[[int], int]] = None,
     34 ):
     35     """Generate ragged arrays archive from a list of trajectory files
     36 
     37     Args:
   (...)
     46         obj: ragged array class object
     47     """
     48     # if no method is supplied, get the dimension from the preprocessing function

TypeError: 'ABCMeta' object is not subscriptable
```

As I understand it, the typing syntax for stuff in `collections.abc` was not introduced until Python 3.9.

We can remove the subscript here, or we can require Python 3.9. I recommend the latter.

milancurcic:
This introduces a minimum requirement for Python 3.9 in package manifests.

@philippemiron Do you know if `tests/environment.yml` used for anything and does that need updating as well? It's the same as the one in the root directory. I'm not as familiar with Conda.

Closes #33.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -2,7 +2,7 @@ name: clouddrift
 channels:
   - conda-forge
 dependencies:
-  - python>=3.8
+  - python>=3.9
   - matplotlib>=3.5.2
   - numpy>=1.21.6
   - pandas>=1.4.4
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,7 +11,7 @@ authors = [
 ]
 description = "Accelerating the use of Lagrangian data for atmospheric, oceanic, and climate sciences"
 readme = "README.md"
-requires-python = ">=3.8"
+requires-python = ">=3.9"
 classifiers = [
     "Programming Language :: Python :: 3",
     "License :: OSI Approved :: Apache Software License",
--- a/tests/environment.yml
+++ b/tests/environment.yml
@@ -2,7 +2,7 @@ name: clouddrift
 channels:
   - conda-forge
 dependencies:
-  - python>=3.8
+  - python>=3.9
   - matplotlib>=3.5.2
   - numpy>=1.21.6
   - pandas>=1.4.4


philippemiron:
So there are two environment files.

- `~/environment.yml` installs clouddrift from pip (eventually from conda)
- `tests/environment.yml` is used during the GitHub Actions (https://github.com/Cloud-Drift/clouddrift/actions/workflows/ci.yml) and use the latest version directly from the GitHub repo every time something gets updated.

There might be a better way to do this.

milancurcic:
OK, cool. In that case, let's leave it, and I bumped the version that one as well.

I'll set myself up with Conda soon so I can test these things locally.

PR #35, opened 2022-10-05
Organize, make more concise, and expand the install instructions

milancurcic:
This PR makes the text more concise, expands the install instructions (most people will expect to see it here), and tidies things up a little bit.

I removed the statement and links to the project owners. IMO those should be visible from the Org page and contributors list, and here it may convey a message of project ownership and could discourage contributions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ cd clouddrift
 conda env create --file environment.yml
 ```
 
-Once the library is in Beta, a conda-forge package will be available to simplify the installation process.
+In the future, a conda-forge package will be available to simplify the installation process.
 
 ### Run the tests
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -6,18 +6,60 @@
...
+
+You can also install CloudDrift using Conda:
+
+```
+git clone https://github.com/cloud-drift/clouddrift
+cd clouddrift
+conda env create --file environment.yml
+```
+
+Once the library is in Beta, a conda-forge package will be available to simplify the installation process.
```
selipot, line 41: I think the library is beyond *Beta* at this stage, isn't it? I recall the issue of not having a conda install yet is because of the the awkward library dependency. What about writing:

"In the future, a conda-forge package will be available to simplify the installation process."

selipot:
For the line about the conda package install, I suggest writing instead:

"In the future, a conda-forge package will be available to simplify the installation process."

I think the bottleneck preventing that is `awkward array`, is this correct @philippemiron ?

README.md
milancurcic: I agree, no need to label it as Beta IMO, but wasn't sure so I left it alone.

philippemiron: Sounds good! I see that the awkward version is now updated on conda (https://anaconda.org/conda-forge/awkward). I'll give it a try on Monday, but it should work now to update the environment and remove installing packages with pip.

selipot: That's great news!

selipot: I am getting confused here again: now that `awkward` can be installed with conda, we can strip the `environment.yml` file of `pip` and pip install of awkward and move the awkward install back to conda ...  but should we still specify to install clouddrift?

PR #51, opened 2022-11-11
Implement `RaggedArray.from_xarray()`

### LINKED ISSUE #44: Implement `RaggedArray.from_xarray()`
As discussed with @selipot on 10/20, it's in scope for clouddrift to allow getting a `RaggedArray` instance from an `xarray.Dataset`. 

`RaggedArray.from_netcdf()` already does this internally, i.e.:

https://github.com/Cloud-Drift/clouddrift/blob/437c0e4d8885d1ed4aeb08775b3dc41a7473f871/clouddrift/dataformat.py#L78-L97

Some assumptions are currently needed for the dimension names (currently assumed "traj" and "obs"). The method should allow the user to specify dimension names to use, in absence of an established convention.

milancurcic:
* [x] Implementation
* [x] Test
* [x] Docs

Closes #44

### DIFF
--- a/clouddrift/dataformat.py
+++ b/clouddrift/dataformat.py
@@ -69,34 +69,7 @@ class RaggedArray:
         Returns:
             obj: ragged array class object
         """
-        coords = {}
-        metadata = {}
-        data = {}
-        attrs_global = {}
-        attrs_variables = {}
-
-        with xr.open_dataset(filename) as ds:
-            nb_traj = ds.dims["traj"]
-            nb_obs = ds.dims["obs"]
-
-            attrs_global = ds.attrs
-
-            for var in ds.coords.keys():
-                coords[var] = ds[var].data
-                attrs_variables[var] = ds[var].attrs
-
-            for var in ds.data_vars.keys():
-                if len(ds[var]) == nb_traj:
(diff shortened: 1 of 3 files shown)


philippemiron (approved):
I added a test to check if the dimension names are not "traj" and "obs". Also, you can do `from clouddrift import RaggedArray` directly because `__init__.py` imports automatically the class in the `dataformat.py`.

philippemiron:
Actually, I have two more comments to discuss:
- wouldn't it be more useful also to have the default parameters on the `from_netcdf` method? 
- we could also copy/paste this wrapper and call it `from_zarr` and it should work seamlessly

milancurcic:
Thanks, Philippe.

On the first point, yes, it would be more useful, but I chose to leave it alone for now because we're considering removing readers and writers for file formats (parquet and netcdf). The rationale is that the users will work with data in xarray or awkward, and they can read and write from and to files from those libraries.

And similar for zarr. Most likely we'll just show the user how to do it from xarray. Less likely, you're right, it can be the same wrapper as from_netcdf.

Another reason that it wouldn't be part of this PR is that I want to keep PRs small and focused (single feature).

milancurcic:
Thank you for the review!

PR #57, opened 2022-12-14
Awkward v2

### LINKED ISSUE #55: Upgrade to awkward v2
Awkward v2 was released on [Dec 9](https://github.com/scikit-hep/awkward/releases/tag/v2.0.0). As we don't require `<2` for `awkward` in pyproject.toml `dependencies` field, the `awkward` import will need to change from

```python
import awkward._v2 as ak
``` 

to

```python
import awkward as ak
```

and require `awkward>=2.0.0` in pyproject.toml.

Thanks to Ibis Gonzalez for reporting.

philippemiron:
Fix #55

### DIFF
--- a/clouddrift/dataformat.py
+++ b/clouddrift/dataformat.py
@@ -1,4 +1,4 @@
-import awkward._v2 as ak
+import awkward as ak
 import xarray as xr
 import numpy as np
 from collections.abc import Callable
--- a/clouddrift/select.py
+++ b/clouddrift/select.py
@@ -1,4 +1,4 @@
-import awkward._v2 as ak
+import awkward as ak
 import numpy as np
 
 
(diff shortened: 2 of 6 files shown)


milancurcic:
Thanks. It looks like we got two same PRs at the same time (#58). Any preference on not changing the dependency version for awkward (my PR) vs changing it (your PR)?

milancurcic:
@selipot will postpone the release for tomorrow while waiting for more info.

selipot:
More info from? I do not understand the difference for the dependency so please decide.

philippemiron:
@milancurcic only thing is that with 1.9.0 then `.to_packed()` will not work.. So I'm not sure if the try/except import is that useful.

milancurcic:
Ah, if only I saw your response a minute earlier. That's a good reason to not allow 1.9+.

milancurcic:
I reverted #58 with #59 and will merge this one instead.

milancurcic:
Thanks again, @philippemiron. For future PRs, let's all please use branches in our own forks rather than in cloud-drift/clouddrift, to keep this repo clean.

PR #64, opened 2022-12-28
Velocity from positions

### LINKED ISSUE #62: Function to compute velocity from positions
Let's discuss the API for this function.

```python
def velocity_from_positions(
    x: xr.DataArray,
    y: xr.DataArray,
    time: xr.DataArray,
    coords: str='spherical', # where x is lon and y is lat, otherwise can be 'cartesian' where x is eating and y is northing
    order: int=1, # can also be 2 for centered difference; we can discuss if higher orders may be desired
) -> Tuple(xr.DataArray, xr.DataArray)
```

I wonder if we shouldn't use `coords` to avoid confusion with the Xarray special `coords`. Perhaps `coord_system` or `coordinate_system`?

Until there is native support for ragged arrays in Xarray, the user should be careful to not run this on multiple consecutive trajectories. It will be easy for this function to detect that (e.g. jump in time and space) and issue a warning.

milancurcic:
This PR implements the `velocity_from_positions` that returns Cartesian (u, v) velocity from arrays of Cartesian (easting, northing) or spherical (lon, lat) positions.

* [x] Implementation
  * [x] Haversine distance and bearing
  * [x] velocity_from_positions 
* [x] Tests:
  - [x] Haversine tests
  - [x] velocity_from_positions tests
* [x] Docstrings

I considered using the [haversine](https://github.com/mapado/haversine) package as a dependency because it implements the distance function. However, considering that it doesn't implement the Haversine bearing, I decided to not add it as a dependency and implement a subset of functionality that we need directly in this package in the `clouddrift.haversine` module.

Using the Haversine bearing to compute Cartesian dx, dy is limited because it assumes a plane rather than a sphere. For a dlon, dlat of ~0.1 degree, the accuracy is about 1e-5 rad. The accuracy decreases with dlon and dlat. Any feedback on how to improve this is very welcome.

Closes #62.

### DIFF
--- a/clouddrift/analysis.py
+++ b/clouddrift/analysis.py
@@ -1,11 +1,10 @@
-from datetime import datetime
 import numpy as np
 from typing import Optional, Tuple
 import xarray as xr
 from clouddrift.haversine import distance, bearing
 
 
-def velocity_from_positions(
+def velocity_from_position(
     x: np.ndarray,
     y: np.ndarray,
     time: np.ndarray,
--- a/clouddrift/analysis.py
+++ b/clouddrift/analysis.py
@@ -1,15 +1,16 @@
 from datetime import datetime
+import numpy as np
 from typing import Optional, Tuple
 import xarray as xr
 
 
 def velocity_from_positions(
-    x: xr.DataArray[float],
-    y: xr.DataArray[float],
-    time: xr.DataArray[datetime],
+    x: xr.DataArray,
+    y: xr.DataArray,
+    time: xr.DataArray,
     coord_system: Optional[str] = "spherical",
     difference_scheme: Optional[str] = "forward",
-) -> Tuple[xr.DataArray[float], xr.DataArray[float]]:
+) -> Tuple[xr.DataArray, xr.DataArray]:
     """Compute velocity in meters per second given arrays of positions and
     time.
 
@@ -28,7 +29,11 @@ def velocity_from_positions(
 
     Forward and backward schemes are effectively the same except that the
     position at which the velocity is evaluated is shifted one element down in
-    the backward scheme relative to the forward scheme.
+    the backward scheme relative to the forward scheme. In the case of a
+    forward or backward difference scheme, the last or first element of the
+    velocity, respectively, is extrapolated from its neighboring point. In the
+    case of a centered difference scheme, the start and end boundary points are
+    evaluated using the forward and backward difference scheme, respectively.
 
     Args:
         x (xr.DataArray[float]): An array of x-positions (longitude in degrees or easting in meters)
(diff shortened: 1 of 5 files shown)


jonathanlilly:
I used the spherical law of cosines to compute bearing in latlon2uv, see 
http://www.siranah.de/html/sail042e.htm

The relevant lines are
xx=sind(lon2-lon).*cosd(lat2);
yy=cosd(lat).*sind(lat2)-sind(lat).*cosd(lat2).*cosd(lon2-lon);
gamma=atan2(yy,xx);

This is defined relative to east so the arguments to arctan function are swapped relative to the link above.

What is the haversine formula for bearing that you refer to?

milancurcic:
Thanks @jonathanlilly, that's very helpful. I mistakenly thought it was based on the Haversine function. It turns out it's the same formula that you use, except that I didn't flip the x, y args to the atan2, so my bearing was defined as zero toward North and increasing clockwise. So rather than mangling the sines and cosines on the (u, v) side, I changed the bearing function like you did so that it has a conventional orientation.

jonathanlilly:
Hi Milan, 

Great, I'm glad that was helpful.  

You might also like the attached, from a work to be submitted in early 2023.  The equations of interest are (33--35), which are derived in Appendices B and C.

The main point is that you can compute the great circle distance using dx and dy in a tangent plane.  I haven't seen this result elsewhere, though I'm sure it's "known".   These are the same expressions that come up in computing the bearing.  Thus, using equation (35) of the attached is a numerically efficient way to compute the great circle distance, because you've already computed dx and dy.  This is shown in Appendix C to be equivalent to the haversine formula. 

I use this approach in the Matlab routine latlon2xy.

[mapping6.pdf](https://github.com/Cloud-Drift/clouddrift/files/10320236/mapping6.pdf)

philippemiron:
Is there a reason to use `xr.DataArray` inside `velocity_from_positions`? 

If it's to keep the metadata, I guess we need to include new attributes `long_name`, `units`, etc. before returning the calculated velocities. Otherwise, since this function is not using any xarray (or Awkward Array) features it might be better to keep this as regular NumPy arrays.

milancurcic:
I don't think so. I initially created them as DataArrays because we expect many users to come from Xarray datasets, but this function should work with any array-like. So, this is more of an oversight than intentional. Good catch, thanks.

selipot:
Is there anything I could do to help with this PR? I am eager to try this function on some dataset.

milancurcic:
You can play with it now--only tests are outstanding.

milancurcic:
Regarding the naming, I don't recall that we arrived at the specific name that we want yesterday on the call. Currently, it's `velocity_from_positions`, and we discussed just `velocity`. There may be a use case to have a function `velocity_from_acceleration`, in which case it would be a good idea to keep the longer and more specific name. We also briefly discussed `velocity_from_positions` vs. `positions_to_velocity`. I think the former is a tad easier to parse--the first word that a reader reads may be the most impactful and should be reflective of what the function returns.

selipot:
fine, let's go for `velocity_from_position` then! We can always rename later if needed

selipot (approved):
Love it!-

PR #67, opened 2023-01-07
Make analysis and haversine module available in the clouddrift namespace

milancurcic:


milancurcic:
I also changed the import everything from dataformat to import only `RaggedArray` to keep the user's namespace clean from all the libraries that `RaggedArray` uses internally.

PR #71, opened 2023-01-13
Allow multi-dimensional inputs to `velocity_from_position`

### LINKED ISSUE #69: `velocity_from_position`: Handle n-d arrays
Previous discussion in #68.

This issue is to discuss whether and how should `velocity_from_position` handle n-d arrays. Some specific questions:

* Is 2-d sufficient for most use cases or is n-d also useful? The second obvious dimension here are trajectories.
* If not, what are the use cases for n-d?

milancurcic:
This PR adds the capability for `velocity_from_position` to accept inputs of arbitrary rank (but they still must be of the same shape). I added tests for 2-d and 3-d inputs.

Closes #69.

### DIFF
--- a/clouddrift/analysis.py
+++ b/clouddrift/analysis.py
@@ -15,8 +15,11 @@ def velocity_from_position(
     time.
 
     x and y can be provided as longitude and latitude in degrees if
-    coord_system == "spherical" (default), or as northing and easting in meters
+    coord_system == "spherical" (default), or as Easting and Northing in meters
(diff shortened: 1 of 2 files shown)


selipot:
For this and future functions that are dimension-aware, could we make it more flexible for the user by passing arguments as needed `time_axis = n_1`, `zonal_axis = n_2`, `meridional_axis=n_3`, etc? I think that would be useful.

Issue #72, opened 2023-01-14
Units of time and the result in `velocity_from_position`


milancurcic:
In https://github.com/Cloud-Drift/clouddrift/issues/62#issuecomment-1382388478 @selipot wrote:

> I think the function returns the velocity in meter per unit of time input, correct? If yes we need to update the help. Can the function be made aware of the unit of the time and automatically return the velocity in m/s? with an option for that behavior to be disabled?

milancurcic:
You're right, I had assumed that the time is in seconds but haven't documented it. But as you point out, the user can pass the time array in any unit and get the velocity back in m / time unit of the input. The docstring should certainly be fixed to address this.

> Can the function be made aware of the unit of the time and automatically return the velocity in m/s? with an option for that behavior to be disabled?

It could but I don't think we should do it. It would add significant complexity for us for something that doesn't seem to me as much benefit (it's very easy to divide or multiply by e.g. 3600 or 86400). I recommend that we clearly document the behavior for example:

> The resultant velocity is in units of meters per unit of time passed as input. For example, if the input time is in the units of seconds, the returned velocity is in the units of meters per second; if the input time is in the units of days, the returned velocity is in the units of meters per days.

PR #74, opened 2023-01-14
Rework the docstring to address the units of measure

milancurcic:
Partially addresses #72.


selipot (changes requested):
I edited the docstring with my own commit but it looks like a lint test subsequently failed??

milancurcic:
I just ran black on it to format it. It seems like the linter didn't like the trailing whitespace. Kind of annoying.

Thanks for the edit, I think it's better worded now. Are more changes needed before the merge?

selipot:
I think it's good to go.

PR #76, opened 2023-01-16
Allow specifying `time_axis` in `velocity_from_position`

### LINKED ISSUE #73: Allow specifying the time axis for multi-dimensional input arrays to `velocity_from_position`
In https://github.com/Cloud-Drift/clouddrift/pull/71#issuecomment-1381932552 @selipot wrote:

> For this and future functions that are dimension-aware, could we make it more flexible for the user by passing arguments as needed time_axis = n_1?

milancurcic:
This PR implements #73.

It took some trial and error to get right. Multidimensional stuff getting reshaped and reordered can give me a headache, but I think the current implementation works, based on one test with 3-d input arrays where the time axis is the one in the middle, and based on my playing interactively with a few GLAD trajectories.

The implementation is, in a nutshell:

1. `time_axis` is -1 (last) if not present
2. If `time_axis` is present and different from -1 or `size(x.shape) - 1`, transpose the input arrays (into new copies, of course) so that the new time axis is last; the rest of the algorithm is the same as before
3. If inputs were transposed, transpose back so that the output velocities have the same shape as inputs.

With this PR merged, let's release 0.5.0 as long as we don't discover other problems with the function in the meantime.

Closes #73.

### DIFF
--- a/clouddrift/analysis.py
+++ b/clouddrift/analysis.py
@@ -72,19 +72,18 @@ def velocity_from_position(
             f"time_axis ({time_axis}) is outside of the valid range ([-1, {len(x.shape) - 1}])."
         )
 
-    # If time_axis is not the last one, reshape the inputs
+    # Nominal order of axes on input, i.e. (0, 1, 2, ..., N-1)
+    target_axes = list(range(len(x.shape)))
+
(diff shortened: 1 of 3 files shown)


selipot (approved):
works good, was able to test.

PR #88, opened 2023-01-27
GDP-hourly adapter

### LINKED ISSUE #60: GDP hourly dataset adapter
Part of #53.

Incorporate from clouddrift-examples/data/gdp.py into clouddrift/adapters.py.

milancurcic:
This adapts Philippe's helper module clouddrift-examples/data/gdp.py into the core library as `clouddrift.adapters.gdp`. Most or all of it is meant to be internal to clouddrift, i.e. not part of the user-facing API, but can be of course used as a reference on how to ingest custom datasets into clouddrift.

The end goal is to provide to the user a convenience `from_gdp` function that returns a ragged xarray Dataset. 

**One obvious question is whether the end-user ever need to load ragged array xarray Dataset from raw GDP (individual NetCDF files), or is the ragged-array GDP NetCDF on AWS always sufficient?**

If the ragged-array GDP on AWS is all the end-users would need, then all this probably doesn't belong here, and could be included in an auxiliary repo like `clouddrift-tools` or `clouddrift-dev` or similar. And the `from_gdp` method would simply point to the AWS URL for convenience, and the user can take it from there.

However, if there's still a need for non-ragged (raw) GDP NetCDF access to the end-users, then I think this gdp adapter module is appropriate.

This module doesn't deviate much from Philippe's original in clouddrift-examples. Most changes are housekeeping, like removing the global state, tidy up docstrings, and using a system-provided temporary directory for downloads instead of the user's filesystem (i.e. `../data`).

Implementation:

* [x] Remove global state
* [x] Use system-provided temporary dir for downloads
* [ ] ~~`from_gdp1h() -> xr.Dataset` convenience wrapper~~
 
Closes #60.

### DIFF
--- a/clouddrift/adapters/gdp.py
+++ b/clouddrift/adapters/gdp.py
@@ -7,8 +7,6 @@ import concurrent.futures
 import re
 from tqdm import tqdm
 import os
-from os.path import isfile, join, exists
-from dataclasses import dataclass
 import warnings
 
 
@@ -26,7 +24,7 @@ def parse_directory_file(filename: str) -> pd.DataFrame:
     """
     aoml_dirfl_url = "https://www.aoml.noaa.gov/ftp/pub/phod/buoydata/"
 
-    df = pd.read_csv(join(aoml_dirfl_url, filename), delimiter="\s+", header=None)
+    df = pd.read_csv(os.path.join(aoml_dirfl_url, filename), delimiter="\s+", header=None)
     # combine the date and time columns to easily parse dates below
     df[4] += " " + df[5]
     df[8] += " " + df[9]
@@ -57,7 +55,7 @@ file_pattern = "drifter_{id}.nc"
 
 # create subdirectory
 folder = "../data/raw/gdp-v2.00/"
-os.makedirs(folder, exist_ok=exists(folder))  # create raw data folder
+os.makedirs(folder, exist_ok=os.path.exists(folder))  # create raw data folder
 
 # directory files
 dirfl_names = [
@@ -86,7 +84,7 @@ def fetch_netcdf(url, file):
     """
     Download and save file from the given url (if not present)
     """
-    if not isfile(file):
+    if not os.path.isfile(file):
         req = urllib.request.urlretrieve(url, file)
 
 
@@ -122,8 +120,8 @@ def download(drifter_ids: list = None, n_random_id: int = None):
         files = []
         for i in drifter_ids:
             file = file_pattern.format(id=i)
-            urls.append(join(aoml_https_url, file))
-            files.append(join(folder, file))
+            urls.append(os.path.join(aoml_https_url, file))
+            files.append(os.path.join(folder, file))
 
         # parallel retrieving of individual netCDF files
         list(
@@ -204,7 +202,7 @@ def drogue_presence(lost_time, time):
 
 def rowsize(index: int) -> int:
     return xr.open_dataset(
-        join(folder, file_pattern.format(id=index)),
+        os.path.join(folder, file_pattern.format(id=index)),
         decode_cf=False,
         decode_times=False,
         concat_characters=False,
@@ -222,7 +220,7 @@ def preprocess(index: int) -> xr.Dataset:
     :return: xr.Dataset containing the data and attributes
     """
     ds = xr.load_dataset(
-        join(folder, file_pattern.format(id=index)),
+        os.path.join(folder, file_pattern.format(id=index)),
         decode_times=False,
         decode_coords=False,
     )
--- a/clouddrift/adapters/gdp.py
+++ b/clouddrift/adapters/gdp.py
@@ -5,6 +5,7 @@ import xarray as xr
 import urllib.request
 import concurrent.futures
 import re
+import tempfile
 from tqdm import tqdm
 import os
 import warnings
@@ -12,7 +13,7 @@ import warnings
 GDP_VERSION = "2.00"
 GDP_DATA_URL = "https://www.aoml.noaa.gov/ftp/pub/phod/lumpkin/hourly/v2.00/netcdf/"
 GDP_FILENAME_PATTERN = "drifter_{id}.nc"
-
+GDP_TMP_PATH = os.path.join(tempfile.gettempdir(), "clouddrift", "gdp")
 
 def parse_directory_file(filename: str) -> pd.DataFrame:
     """Read a directory file which contains metadata of drifters' releases.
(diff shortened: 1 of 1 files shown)


selipot:
Here we need to distinguish two cases: 
- the hourly dataset for which individual netcdf files (let's not call them raw) are available on the AOML FTP server and a file of ragged arrays is available from NOAA NCEI and AWS
- the 6-hourly dataset for which individual netcdf files are available on the AOML FTP server and a file of ragged arrays is only available from http/ftp from the AOML GDP website

I think I see the use of this adapter (adaptor?) for implementing `from_gdp()` but will this be for both datasets? Should we think of both `from_gdp1h` and `from_gdp6h`? and in both cases could we pass a list of ids as an argument? In that case the access to individual files via the FTP server is better? To be discussed.

milancurcic:
Yes, we'd have both `from_gdp1h` and `from_gdp6h`. This PR is for 1h, and the 6h one will reuse most of the internals here.

I think passing ids makes sense if we pull individual files from ftp so that we download only the requested files.. If we pull from ragged array NetCDF, then the lazy downloading can be leveraged by the NetCDF library via xarray. In this case I wouldn't pass ids at this stage.

selipot (approved):
I approve th

philippemiron:
I'm not sure how many people are going to create themselves the ragged array from individual NetCDF archives for the GDP, but I still think it's a good idea to have this centralized in the Clouddrift package.

I'll take a look later and merge it.

clouddrift/adapters/gdp.py
```diff
@@ -0,0 +1,548 @@
...
+import re
+import tempfile
+from tqdm import tqdm
+import os
+import warnings
+
+GDP_VERSION = "2.00"
+GDP_DATA_URL = "https://www.aoml.noaa.gov/ftp/pub/phod/lumpkin/hourly/v2.00/netcdf/"
+GDP_FILENAME_PATTERN = "drifter_{id}.nc"
+GDP_TMP_PATH = os.path.join(tempfile.gettempdir(), "clouddrift", "gdp")
```
philippemiron, line 16: Using a temporary directory means you have to download the files every time you want to recreate a subset ragged array. Maybe this could be an argument and have the default path a temporary one?

milancurcic: Good idea, I agree

selipot: Yes, I second this.

clouddrift/adapters/gdp.py
```diff
@@ -0,0 +1,548 @@
...
+        idx [list]: List of drifters to include in the ragged array
+    Returns:
+        idx [list]: Sorted list of drifters
+    """
+    return df.ID[np.where(np.in1d(df.ID, idx))[0]].values
+
+
+def fetch_netcdf(url: str, file: str):
+    """Download and save the file from the given url, if not already downloaded."""
+    if not os.path.isfile(file):
```
philippemiron, line 89: This only looks if the file exists in the directory. Ideally, that would also need to compare if the server's version is newer.

So locally, you can get the file creation/modification with `os.path`:
```
# file creation
c_timestamp = os.path.getctime(path)

# file modification
timestamp = os.path.getmtime(path)
```
but I'm not sure if there is a way to do this with `urllib`.

milancurcic: Yes, if we make a HEAD request, one of the headers returns the last modified time. Something like

```python
req = urllib.request.Request(url, method='HEAD')
resp = urllib.request.urlopen(req)
last_modified_string = resp.headers.get('Last-modified')
```

(didn't test)

The same HEAD request will also return the `Content-length` header which can be used to compare with the file size on disk and determine if the file was interrupted.

This will significantly slow down the download because we're adding an additional request per file (the files are small so the batch download is latency-bound rather than bandwidth-bound). So if we do it, I'd make it optional with an argument.

milancurcic: Great. We may as well just switch completely to `requests` instead of `urllib` as it's already a dependency and has an overall simpler API.

philippemiron: Agree

philippemiron: FYI: I did a small test. For ten files, it takes ~3.5s to retrieve the headers and compare the times. And it takes ~11s to download.

milancurcic: That's pretty good, IMO, worth it having the check on. In favor of keeping it simple, I change my mind and suggest just always doing this until somebody complains. We can add an option to disable the check then.

milancurcic:
I still have a few things to address here and will merge when ready.

clouddrift/adapters/gdp.py
milancurcic: Opened #94 to track this.

clouddrift/adapters/gdp.py
milancurcic: Opened #95 to track this.

milancurcic:
@selipot and I agreed yesterday that the user-facing `from_gdp1h` should get the ragged array data from AWS, rather than processing individual files.

This can thus be cleanly separated from this PR and I track it in #96.

It may still be useful to have the custom GDP pre-processing as part of the core lib, so we'll merge this here.

philippemiron:
I missed the merge of this today, I don't really want to create an issue for this but could you modify the preprocess function by, removing this:
```
# set coordinates
ds = ds.set_coords(["ids", "longitude", "lon360", "latitude", "time"])
```
since it's not needed anymore and we can leave whatever coordinates they set. And include this instead to rename the longitude and latitude.
```
# rename variables
ds = ds.rename_vars({"longitude": "lon", "latitude": "lat"})
```

philippemiron:
> @selipot and I agreed yesterday that the user-facing `from_gdp1h` should get the ragged array data from AWS, rather than processing individual files.
> 
> This can thus be cleanly separated from this PR and I track it in #96.
> 
> It may still be useful to have the custom GDP pre-processing as part of the core lib, so we'll merge this here.

Aren't we (or someone at NOAA) going to need the `gdp_1h` processing functions that read the individual NetCDF files to create the ragged array? I do think it should be included in the adapters too.

selipot:
Definitively, for both hourly and 6-hourly! So should it be with `from_gdp1h` or in  the`clouddrift-example` repo?

philippemiron:
I would suggest to create the adapters for the 1-hourly so we have both there, and used them in the cloudrift-examples. 

Then, we can also have another example using the AWS file.

milancurcic:
@philippemiron can you clarify whether you mean defining the `from_gdp*` functions in the core library or in the examples?

IMO the user-facing function should be in the core lib. The developer function (individual files -> ragged NetCDF) can be in either core lib, auxiliary repo like clouddrift-dev, or clouddrift-examples, though the last one seems least appropriate.

milancurcic:
For the developer function, my preference is to define it here, perhaps `clouddrift.adapters.gdp.to_raggedarray`. Then every custom adapter module could have a consistently named `to_raggedarray` function that developers could use to generate new versions of the ragged array NetCDF, and contributors/users could use as a template to contribute new custom adapters to the library.

philippemiron:
> For the developer function, my preference is to define it here, perhaps `clouddrift.adapters.gdp.to_raggedarray`. Then every custom adapter module could have a consistently named `to_raggedarray` function that developers could use to generate new versions of the ragged array NetCDF, and contributors/users could use as a template to contribute new custom adapters to the library.

This sounds perfect to me.

PR #89, opened 2023-01-29
pypi actions

### LINKED ISSUE #85: create github actions to update pypi package
- [x] create secret in the CloudDrift account for pypitest and pypi API tokens
- [x] create a rule to build (this should work):
```
on:
  release:
    types:
      - published
  push:
    tags:
      - 'v*'
```
that triggers on new release, if the tags starts with v.
- [x] send to pypi test
- [x] send to pypi
- [x] validate that the package is properly upload to pypi on a new version released

Reference (https://github.com/OceanParcels/parcels/blob/master/.github/workflows/pypi-release.yaml)

philippemiron:
This resolves #85. 

PS: I didn't use my fork because it requires secrets that are stored in the Clouddrift settings.

### DIFF
--- a/.github/workflows/pypi.yml
+++ b/.github/workflows/pypi.yml
@@ -1,0 +1,97 @@
+name: pypi
+on:
+  release:
+    types:
+      - published
+  push:
+    tags:
+      - 'v*'
+  workflow_dispatch:
+
+jobs:
+  build-artifacts:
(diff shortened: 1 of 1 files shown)


philippemiron:
Looks like this has to be merged to test. This is one of the annoying things with GitHub actions.

Issue #82, opened 2023-01-22
RaggedArray to accept any type of coordinates and variables


selipot:
The function `RaggedArray` does seem to expect lat, lon, time as coordinates. It should be more flexible and allow ad hoc coordinates.

PR #92, opened 2023-01-29
Refactor coordinates

philippemiron:
Implemented #82 and what was discussed here https://github.com/Cloud-Drift/clouddrift-examples/issues/18.

The main API difference is with the RaggedArray.from_files().
```
def from_files(
        cls,
        indices: list,
        preprocess_func: Callable[[int], xr.Dataset],
        **vars_coords: list**,
        vars_meta: Optional[list] = [],
        vars_data: Optional[list] = [],
        rowsize_func: Optional[Callable[[int], int]] = None,
    ):
```
Above are the new parameters. The coordinates are now passed as a simple list. The pros and cons of this are:
- \+ more streamline with the other arguments, `coords`, `metadata`, `data` are now all lists.
- \+ a user can easily add/remove variables to the coordinates list. They don't have to be set as coordinates in `xarray` by the preprocessing functions anymore.
- \+ not assuming any number of coordinate variables.
- \- we can't rename the coordinates at this point. 
- \+ we can't rename the coordinates at this point (😀); I now think that might have been a bit confusing. Now, if you want to modify anything in the original dataset (like the coordinate names, attributes, etc.), it is all done in the `preprocess_func`.

As I said, we are not assuming that the coordinates in a previously constructed ragged array are `["ids", "lon", "lat", "time"]`. For the functions related to `xarray` (`from_xarray()`, `from_netcdf()`), this has no effect since any variables identified as a coordinate will be found in `ds.coords.keys()` and handled properly. However, there is no concept of *coordinates* in a parquet file, so this requires adding a `vars_coords` parameter to properly retrieve all coordinate variables in Awkward related functions (`from_awkward` and `from_parquet()`).

PS: I can take care to update `clouddrift-examples` afterward.

### DIFF
--- a/clouddrift/dataformat.py
+++ b/clouddrift/dataformat.py
@@ -27,7 +27,7 @@ class RaggedArray:
     def from_awkward(
         cls,
         array: ak.Array,
-        name_coords: Optional[list] = ["time", "lon", "lat", "ids"],
+        vars_coords: Optional[list] = ["time", "lon", "lat", "ids"],
     ):
         """Load a RaggedArray instance from an Awkward Array.
 
@@ -112,7 +112,7 @@ class RaggedArray:
 
     @classmethod
     def from_parquet(
-        cls, filename: str, name_coords: Optional[list] = ["time", "lon", "lat", "ids"]
+        cls, filename: str, vars_coords: Optional[list] = ["time", "lon", "lat", "ids"]
     ):
         """Read a ragged arrays archive from a parquet file
 
@@ -123,7 +123,7 @@ class RaggedArray:
         Returns:
             obj: ragged array class object
         """
-        return cls.from_awkward(ak.from_parquet(filename), name_coords)
+        return cls.from_awkward(ak.from_parquet(filename), vars_coords)
 
     @classmethod
     def from_xarray(cls, ds: xr.Dataset, dim_traj: str = "traj", dim_obs: str = "obs"):
--- a/clouddrift/dataformat.py
+++ b/clouddrift/dataformat.py
@@ -194,6 +194,17 @@ class RaggedArray:
     def attributes(
         ds: xr.Dataset, name_coords: list, name_meta: list, name_data: list
     ) -> Tuple[dict, dict]:
+        """Returns the global attributes and the attributes of all variables (name_coords, name_meta, and name_data) from a xr.Dataset
+
+        Args:
+            ds (xr.Dataset): _description_
+            name_coords (list): Name of the coordinate variables to include in the archive
+            name_meta (list): Name of metadata variables to include in the archive (Defaults to [])
+            name_data (list): Name of the data variables to include in the archive (Defaults to [])
+
+        Returns:
+            Tuple[dict, dict]: the global and variables attributes
+        """
         attrs_global = ds.attrs
 
         # coordinates, metadata, and data
--- a/clouddrift/dataformat.py
+++ b/clouddrift/dataformat.py
@@ -199,8 +199,8 @@ class RaggedArray:
         attrs_variables = {}
 
         # coordinates
-        for var in vars_coords.keys():
-            attrs_variables[var] = ds[vars_coords[var]].attrs
+        for var in vars_coords:
+            attrs_variables[var] = ds[var].attrs
 
         # metadata and data
         for var in vars_meta + vars_data:
@@ -213,7 +213,7 @@ class RaggedArray:
         preprocess_func: Callable[[int], xr.Dataset],
         indices: list,
         rowsize: list,
-        vars_coords: dict,
+        vars_coords: list,
         vars_meta: list,
         vars_data: list,
     ) -> Tuple[dict, dict, dict]:
--- a/clouddrift/dataformat.py
+++ b/clouddrift/dataformat.py
@@ -213,9 +213,9 @@ class RaggedArray:
         preprocess_func: Callable[[int], xr.Dataset],
         indices: list,
         rowsize: list,
-        vars_coords: list,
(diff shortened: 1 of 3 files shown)


selipot (dismissed):
I have tested the PR by running `cloudrift-examples/gdp.ipynb` and `gdp-6hourly.ipynb` and I am running into the following issues:

For `RaggedArray.from_files`, `var_coords` is a dictionary mapping the mandatory coordinates list to the name of those variables in the dataset. So having a key/value pair `{'lon':'longitude'}` should work because we want the coordinate `'lon'` to be the variable `'longitude'` found in the individual NetCDF files. This does not work and `{'longitude':'lon'}` works instead, which makes longitude a coordinate.

Later, `ra = RaggedArray.from_parquet('../data/process/gdp_v2.00.parquet')` does not work because the method seems to expect `lon` which it cannot find because we previously written it with `longitude` and not `lon`.

clouddrift/dataformat.py
```diff
@@ -60,20 +64,20 @@ def from_files(
         cls,
         indices: list,
         preprocess_func: Callable[[int], xr.Dataset],
```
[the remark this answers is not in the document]
philippemiron: No? It's `Callable[[int], xr.Dataset]` -> `[int]` is the argument, and the it returns a `xr.Dataset`.

philippemiron:
I think you might have missed this description at the top [here](https://github.com/Cloud-Drift/clouddrift/pull/92#issue-1561491853). Or it was really not clear.

> I have tested the PR by running `cloudrift-examples/gdp.ipynb` and `gdp-6hourly.ipynb` and I am running into the following issues:

> For `RaggedArray.from_files`, `var_coords` is a dictionary mapping the mandatory coordinates list to the name of those variables in the dataset. So having a key/value pair `{'lon':'longitude'}` should work because we want the coordinate `'lon'` to be the variable `'longitude'` found in the individual NetCDF files. This does not work and `{'longitude':'lon'}` works instead, which makes longitude a coordinate.

`var_coords` is now a list. For example, `var_coords = ["ids", "lon", "lat", "ids"]` or `var_coords = ["ids", "lon", "lon360", "lat", "ids"]`. Those variable names have to match what is return by the `preprocessing_func`. More below.

> Later, `ra = RaggedArray.from_parquet('../data/process/gdp_v2.00.parquet')` does not work because the method seems to expect `lon` which it cannot find because we previously written it with `longitude` and not `lon`.

Here, if you want to use `lon`, you have to modify the `data/gdp.py` to rename_vars at the ends. Otherwise you can use `longitude`, `latitude`, etc. Finally, this function also needs the `list` of `var_coords` as argument, `ra = RaggedArray.from_parquet('../data/process/gdp_v2.00.parquet', var_coords)`.

selipot:
Ok @philippemiron my apologies, I indeed forgot about your bulleted list above. Unless @milancurcic raises objections we should merge. And i'll take you up on your offer to modify accordingly the `clouddrift-examples` notebooks

milancurcic:
I'll review tomorrow.

clouddrift/dataformat.py
```diff
@@ -24,11 +24,16 @@ def __init__(
...
+    def from_awkward(
+        cls,
+        array: ak.Array,
+        vars_coords: Optional[list] = ["time", "lon", "lat", "ids"],
+    ):
         """Load a RaggedArray instance from an Awkward Array.
 
         Args:
             array (ak.Array): Awkward Array instance to load the data from
+            name_coords (list, optional): Names of the coordinate variables in the ragged arrays
```
milancurcic, line 36: `name_coords` or `vars_coords`? I suggest just `coords`.

clouddrift/dataformat.py
```diff
@@ -60,20 +64,20 @@ def from_files(
         cls,
         indices: list,
         preprocess_func: Callable[[int], xr.Dataset],
-        vars_coords: dict,
-        vars_meta: list = [],
-        vars_data: list = [],
+        vars_coords: list,
+        vars_meta: Optional[list] = [],
+        vars_data: Optional[list] = [],
```
milancurcic, line 69: `vars_` doesn't seem to me like it adds meaning or distinction. What do you think about just `coords`, `meta`, and `data`?

```suggestion
        coords: list,
        meta: Optional[list] = [],
        data: Optional[list] = [],
```

clouddrift/dataformat.py
```diff
@@ -107,16 +111,19 @@ def from_netcdf(cls, filename: str):
...
     @classmethod
-    def from_parquet(cls, filename: str):
+    def from_parquet(
+        cls, filename: str, vars_coords: Optional[list] = ["time", "lon", "lat", "ids"]
+    ):
         """Read a ragged arrays archive from a parquet file
 
         Args:
             filename (str): filename of parquet archive
+            name_coords (list, optional): Names of the coordinate variables in the ragged arrays
```
milancurcic, line 121: As above, should be `vars_coords` if not just `coords`.

milancurcic (approved):
Thanks for this PR. I only left a few suggestions for argument naming.

clouddrift/dataformat.py
philippemiron: But there is already a `coords` in RaggedArray which is a dictionary containing the data, so I think it would be better to avoid having the same variables.

My choice would be to rename the `vars_` -> `name_`, but I'm open to suggestions.

milancurcic: Ah OK, that's fine then. In that case let's just make the docstring consistent with the arguments list.

milancurcic:
Looks great, please merge when ready.

PR #98, opened 2023-02-01
remove extra environment file

philippemiron:
I believe the main `environment.yaml` file initially installed the clouddrift pip package and the `tests/environment.yaml` one directly from the Github. Now they are both installing from the repository (?).

At the same time, now that we have a conda package, there is no need to create an environment other than to test the development release, so let's keep only the main file.


milancurcic:
Sounds good to me--did you mean to close this without merging?

philippemiron:
Yes I closed because of a main question:

Do we want the doc to be based on the pip/conda package? or we want it based on the latest code in the repo? Depending on that, we might need a different env file.

milancurcic:
I assume most users will install from a PyPI or Conda release, so I think the docs should mirror that rather than the main. I think that means that we should modify the docs CI to publish new docs only on release.

philippemiron:
Yes, and then we can use the same environment file for everything.

Issue #97, opened 2023-02-01
lint


philippemiron:
Ok, I was really confused by this linting error.

The files are fine, but there is a circular import... See for example when you run:

```
$ black haversine.py dataformat.py
All done! ✨ 🍰 ✨
2 files left unchanged.
(research) pmiron@m2air ~/Downloads/clouddrift/clouddrift
$ Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/Users/pmiron/micromamba/envs/research/lib/python3.10/multiprocessing/__init__.py", line 16, in <module>
    from . import context
  File "/Users/pmiron/micromamba/envs/research/lib/python3.10/multiprocessing/context.py", line 6, in <module>
    from . import reduction
  File "/Users/pmiron/micromamba/envs/research/lib/python3.10/multiprocessing/reduction.py", line 16, in <module>
    import socket
  File "/Users/pmiron/micromamba/envs/research/lib/python3.10/socket.py", line 54, in <module>
    import os, sys, io, selectors
  File "/Users/pmiron/micromamba/envs/research/lib/python3.10/selectors.py", line 12, in <module>
    import select
  File "/Users/pmiron/Downloads/clouddrift/clouddrift/select.py", line 1, in <module>
    import awkward as ak
  File "/Users/pmiron/micromamba/envs/research/lib/python3.10/site-packages/awkward/__init__.py", line 7, in <module>
    import awkward._nplikes
  File "/Users/pmiron/micromamba/envs/research/lib/python3.10/site-packages/awkward/_nplikes.py", line 7, in <module>
    import numpy
  File "/Users/pmiron/micromamba/envs/research/lib/python3.10/site-packages/numpy/__init__.py", line 140, in <module>
    from . import core
  File "/Users/pmiron/micromamba/envs/research/lib/python3.10/site-packages/numpy/core/__init__.py", line 100, in <module>
    from . import _add_newdocs_scalars
  File "/Users/pmiron/micromamba/envs/research/lib/python3.10/site-packages/numpy/core/_add_newdocs_scalars.py", line 9, in <module>
    import platform
  File "/Users/pmiron/micromamba/envs/research/lib/python3.10/platform.py", line 119, in <module>
    import subprocess
  File "/Users/pmiron/micromamba/envs/research/lib/python3.10/subprocess.py", line 223, in <module>
    _PopenSelector = selectors.SelectSelector
[...]

PR #100, opened 2023-02-01
Rename vars instead of setting coords in GDP preproc

milancurcic:
Addresses https://github.com/Cloud-Drift/clouddrift/pull/88#issuecomment-1411421409

### DIFF
--- a/clouddrift/adapters/gdp.py
+++ b/clouddrift/adapters/gdp.py
@@ -542,7 +542,7 @@ def preprocess(index: int) -> xr.Dataset:
         ds[var].attrs = vars_attrs[var]
     ds.attrs = attrs
 
-    # set coordinates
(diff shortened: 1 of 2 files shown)


milancurcic:
@philippemiron do you understand why black in CI thinks the analysis.py file should be reformatted? My local black is happy with it. I updated to the latest version. Sometimes this lint job is more annoying than helpful.

philippemiron:
Yes, I understood after 1h yesterday.. check #97.

milancurcic:
It seems to me like this is different. In this PR:

* My local black is happy with the files as they are;
* The CI black wants to format clouddrift/analysis.py (based on the log it seems like it wants to remove empty lines).

In #97, black seems to fail after being happy to not format dataformat.py and haversine.py?

philippemiron:
I get the same error with any combinations of at least 2 files. But you are right, the output message is not the same on the Github actions.

PR #101, opened 2023-02-01
`gdp.to_raggedarray`

milancurcic:
Implement a function `gdp.to_raggedarray` that is a convenience wrapper of executable code from the GDP hourly example.

`clouddrift.adapters.gdp.to_raggedarray()` can now be used to generate the ragged array NetCDF from hourly individual files.

Also bumped version to 0.5.0.

Not adding this to the Sphinx docs as its not intended for users. Is that reasonable? We can add this module to the API docs if you think it would be useful.

### DIFF
--- a/clouddrift/adapters/gdp.py
+++ b/clouddrift/adapters/gdp.py
@@ -17,6 +17,72 @@ GDP_DATA_URL = "https://www.aoml.noaa.gov/ftp/pub/phod/lumpkin/hourly/v2.00/netc
 GDP_FILENAME_PATTERN = "drifter_{id}.nc"
 GDP_TMP_PATH = os.path.join(tempfile.gettempdir(), "clouddrift", "gdp")
 
+GDP_COORDS = [
+    "ids",
+    "time",
+]
+GDP_METADATA = [
+    "ID",
+    "rowsize",
+    "WMO",
+    "expno",
+    "deploy_date",
+    "deploy_lat",
+    "deploy_lon",
+    "end_date",
+    "end_lat",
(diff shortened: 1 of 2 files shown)


selipot:
Could this type of function systematically include an example usage in the help? is that common/good practice?

milancurcic:
I generally like examples in docstrings. I think it's a matter of style. For example, here's an examples section for `help(np.diff)`:

```
    Examples
    --------
    >>> x = np.array([1, 2, 4, 7, 0])
    >>> np.diff(x)
    array([ 1,  2,  3, -7])
    >>> np.diff(x, n=2)
    array([  1,   1, -10])
    
    >>> x = np.array([[1, 3, 6, 10], [0, 5, 6, 8]])
    >>> np.diff(x)
    array([[2, 3, 4],
           [5, 1, 2]])
    >>> np.diff(x, axis=0)
    array([[-1,  2,  0, -2]])
    
    >>> x = np.arange('1066-10-13', '1066-10-16', dtype=np.datetime64)
    >>> np.diff(x)
    array([1, 1], dtype='timedelta64[D]')
```

Xarray also includes examples in docstrings for some functions, though not all.

I'm in favor of it.

selipot:
Yes, that is great. Let's incorporate those systematically, if possible.

milancurcic:
Cool. I'll add examples for this function in this PR, and examples for the rest of the library in a separate PR.

milancurcic:
Will add examples to the docstrings in a separate PR and merge this as is. Thanks all.

PR #110, opened 2023-02-08
Update docs.yml

philippemiron:
I saw that this ran two times during the release of v0.5.0; one time on the commit with a tag v* and another time on the package release. I suggest we update the doc only once when we published the package.


milancurcic (approved):
Good catch, thanks!

PR #111, opened 2023-02-09
Enable GDP experimental

milancurcic:
This PR modifies a few things to allow processing GDP experimental data. To do this, I:

* Added an optional `url` parameter to `gdp.to_raggedarray` to allow passing the experimental URL; it otherwise defaults to stable;
* In `gdp.preprocess`, handle missing variables and issue warnings for variables that are missing
* In `RaggedArray.from_files`, `preprocess` and `rowsize` funcs now have a second parameter `filename_pattern`, which was previously a hardcoded global variable in the gdp module. This is now decoupled.  I'm not super happy about this approach, but the alternative is to make a separate module for experimental feed, which to me seems worse.

How to use it:

```python
from clouddrift.adapters.gdp import to_raggedarray, GDP_DATA_URL_EXPERIMENTAL
ra = to_raggedarray(n_random_id=100) # GDP_DATA_URL by default
ra_exp = to_raggedarray(n_random_id=100, url=GDP_DATA_URL_EXPERIMENTAL)
```

Example in action:

```
In [1]: from clouddrift.adapters.gdp import to_raggedarray, GDP_DATA_URL_EXPERIMENTAL

In [2]: ra = to_raggedarray(n_random_id=100) # GDP_DATA_URL by default
Downloading files: 100%|██████████████████████| 100/100 [00:20<00:00,  4.84it/s]
Retrieving the number of obs: 100%|███████████| 100/100 [00:01<00:00, 87.86it/s]
/home/milan/.local/lib/python3.10/site-packages/xarray/coding/variables.py:148: FutureWarning: elementwise comparison failed; returning scalar instead, but in the future will perform elementwise comparison
  condition |= data == fv
Filling the Ragged Array: 100%|███████████████| 100/100 [00:03<00:00, 31.75it/s]
In [3]: ra_exp = to_raggedarray(n_random_id=100, url=GDP_DATA_URL_EXPERIMENTAL)
Downloading files: 100%|██████████████████████| 100/100 [00:11<00:00,  8.43it/s]
Retrieving the number of obs: 100%|████████████| 99/99 [00:00<00:00, 110.63it/s]
/home/milan/Work/clouddrift/clouddrift/clouddrift/adapters/gdp.py:348: UserWarning: Variable sst not found; skipping.
  warnings.warn(f"Variable {var} not found; skipping.")
/home/milan/Work/clouddrift/clouddrift/clouddrift/adapters/gdp.py:348: UserWarning: Variable sst1 not found; skipping.
  warnings.warn(f"Variable {var} not found; skipping.")
/home/milan/Work/clouddrift/clouddrift/clouddrift/adapters/gdp.py:348: UserWarning: Variable sst2 not found; skipping.
  warnings.warn(f"Variable {var} not found; skipping.")
/home/milan/Work/clouddrift/clouddrift/clouddrift/adapters/gdp.py:348: UserWarning: Variable err_sst not found; skipping.
  warnings.warn(f"Variable {var} not found; skipping.")
/home/milan/Work/clouddrift/clouddrift/clouddrift/adapters/gdp.py:348: UserWarning: Variable err_sst1 not found; skipping.
  warnings.warn(f"Variable {var} not found; skipping.")
/home/milan/Work/clouddrift/clouddrift/clouddrift/adapters/gdp.py:348: UserWarning: Variable err_sst2 not found; skipping.
  warnings.warn(f"Variable {var} not found; skipping.")
...
SNIP
In [4]: ds_exp = ra_exp.to_xarray()

In [5]: ds_exp
Out[5]: 
<xarray.Dataset>
Dimensions:                (traj: 99, obs: 929945)
Coordinates:
    ids                    (obs) int64 9505793 9505793 ... 300234065662800
    time                   (obs) float64 8.243e+08 8.243e+08 ... 1.667e+09
Dimensions without coordinates: traj, obs
Data variables: (12/48)
    ID                     (traj) int64 9505793 32431 ... 300234065662800
    rowsize                (traj) int32 10566 15552 11876 ... 4762 2372 3409
    WMO                    (traj) int32 6200507 3100963 ... 6203746 5601562
    expno                  (traj) int32 9435 1195 9325 ... 21312 21321 31312
    deploy_date            (traj) float64 8.243e+08 9.607e+08 ... 1.654e+09
    deploy_lat             (traj) float64 48.22 -25.96 -40.0 ... 66.0 -18.96
    ...                     ...
    err_lat                (obs) float64 2.14e-08 6.82e-08 ... 1.131e-08
    err_lon                (obs) float64 1.129e-07 3.989e-07 ... 1.208e-08
    err_ve                 (obs) float64 0.5218 0.925 ... 0.000854 0.001058
    err_vn                 (obs) float64 0.1318 0.1657 ... 0.000854 0.001058
    gap                    (obs) float64 2.678e+03 3.37e+03 ... 1.797e+04 0.0
    drogue_status          (obs) bool True True True True ... True True True
Attributes: (12/16)
    title:             Global Drifter Program hourly drifting buoy collection
    history:           version 2.00. Metadata from dirall.dat and deplog.dat
    Conventions:       CF-1.6
    date_created:      2023-02-09T18:02:50.433369
    publisher_name:    GDP Drifter DAC
    publisher_email:   aoml.dftr@noaa.gov
    ...                ...
    contributor_name:  NOAA Global Drifter Program
    contributor_role:  Data Acquisition Center
    institution:       NOAA Atlantic Oceanographic and Meteorological Laboratory
    acknowledgement:   Elipot, Shane; Sykulski, Adam; Lumpkin, Rick; Centurio...
    summary:           Global Drifter Program hourly data
    doi:               10.25921/x46c-3620
```

### DIFF
--- a/clouddrift/adapters/gdp.py
+++ b/clouddrift/adapters/gdp.py
@@ -92,17 +92,18 @@ GDP_DATA = [
 
 
 def parse_directory_file(filename: str) -> pd.DataFrame:
-    """Read a directory file which contains metadata of drifters' releases.
+    """Read a GDP directory file that contains metadata of drifter releases.
 
-    Due to naming of these files, a manual update of the last file name after an
-    update of the dataset is needed.
+    Parameters
+    ----------
+    filename : str
+        Name of the directory file to parse.
 
-    Args:
-        filename (str): Name of the directory file
-    Returns:
-        pd.DataFrame: Sorted list of drifters
+    Returns
+    -------
+    df : pd.DataFrame
+        List of drifters from a single directory file as a pandas DataFrame.
     """
-
     GDP_DIRECTORY_FILE_URL = "https://www.aoml.noaa.gov/ftp/pub/phod/buoydata/"
     df = pd.read_csv(
         os.path.join(GDP_DIRECTORY_FILE_URL, filename), delimiter="\s+", header=None
@@ -134,8 +135,13 @@ def parse_directory_file(filename: str) -> pd.DataFrame:
 
 
 def get_gdp_metadata() -> pd.DataFrame:
-    """Download and parse GDP metadata and return it as a Pandas DataFrame."""
+    """Download and parse GDP metadata and return it as a Pandas DataFrame.
 
+    Returns
+    -------
+    df : pd.DataFrame
+        Sorted list of drifters as a pandas DataFrame.
+    """
     directory_file_pattern = "dirfl_{low}_{high}.dat"
 
     dfs = []
@@ -158,20 +164,33 @@ def get_gdp_metadata() -> pd.DataFrame:
 
 def order_by_date(df: pd.DataFrame, idx: list[int]) -> np.ndarray[int]:
     """From the previously sorted directory files DataFrame, return the drifter
-    indices sorted by their end date.
+    indices sorted by their deployment date.
 
-    Args:
-        idx [list]: List of drifters to include in the ragged array
-    Returns:
-        idx [list]: Sorted list of drifters
+    Parameters
+    ----------
+    idx : list
+        List of drifters to include in the ragged array
+
+    Returns
+    -------
+    idx : list
+        List of drifters sorted by their end date.
     """
     return df.ID[np.where(np.in1d(df.ID, idx))[0]].values
 
 
 def fetch_netcdf(url: str, file: str):
-    """Download and save the file from the given url, if not already downloaded."""
+    """Download and save the file from the given url, if not already downloaded.
+
+    Parameters
+    ----------
+    url : str
+        URL from which to download the file.
+    file : str
+        Name of the file to save.
+    """
     if not os.path.isfile(file):
-        req = urllib.request.urlretrieve(url, file)
+        urllib.request.urlretrieve(url, file)
 
 
 def download(
@@ -179,13 +198,19 @@ def download(
 ):
     """Download individual NetCDF files from the AOML server.
 
-    Args:
-        drifter_ids [list]: list of drifter to retrieve (Default: all)
-        n_random_id [int]: randomly select n drifter NetCDF files
-        url [str]: URL from which to download the data (Default: GDP_DATA_URL).
-                   Alternatively, it can be GDP_DATA_URL_EXPERIMENTAL.
-    Returns:
-        out [list]: list of retrived drifters
+    Parameters
+    ----------
+    drifter_ids : list
+        List of drifter to retrieve (Default: all)
+    n_random_id : int
+        Randomly select n_random_id drifter IDs to download (Default: None)
+    url : str
+        URL from which to download the data (Default: GDP_DATA_URL). Alternatively, it can be GDP_DATA_URL_EXPERIMENTAL.
+
+    Returns
+    -------
+    out : list
+        List of retrived drifters
     """
 
     # Create a temporary directory if doesn't already exists.
@@ -241,14 +266,22 @@ def download(
 
 
 def decode_date(t):
-    """The date format is specified in 'seconds since 1970-01-01 00:00:00' but
+    """The date format is specified as 'seconds since 1970-01-01 00:00:00' but
     the missing values are stored as -1e+34 which is not supported by the
     default parsing mechanism in xarray.
 
     This function returns replaced the missing value by NaN and returns a
     datetime instance.
-    :param t: date
-    :return: datetime
+
+    Parameters
+    ----------
+    t : array
+        Array of time values
+
+    Returns
+    -------
+    out : datetime
+        Datetime instance with the missing value replaced by NaN
     """
     nat_index = np.logical_or(np.isclose(t, -1e34), np.isnan(t))
     t[nat_index] = np.nan
@@ -258,6 +291,13 @@ def decode_date(t):
 def fill_values(var, default=np.nan):
     """Change fill values (-1e+34, inf, -inf) in var array to the value
     specified by default.
+
+    Parameters
+    ----------
+    var : array
+        Array to fill
+    default : float
+        Default value to use for fill values
     """
     missing_value = np.logical_or(np.isclose(var, -1e34), ~np.isfinite(var))
     if np.any(missing_value):
@@ -265,12 +305,21 @@ def fill_values(var, default=np.nan):
     return var
 
 
-def str_to_float(value: str, default=np.nan) -> float:
+def str_to_float(value: str, default: float = np.nan) -> float:
     """Convert a string to float, while returning the value of default if the
     string is not convertible to a float, or if it's a NaN.
 
-    :param value: str
-    :return: bool
+    Parameters
+    ----------
+    value : str
+        String to convert to float
+    default : float
+        Default value to return if the string is not convertible to float
+
+    Returns
+    -------
+    out : float
+        Float value of the string, or default if the string is not convertible to float.
     """
     try:
         fvalue = float(value)
@@ -285,25 +334,37 @@ def str_to_float(value: str, default=np.nan) -> float:
 def cut_str(value: str, max_length: int) -> np.chararray:
     """Cut a string to a specific length and return it as a numpy chararray.
 
-    Args:
-        value (str): String to cut
-        max_length (int): Length of the output
-    Returns:
-        out (np.chararray): String with max_length characters
+    Parameters
+    ----------
+    value : str
+        String to cut
+    max_length : int
+        Length of the output
+
+    Returns
+    -------
+    out : np.chararray
+        String with max_length characters
     """
     charar = np.chararray(1, max_length)
     charar[:max_length] = value
     return charar
 
 
-def drogue_presence(lost_time, time):
+def drogue_presence(lost_time, time) -> bool:
     """Create drogue status from the drogue lost time and the trajectory time.
 
-    Args:
-        lost_time: Timestamp of the drogue loss (or NaT)
-        time: Observation time
-    Returns:
-        out (bool): True if drogues and False otherwise
+    Parameters
+    ----------
+    lost_time
+        Timestamp of the drogue loss (or NaT)
+    time
+        Observation time
+
+    Returns
+    -------
+    out : bool
+        True if drogues and False otherwise
     """
     if pd.isnull(lost_time) or lost_time >= time[-1]:
         return np.ones_like(time, dtype="bool")
@@ -322,13 +383,21 @@ def rowsize(index: int, filename_pattern: str) -> int:
 
 
 def preprocess(index: int, filename_pattern: str) -> xr.Dataset:
-    """Extract and preprocess the Lagrangian data and attributes. This function
-    takes an identification number that can be used to: create a file or url
-    pattern or select data from a Dataframe. It then preprocess the data and
-    returns a clean xarray Dataset.
+    """Extract and preprocess the Lagrangian data and attributes.
 
-    :param index: drifter's identification number
-    :return: xr.Dataset containing the data and attributes
+    This function takes an identification number that can be used to create a
+    file or url pattern or select data from a Dataframe. It then preprocesses
+    the data and returns a clean Xarray Dataset.
+
+    Parameters
+    ----------
+    index : int
+        Drifter's identification number
+
+    Returns
+    -------
+    ds : xr.Dataset
+        Xarray Dataset containing the data and attributes
     """
     ds = xr.load_dataset(
         os.path.join(GDP_TMP_PATH, filename_pattern.format(id=index)),
@@ -673,16 +742,21 @@ def to_raggedarray(
     n_random_id: Optional[int] = None,
     url: Optional[str] = GDP_DATA_URL,
 ) -> RaggedArray:
-    """Download and process individual GDP hourly files and return a
-    RaggedArray instance with the data.
+    """Download and process individual GDP hourly files and return a RaggedArray
+    instance with the data.
 
-    Args:
-        drifter_ids [list]: list of drifters to retrieve (Default: all)
-        n_random_id [int]: randomly select n drifter NetCDF files
-    Returns:
-        out [RaggedArray]: A RaggedArray instance of the requested dataset
+    Parameters
+    ----------
+    drifter_ids : list[int], optional
+        List of drifters to retrieve (Default: all)
+    n_random_id : list[int], optional
+        Randomly select n_random_id drifter NetCDF files
+
+    Returns
+    -------
+    out : RaggedArray
+        A RaggedArray instance of the requested dataset
     """
-
     ids = download(drifter_ids, n_random_id, url)
 
     if url == GDP_DATA_URL:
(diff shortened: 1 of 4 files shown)


philippemiron:
What do you think if we change `preprocess` and `rowsize` API to `(int, **kwargs)` instead of the proposed `(int, filename_pattern)`? 

I believe it would make it a bit more flexible to be able to pass any needed keyword arguments later on.

clouddrift/dataformat.py
```diff
@@ -210,14 +226,18 @@ def attributes(
...
+            if var in ds.keys():
+                attrs_variables[var] = ds[var].attrs
+            else:
+                warnings.warn(f"Variable {var} requested but not found; skipping.")
 
         return attrs_global, attrs_variables
 
     @staticmethod
     def allocate(
         preprocess_func: Callable[[int], xr.Dataset],
```
philippemiron, line 238: function definition

clouddrift/dataformat.py
```diff
@@ -228,6 +248,7 @@ def allocate(
         Args:
             preprocess_func (Callable[[int], xr.Dataset]): returns a processed xarray Dataset from an identification number
```
philippemiron, line 249: function definition

clouddrift/dataformat.py
```diff
@@ -68,6 +68,7 @@ def from_files(
         name_meta: Optional[list] = [],
         name_data: Optional[list] = [],
         rowsize_func: Optional[Callable[[int], int]] = None,
```
philippemiron, line 70: function definition

clouddrift/dataformat.py
```diff
@@ -168,13 +183,14 @@ def from_xarray(cls, ds: xr.Dataset, dim_traj: str = "traj", dim_obs: str = "obs
 
     @staticmethod
     def number_of_observations(
-        rowsize_func: Callable[[int], int], indices: list
+        rowsize_func: Callable[[int], int], indices: list, filename_pattern: str
     ) -> np.array:
         """Iterate through the files and evaluate the number of observations.
 
         Args:
             rowsize_func (Callable[[int], int]): returns number observations of a trajectory from its identification number
```
philippemiron, line 191: function definition

selipot:
This worked for 100 drifters and I also tried for the full dataset. The downloading steps indicates 19284 files but the final ragged array had `traj=19082`. 

Screenshot 2023-02-10 at 2 32 27 AM

What is the discrepancy?

clouddrift/adapters/gdp.py
```diff
@@ -19,7 +19,9 @@
 
 GDP_VERSION = "2.00"
```
selipot, line 20: We're moving to version 2.01 so how to make this changeable?

selipot, line 20: bump to 2.01

clouddrift/adapters/gdp.py
```diff
@@ -19,7 +19,9 @@
...
 GDP_DATA_URL = "https://www.aoml.noaa.gov/ftp/pub/phod/lumpkin/hourly/v2.00/netcdf/"
-GDP_FILENAME_PATTERN = "drifter_{id}.nc"
+GDP_DATA_URL_EXPERIMENTAL = (
```
philippemiron, line 22: I forgot where (I believe on the Pangeo JupyterHub) FTP links are blocked for security reasons. So I would suggest we switch to the https:// URL pointing to the same location: https://www.aoml.noaa.gov/ftp/pub/phod/lumpkin/hourly/experimental/.

clouddrift/adapters/gdp.py
philippemiron: > 



> This worked for 100 drifters and I also tried for the full dataset. The downloading steps indicates 19284 files but the final ragged array had `traj=19082`.
> 
> Screenshot 2023-02-10 at 2 32 27 AM
> 
> What is the discrepancy?

My only guess is if two files have different names and their IDs are the same, you would lose 1 file in the ragged array.

selipot: ``
d
d
``

selipot: I did a naive search for this issue and I can confirm that there are 19284 unique field `ID` in the 19284 files. So the issue must be from somewhere else?

selipot:
Here is an issue with this PR: for the purpose of test one might try consecutively

`ra = to_raggedarray(n_random_id=100) # GDP_DATA_URL by default`

`ra_exp = to_raggedarray(url=GDP_DATA_URL_EXPERIMENTAL)`

but this downloads files in both instances in the same temporary directory (files like drifter_XXXX.nc in the first instance and files like drifter_hourly_XXXX.nc in the second instance) which in the second case are all aggregated to form the ragged array! You can actually see that in the output from @milancurcic at the beginning of this conversation. In the experimental case his ragged array files contains only 99 traj whereas he downloaded 100 for the experimental case. Testing further but I think this is what is happening. Why don't we explicitly pass the download directory?

selipot:
Well maybe not. Redid a test after deleting the temporary folder and I still form 19082 trajectories from 19284 files so something is amiss in the code, not the files.

philippemiron:
> Well maybe not. Redid a test after deleting the temporary folder and I still form 19082 trajectories from 19284 files so something is amiss in the code, not the files.

Did you see my comment? I think it comes from the directory files which does not contains does new drifter IDs.

selipot:
@philippemiron I saw your comment about two files having potentially the same ID field? I did not understand it as meaning the directory file was missing IDs.

philippemiron:
The order by date function just creates an ordered list of ids from the directory file by the deployment date.

For example idx = [1, 5, 8, 6].

Then if you ask for d_id = [6,1,5]. It will look at the order of those drifters, and return [1,5,6]. If you say I want drifters [1, 6, 8, 12], since 12 is not in the idx list, it will just returned [1,8,6].

selipot:
Ok so if we do not apply the order by date function then we should not have this problem? My goal is to 1) create the new ragged array asap and 2) inform the DAC what they need to correct.

philippemiron:
Sorry for the delay, I was trying to test before answering but I'm getting an error and can't manage to create a ragged array. Looks like the function `rowsize` is not getting the right indices. I'll wait for the update to this PR to test again.

milancurcic:
Tangential to this, I don't understand how `order_by_date` does any sorting:

```python
def order_by_date(df: pd.DataFrame, idx: list[int]) -> np.ndarray[int]:
    """From the previously sorted directory files DataFrame, return the drifter
    indices sorted by their end date.

    Args:
        idx [list]: List of drifters to include in the ragged array
    Returns:
        idx [list]: Sorted list of drifters
    """         
    return df.ID[np.where(np.in1d(df.ID, idx))[0]].values
```
As I understand it, `np.in1d(...)` returns True where an element in `df.ID` is present in `idx` and False otherwise. And `np.where()` of that returns the indices. So it seems to me that any ordering here happens only if the IDs in the directory files are already sorted by end date. If that's true, this may be error prone because if the ordering method in the directory file changes between dataset releases, this function won't be doing what it says it does anymore. It's very possible that I'm missing something obvious here and just confused. I apologize if so.

Why don't we use the metadata from the downloaded files to sort by end date?

philippemiron:
> Tangential to this, I don't understand how `order_by_date` does any sorting:
> 
> ```python
> def order_by_date(df: pd.DataFrame, idx: list[int]) -> np.ndarray[int]:
>     """From the previously sorted directory files DataFrame, return the drifter
>     indices sorted by their end date.
> 
>     Args:
>         idx [list]: List of drifters to include in the ragged array
>     Returns:
>         idx [list]: Sorted list of drifters
>     """         
>     return df.ID[np.where(np.in1d(df.ID, idx))[0]].values
> ```
> 
> As I understand it, `np.in1d(...)` returns True where an element in `df.ID` is present in `idx` and False otherwise. And `np.where()` of that returns the indices. So it seems to me that any ordering here happens only if the IDs in the directory files are already sorted by end date. If that's true, this may be error prone because if the ordering method in the directory file changes between dataset releases, this function won't be doing what it says it does anymore. It's very possible that I'm missing something obvious here and just confused. I apologize if so.
> 
> Why don't we use the metadata from the downloaded files to sort by end date?

The sorting is done by sorting the metadata already. See here,


```
def get_gdp_metadata() -> pd.DataFrame:
    """Download and parse GDP metadata and return it as a Pandas DataFrame."""

    directory_file_names = [
        "dirfl_1_5000.dat",
        "dirfl_5001_10000.dat",
        "dirfl_10001_15000.dat",
        "dirfl_15001_jul22.dat",
    ]
    df = pd.concat([parse_directory_file(f) for f in directory_file_names])
    df.sort_values(["Deployment_date"], inplace=True, ignore_index=True)
    return df
```

The metadata files are read, and then the values are sorted by the `Deployment_date`.

milancurcic:
Thanks, Philippe, that's what I thought. OK, we can address it in a separate PR, it's not a big deal here.

@selipot how would you like to proceed here? Run `order_by_date` and get a subset of drifters or don't `order_by_date` and get all available?

selipot:
I am waiting to hear some clarification from @RickLumpkin. I think we should create a ragged array file which contains the drifters listed in the currently and publicly available directory files (latest as of right now is `dirlf_15001_jul222.dat`) but it seems that the experimental set contains some drifters (202) that are not in that file but in another DAC internal directory file. So in effect applying any date sorting from the directory files filters out those unreferenced drifters. 

As for sorting by end date that is not ideal because all drifters still active by the cutoff date of an update have the same end date so I would prefer sorting by the start date in the directory file. There still remain the issues of some deployment dates (from the [deplog](https://www.aoml.noaa.gov/phod/dac/deployed.html) ) being later than the start date. Some of these are unfixable and @RickLumpkin will be replacing those by `_FillValues`. 

So let's wait just a bit for more info from the DAC.

RickLumpkin:
Hi all - I've blocked some time today to work on this.  I'll provide an
update if I can make significant progress.

milancurcic:
Update: There are now 19082 files in experimental and only 3 are not found in the metadata (I expanded the metadata to end with oct22, which used to be jul22). These 3 are 12294040, 12469250, and 60725400.

```python
In [9]: drifter_ids.size
Out[9]: 19082

In [10]: gdp_metadata = get_gdp_metadata()

In [11]: gdp_metadata.ID.values.size
Out[11]: 27045

In [12]: np.intersect1d(drifter_ids, gdp_metadata.ID.values).size
Out[12]: 19079

In [13]: np.setdiff1d(drifter_ids, gdp_metadata.ID.values).size
Out[13]: 3

In [14]: np.setdiff1d(drifter_ids, gdp_metadata.ID.values)
Out[14]: array([12294040, 12469250, 60725400])
```

milancurcic:
@RickLumpkin is it feasible to publish a single, concatenated directory file `dirfl.dat`, which may update its contents over time but keeps the name and URL constant? Currently the last file in the list is a moving target and changed its name:

```diff
-        "dirfl_15001_jul22.dat",
+        "dirfl_15001_oct22.dat",
```

which is fragile to have hardcoded in a library.

RickLumpkin:
Hi Milan - it's possible to create a copy of the latest version as
something like dirfl_15001_current.dat.  Would that work?

milancurcic:
Yes, any predictable naming scheme will work. For example, I assume that `dirfl_15001_current.dat` will at some point be renamed to `dirfl_15001_20000.dat`, and a new `dirfl_20001_current.dat` will be introduced. If you agree to this scheme, then in the library we'll parse the files according to these rules:

* First file is `dirfl_1_5000.dat`;
* Last file begins with `dirfl_` and ends with `_current.dat`;
* Intermediate files follow the pattern of the first file but incremented by 5000.

RickLumpkin:
Hi Rick, Shane, Milan, et. al 

I’ve generated the requested files and pushed them to the web/ftp. 
As requested, you’ll now find a dirfl_15001_current.dat file, with an accompanying buoydata_15001_current.dat.gz file. 
These are accessible via ftp.aom.noaa.gov/phod/pub/buoydata <http://ftp.aom.noaa.gov/phod/pub/buoydata>, as well as https://www.aoml.noaa.gov/ftp/pub/phod/buoydata/ <https://www.aoml.noaa.gov/ftp/pub/phod/buoydata/>. 

Shane, would you like to have the following files available in _current format: 
buoydata_15001_mmmyy_edited_pfiles.dat.gz
buoydata_15001_mmmyy_edited_sfiles.dat.gz
buoydata_15001_mmmyy_rawfiles.dat.gz
dirfl_15001_mmmyy_psfiles.dat
dirfl_15001_mmmyy_rawfiles.dat

If so, I’ll generate/update these to ftp.aoml.noaa.gov/phod/pub/pazos/data/shane/sst/ <http://ftp.aoml.noaa.gov/phod/pub/pazos/data/shane/sst/>. 

Thank you all,
Shaun 

Shaun R. Dolk
NOAA/AOML/PhOD
4301 Rickenbacker Cswy
Miami, Florida U.S.A. 33149

Email: ***@***.*** ***@***.***>
Office: 305.361.4546

RickLumpkin:
Hi Milan,

You are not doing anything wrong, if you're looking at the hourly data.
Shaun found that 43 IDs still inadvertently had their 8-digit IDs assigned,
and has switched them to their 15-digit versions.  I have corrected this in
the 6h Matlab and NetCDF files, but not yet in the hourly files.  I'll let
you all know when it's fixed.

  Best,
   Rick

milancurcic:
Rick, great to hear. OK, I'll wait for the hourly data then. I misunderstood that it was already processed. Thanks!

RickLumpkin:
You're welcome - it was reprocessed, but with this recent ID fix it now
needs to be re-reprocessed.

RickLumpkin:
Dear all,

I have now regenerated all of the "experimental" hourly NetCDF files,
containing locations and velocities, at
ftp://ftp.aoml.noaa.gov/phod/pub/lumpkin/hourly/experimental
This should fix the problems with the 43 IDs that were switched from
8-digit to 15-digit.

  Best,
   Rick

milancurcic:
Hi Rick, am I still looking at this too soon?

There are many files in experimental/ that have less-than-15-digit IDs. There are still 43 files that have mismatch with the directory file, and they're the same IDs as before.

RickLumpkin:
Ah, I created and uploaded the 43 new files, but accidentally didn't delete
the 43 files with the truncated IDs.  Hold on ...

PS many older drifters, including all Argos drifters, will have less than
15 digits in their ID.

RickLumpkin:
OK, I've removed those NetCDF files from the directory.  They are for the
following IDs that have been replaced with 15-digit IDs:
    12294040
    12469250
    60725400
     2020620
     2020710
     2021710
     2022690
     2022710
     2023710
     2024660
     2024710
     2025710
     2026650
     2026710
     2027650
     2027690
     2027700
     2028610
     2028650
     2028700
     2029610
     2029700
     2516470
     2520260
     2520280
     2521250
     2521260
     2522250
     2522260
     2523250
     2523260
     2524160
     2524250
     2525260
     2527250
     2527260
     2528250
     2528260
     2529260
     2624430
     2720390
     2720460
     2726390

Issue #70, opened 2023-01-08
`velocity_from_position`: Handle ragged array


milancurcic:
Previous discussion in #68.

@selipot suggested that `velocity_from_position` should also handle ragged arrays as input. Let's discuss here what these ragged arrays look like. I.e. is the ragged array in the form of an xarray Dataset as generated by clouddrift or something else?

selipot:
I would say it can be Xarray Datarray, Akward array, numpy array? In cases other than the nested awkward array we need to think of the way by which the function is made aware of trajectory breaks with an extra argument perhaps? As an example the 'id' or the 'rowsize' variables? This method will be generalizable to any other function we will write. We can go back to the EarthCube notebook to see how this was handled.

selipot:
Could any of our analysis functions could take optional arguments specifying the underlying structure of the data? I.e. option `rowsize=rowsize` for ragged arrays and `dim='obs` or `axis=n` for structured arrays such as xarray DataArrays?

milancurcic:
I like the `rowsize` approach.

If `rowsize` (array-like of ints, optional, default None) is absent, the computation defaults to the N-d structured array implementation.

If `rowsize` is provided, require that `x`, `y`, and `time` are 1-d arrays and apply boundary conditions at the start and end of each segment.

`dim` argument doesn't apply since our function tasks array-likes (I think this is a good design choice), and we already have `time_axis` to specify along which axis to differentiate.

If `rowsize` is provided, we can raise a warning if `time_axis` is also provided, and proceed with the computation (if we want to be more lax), or raise an error if we want to be more strict. Either is OK, it's a style choice.

milancurcic:
@selipot and I discussed jLab's `splitcell` (IIRC) function which splits a ragged array into a list of varying-length arrays.

That's easy for us to do as well given that we have `rowsize` for housekeeping. This made me think of an alternative approach to handling ragged array in `velocity_from_position` and similar functions that need to be trajectory-bounds aware:

The function could accept, in addition to array-likes, lists of array-likes. If the arguments are lists of array-likes, their elements are assumed to be contiguous arrays (trajectories), and the function would recursively run itself on each element and return lists of array-likes as result.

[...]

PR #112, opened 2023-02-15
Function to unpack a ragged array into a list of regular arrays

milancurcic:
* [x] Implementation
* [x] Tests
* [x] Example use in docstring

This function unpacks a ragged array into a list of regular arrays determined by rowsize. In support of #70 and other trajectory-bound-aware functions.

It's implemented using Philippe's EarthCube presentation approach (Xarray case 3).

Before proceeding to complete this PR, I'd like to get feedback on whether you like the direction and the API.

This function works with NumPy arrays and Xarray DataArrays, however passing DataArrays appears to be about 2 orders of magnitude slower. Here's a small benchmark with a ragged Xarray Dataset with 100 trajectories:

```
In [192]: %timeit res = unpack(ds.lon, ds.rowsize)
9.62 ms ± 9.27 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [193]: %timeit res = unpack(ds.lon.values, ds.rowsize.values)
163 µs ± 64.7 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

Example:

```python
from clouddrift.adapters.gdp import to_raggedarray
from clouddrift.analysis import velocity_from_position
from clouddrift.dataformat import unpack

ds = to_raggedarray(n_random_id=100).to_xarray()

for lon, lat, time in list(zip(
    unpack(ds.lon.values, ds.rowsize),
    unpack(ds.lat.values, ds.rowsize),
    unpack(ds.time.values, ds.rowsize)
)):
    u, v = velocity_from_position(lon, lat, time)
    plt.scatter(lon, lat, c=np.sqrt(u**2 + v**2), s=10)
plt.colorbar()
```

Figure_1

### DIFF
--- a/clouddrift/dataformat.py
+++ b/clouddrift/dataformat.py
@@ -384,9 +384,9 @@ def unpack_ragged(
 ) -> list[np.ndarray]:
     """Unpack a ragged array into a list of regular arrays.
 
-    Unpacking ragged_array as an np.ndarray is about 2 orders of magnitude
-    faster than unpacking it as an xr.DataArray, so unles you need a DataArray
-    as the result, we recommend passing np.ndarray as input.
+    Unpacking a ``np.ndarray`` ragged array is about 2 orders of magnitude
+    faster than unpacking an ``xr.DataArray`` ragged array, so unless you need a
+    ``DataArray`` as the result, we recommend passing ``np.ndarray`` as input.
 
     Parameters
     ----------
--- a/clouddrift/dataformat.py
+++ b/clouddrift/dataformat.py
@@ -395,5 +395,5 @@ def unpack(ragged_array: np.ndarray, rowsize: np.ndarray[int]) -> list[np.ndarra
             correspond to the values in rowsize, and types that correspond
             to the type of ragged_array
     """
-    indices = np.insert(np.cumsum(ds.rowsize.values), 0, 0)
-    return [ragged_array[indices[n]:indices[n+1]] for n in range(indices.size - 1)]
+    indices = np.insert(np.cumsum(np.array(rowsize)), 0, 0)
+    return [ragged_array[indices[n] : indices[n + 1]] for n in range(indices.size - 1)]
--- a/clouddrift/dataformat.py
+++ b/clouddrift/dataformat.py
@@ -379,21 +379,50 @@ class RaggedArray:
         return
 
 
-def unpack(ragged_array: np.ndarray, rowsize: np.ndarray[int]) -> list[np.ndarray]:
+def unpack_ragged(
+    ragged_array: np.ndarray, rowsize: np.ndarray[int]
+) -> list[np.ndarray]:
     """Unpack a ragged array into a list of regular arrays.
 
     Unpacking ragged_array as an np.ndarray is about 2 orders of magnitude
     faster than unpacking it as an xr.DataArray, so unles you need a DataArray
     as the result, we recommend passing np.ndarray as input.
 
-    Args:
-        ragged_array (array-like): A ragged_array to unpack
-        rowsize (array-like): An array of integers whose values is the size of
-            each row in the ragged array
-    Returns:
-        unpacked_arrays (list): A list of array-likes with sizes that
-            correspond to the values in rowsize, and types that correspond
-            to the type of ragged_array
+    Parameters
+    ----------
+    ragged_array : array-like
+        A ragged_array to unpack
+    rowsize : array-like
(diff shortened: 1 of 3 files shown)


philippemiron:
Not sure if this is really needed to be honest. And for large arrays you have to duplicate all the memory to do this loop. I guess something like this should work, and there is probably a better way to use `map` and apply the `velocity_from_position` to segments of the arrays.

```
from clouddrift.adapters.gdp import to_raggedarray
from clouddrift.analysis import velocity_from_position
from clouddrift.dataformat import unpack

ds = to_raggedarray(n_random_id=100).to_xarray()

index_traj = np.insert(np.cumsum(rowsize), 0, 0)
for i in range(0, len(rowsize)):
    r = index_traj[i:i+2]
    u, v = velocity_from_position(ds.lon[r], ds.lat[r], ds.lat.time[r])
    plt.scatter(lon, lat, c=np.sqrt(u**2 + v**2), s=10)
plt.colorbar()
```

milancurcic:
I agree with you that it's not strictly needed, but the purpose of it is to abstract away an implementation detail (`rowsize`) from the user.

selipot:
The goal here is to emulate what is done in jLab by @jonathanlilly  with [`cellsplit.m`](https://github.com/jonathanlilly/jLab/blob/master/jCell/cellsplit.m), a function that allows one to split individual trajectories saved in a Matlab cell array into several sub-trajectories (which we will call **segments**) and hence a bigger cell array, based on some criteria such as a time gap. I am not sure what is the best way to do this is as above. 

As an example, starting from the GDP ragged array, remembering that individual trajectories have sometimes gaps of 12h or longer. How to create a new ragged array and/or a new rowsize variable that will index contiguous hourly segments instead of continuous ids?  

Instead of unpack, let's call this modifier/function `segment`; perhaps the syntax is:

```
newrowsize = segment(ds.time.values, ds.rowsize,condition)
```

milancurcic:
I like `segment` as well. However, the purpose of this function (as in, in this PR) is not exactly what cellsplit does. It is merely to unpack individual trajectories from a 1-d ragged array representation, regardless of whether they have gaps or not.

I think we should pursue the cellsplit functionality as well, but as a separate function, because these are two separate features--1) segmenting by trajectory and 2) segmenting by contiguous time intervals.

selipot:
Ok so what's next?

selipot (approved):

selipot:
Maybe it should be called `ragged_to_list` or `unpack_ragged`

clouddrift/dataformat.py
```diff
@@ -377,3 +377,23 @@ def to_parquet(self, filename: str):
         """
         ak.to_parquet(self.to_awkward(), filename)
         return
+
+
+def unpack(ragged_array: np.ndarray, rowsize: np.ndarray[int]) -> list[np.ndarray]:
+    """Unpack a ragged array into a list of regular arrays.
+
+    Unpacking ragged_array as an np.ndarray is about 2 orders of magnitude
```
selipot, line 385: Unpacking a np.ndarray ragged_array is ...

clouddrift/dataformat.py
```diff
@@ -377,3 +377,23 @@ def to_parquet(self, filename: str):
...
+    faster than unpacking it as an xr.DataArray, so unles you need a DataArray
```
selipot, line 386: faster than unpacking a xr.DataArray ragged array, so unless you need a DataArray

milancurcic:
Cool. I'll finish up this PR then (tests and example in docstring) and open an issue for segmenting by time gaps. In light of that, let sort out the naming for these two functions so they don't conflict and they're each clear.

> Maybe it should be called ragged_to_list or unpack_ragged

I actually had `unpack_ragged` first, but then removed it as I thought it was redundant as it's in a ragged array module (and we should rename it as such). `unpack_ragged` is certainly more clear and specific, so I like that. I like `ragged_to_list` slightly less (3 words instead of 2 kind of thing).

So, let's go with `unpack_ragged` here and reserve `segment` for segmenting by time gap?

selipot (changes requested):

selipot:
> So, let's go with `unpack_ragged` here and reserve `segment` for segmenting by time gap?

`unpack_ragged` sounds good. I would like to keep `segment` for segmenting by any arbitrary  criteria, not only time.

milancurcic:
I completed the remaining items and addressed the requested changes, so I'll go ahead and merge. Thanks all.

PR #117, opened 2023-02-18
63 function to recast longitude values 0360 or 180180

### LINKED ISSUE #63: Function to recast longitude values [0,360] or [-180,180]
lon = longitude_unwrap(lon,pivot=[180,360]) : recast longitudes in [-180,180[ or or [0,360[
See in jLab deg180.m, deg360.m and inversion degunwrap.m

philippemiron:
First draft of the recast longitude function (Fix #63). Let me know what you think and what we should modify.

### DIFF
--- a/clouddrift/__init__.py
+++ b/clouddrift/__init__.py
@@ -6,3 +6,4 @@ from clouddrift.dataformat import RaggedArray
 import clouddrift.adapters
 import clouddrift.analysis
 import clouddrift.haversine
+import clouddrift.sphere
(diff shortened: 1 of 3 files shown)


selipot (approved):
Can't test right now but are the range/intervals open or closed? i.e. is recast_longitude(-180,-180) = -180 or 180?

philippemiron:
recast_longitude(-180,-180) = -180, I guess this is the expected behavior?

selipot:
I think that's an acceptable behavior. I see that jlab `deg180(-180)=180` but I don't think there's a wrong or right answer here.

I reallylike the `sphere` module name. 

Thank you @philippemiron !

selipot:
Should we have included for `recast_longitude` an optional argument specifying the dimension over which to apply? Should we set this to be a default for all subsequent functions?

milancurcic:
I think in general we should, but it doesn't apply here because the function is element-wise.

PR #118, opened 2023-02-18
add sphere to the docs

philippemiron:


philippemiron:
Maybe, let's keep it open until @milancurcic sees it because I might be forgetting other things.

milancurcic (approved):
It's correctly included in the API docs index.

However, do check one thing: I noticed that your format of docstrings (e.g. like the one you used here) are not properly rendered by Sphinx. I don't know if it's a missing Sphinx setting or just a bad format. In contrast, look at the `dataformat.unpack_ragged` docstring where I used the NumPy style format and how it's rendered.

I think this PR should just go ahead as is, but in a separate PR we need to go over all docstrings and fix them up. GitHub Copilot is now pretty good at autogenerating them. I suggest just following the NumPy style guide for docstrings.

philippemiron:
I'll take a look later. I can probably go through all the files and do something uniform.

philippemiron:
ps: my branch was not up-to-date with main, so I guess docstring should be fixed in another PR as you suggested. Also, I'm not sure how to proceed, for example it looks like you wrote: `array-like` as type of parameters, and I was using directly the same keyworkd that I used for `typing`..

milancurcic:
> Also, I'm not sure how to proceed, for example it looks like you wrote: array-like as type of parameters, and I was using directly the same keyworkd that I used for typing..

Yes, I wasn't 100% sure what to do about this either. I followed NumPy doc style to use array-like to indicate that the function will work with `np.ndarray` and `xr.DataArray` alike. Perhaps we can just be explicit and say `np.ndarray | xr.DataArray`, although I think this type syntax isn't supported before Python 3.10. Something to check.

Either way, the type hints in the docstring don't need to exactly match those in the function definition. The ones in the function definition are for IDEs to parse, and the ones in the docstring merely need to make sense to the reader.

I think we just need to pick what makes sense for us in this project, and be consistent with it.

PR #119, opened 2023-02-19
Running code coverage with the automatic tests

philippemiron:
This generates a report on the % of line of codes that are covered by the unit tests, and it is uploaded to codecov.io. See here, https://app.codecov.io/gh/Cloud-Drift/clouddrift/tree/codecov/clouddrift, for the `codecov` branch that I did to test the implementation. We are at 79%, which is pretty good! It's quite useful to identify quickly which parts of the code we are not testing.

I also added a badge on the README, right now, it says unknown because it did not run on the main branch yet.

### DIFF
--- a/.github/workflows/black.yml
+++ b/.github/workflows/black.yml
@@ -12,5 +12,5 @@ jobs:
   lint:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


philippemiron:
Ps: it also shows on PR, as you can see below.

milancurcic (approved):
This is useful, thank you!

PR #120, opened 2023-02-20
Update README

### LINKED ISSUE #106: Update README.m
- [ ] Move first and center the link to the documentation
- [ ] move down the install section

What is now the recommended use of the `.yml` file? And as a corollary, the use of the `clouddrift` environment that can be created from it? I understand we removed this from the README some time ago.

milancurcic:
This PR makes a few updates to the README:

* Use active voice in the first sentence
* Remove note about system libs and pip
* Add instructions to build locally with pip
* Bring docs and examples links to the top (fixes #106)

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,6 +97,18 @@ Then, run the tests like this:
 python -m unittest tests/*.py
 ```
 
+### Installing CloudDrift on unsupported platforms
+
+One or more dependencies of CloudDrift may not have pre-built wheels for
+platforms like IBM Power9 or Raspberry Pi.
+If you are using pip to install CloudDrift and are getting errors during the
+installation step, try installing CloudDrift using Conda.
+If you still have issues installing CloudDrift, you may need to install system
+dependencies first.
+Please let us know by opening an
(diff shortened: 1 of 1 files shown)


philippemiron (approved):

milancurcic:
Thanks for the review and feedback.

Yes, on some less common platforms like IBM Power9 or Raspberry Pi, netCDF4 doesn't have pre-built wheels on PyPI, and in that case, you need to install a system dependency by hand.

And Conda is a 3rd party tool made by a for-profit corp that you need to install manually on any OS.

Neither solution is ideal and each has its pros and cons. Moreover, the differences are more cultural than technical. We need to support both kinds of users because neither is insignificant in numbers.

Let's please end the package manager wars here. The best we can do as a package is to support both package managers as best as we can. For Conda, we already go above and beyond thanks to your effort with conda-forge. And for pip, I'll aim to keep the complexity sufficiently low so that CloudDrift continues to build out of the box on the 3 major platforms.

I added a note about building on less common platforms.

selipot:
Tested successfully on my laptop with both pip and conda installs.

Issue #121, opened 2023-02-20
Question about `recast_longitude` implementation


milancurcic:
@philippemiron, I noticed that in `rot` you cast to to complex-phase:

```python
def rot(x: np.ndarray) -> np.ndarray:
    x = np.mod(x + np.pi, 2 * np.pi) - np.pi  # convert to [-pi, pi]
    return np.exp(1j * x)
```

and in `recast_longitude` you use `np.angle`:

```python
    return (
        np.mod(
            np.divide(360, 2 * np.pi)
            * np.unwrap(np.angle(rot(np.divide(2 * np.pi, 360) * (lon - lon0)))),
            360,
        )
        + lon0
    )
```

If `rot` is instead defined as 

```python
def rot(x: np.ndarray) -> np.ndarray:
    return np.mod(x + np.pi, 2 * np.pi) - np.pi  # convert to [-pi, pi]
```

could the `recast_longitude` implementation not be simpler:

```python
    return (
        np.mod(
            np.divide(180, np.pi)
            * np.unwrap(rot(np.deg2rad(lon - lon0))),
            360,
        )
        + lon0
    )
```

Notice the `np.deg2rad`, `np.angle`, and simplified `np.divide`.

It seems to me that it does the same thing without the real-to-complex round-trip, but it's possible that you did what you did for a good reason that I don't see.

milancurcic:
I just tried it and the tests continue to pass. Here's the diff

```diff
diff --git a/clouddrift/sphere.py b/clouddrift/sphere.py
index 186ccc2..2779b25 100644
--- a/clouddrift/sphere.py
+++ b/clouddrift/sphere.py
@@ -25,8 +25,7 @@ def rot(x: np.ndarray) -> np.ndarray:
         rot(np.pi) # approx. (-1+0j)
         rot(3 * np.pi / 2) # approx. (0-1j)
     """
-    x = np.mod(x + np.pi, 2 * np.pi) - np.pi  # convert to [-pi, pi]
-    return np.exp(1j * x)
+    return np.mod(x + np.pi, 2 * np.pi) - np.pi  # convert to [-pi, pi]
 
 
 def recast_longitude(lon: np.ndarray, lon0: Optional[float] = -180) -> np.ndarray:
@@ -77,8 +76,8 @@ def recast_longitude(lon: np.ndarray, lon0: Optional[float] = -180) -> np.ndarra
 
     return (
         np.mod(
-            np.divide(360, 2 * np.pi)
[...]

PR #124, opened 2023-02-22
recast longitude

philippemiron:
- rename function to recast_lon
- added: recast_lon180, recast_lon360 because I guess it can be useful

### DIFF
--- a/clouddrift/sphere.py
+++ b/clouddrift/sphere.py
@@ -1,6 +1,7 @@
 import numpy as np
 from typing import Optional
 
+
 def recast_lon(lon: np.ndarray, lon0: Optional[float] = -180) -> np.ndarray:
     """Recast (convert) longitude values to a selected range of 360 degrees starting from ``lon0``.
 
--- a/clouddrift/sphere.py
+++ b/clouddrift/sphere.py
@@ -23,24 +23,25 @@ def recast_lon(lon: np.ndarray, lon0: Optional[float] = -180) -> np.ndarray:
     By default, ``recast_lon`` converts longitude values to the range
     `[-180, 180]`:
 
-    .. code-block:: python
-
-        recast_lon(200) # -160
+    >>> recast_lon(200)
+    -160
 
     The range of the output longitude is controlled by ``lon0``.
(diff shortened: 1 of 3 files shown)


milancurcic:
Why not also update the implementation as per https://github.com/Cloud-Drift/clouddrift/issues/121#issuecomment-1437781149? If it indeed works, it's super concise and allows us to remove the `rot` thing.

I don't think adding separate functions for specific values of an argument (e.g. `recast_lon360(lon)` vs. `recast_lon(lon, 360)`) is a good idea because it increases the surface area of the API, and the number of tests, for no meaningful gain. But if you and @selipot think it's useful, I don't oppose it, only recommend against it.

selipot:
> Why not also update the implementation as per [#121 (comment)](https://github.com/Cloud-Drift/clouddrift/issues/121#issuecomment-1437781149)? If it indeed works, it's super concise and allows us to remove the `rot` thing.
> 

We can implement it as in  [#121 (comment)](https://github.com/Cloud-Drift/clouddrift/issues/121#issuecomment-1437781149). The current implementation is based on jlab and the matlab expectation that the function `unwrap` takes only radian as an input, hence the use of complex exponential. The numpy version of `unwrap` does not only take radian input hence the possible simpler implementation. Let's do it. 

> I don't think adding separate functions for specific values of an argument (e.g. `recast_lon360(lon)` vs. `recast_lon(lon, 360)`) is a good idea because it increases the surface area of the API, and the number of tests, for no meaningful gain. But if you and @selipot think it's useful, I don't oppose it, only recommend against it.

I want them, I use them all the time. `recast_lon360` is actually `recast_lon(lon,0)`. I don't think we have to worry about the surface area of the API at the moment.

milancurcic:
Sounds good. Let's update the implementation, add tests for the two new functions, and expand the docstrings before merging. Since the previous implementation is already in main, let's wait for this PR before releasing 0.7.0.

milancurcic:
1.0 is so far beyond the horizon that I don't worry about it. I could go at depth why short and frequent release cycles are good.

milancurcic:
This is in good shape now so I'll go ahead and merge. Thanks, all!

Issue #127, opened 2023-02-25
Discussion ragged array support


philippemiron:
I just saw that the segment function supports ragged array, and I was wondering what you think of including something like this inside `clouddrift`.

```python
def apply_ragged(func, arr, rowsize, *args, **kwargs):
    """
    Apply a function over segments of a ragged array
    """
    if not sum(rowsize) == len(x):
            raise ValueError("The sum of rowsize must equal the length of x.")
            
    output = []
    start = 0
    for r in rowsize:
        end = start + int(r)
        output.append(func(arr[start:end], *args, **kwargs))
        start = end
    return np.concatenate(output)

from concurrent import futures
def apply_ragged_thread(func, arr, rowsize, *args, **kwargs):
    """
    Apply a function over segments of a ragged array
    """
    if not sum(rowsize) == len(x):
            raise ValueError("The sum of rowsize must equal the length of x.")
    
    def wrapper(x):
        return func(x, *args, **kwargs)        
        
    with futures.ThreadPoolExecutor() as executor:
        outputs = executor.map(wrapper, np.array_split(arr, np.cumsum(rowsize[:-1])))
    return np.concatenate([o for o in outputs])
```

The main advantage is that we can remove `if rowsize is None:` in `segments` (and future functions that I guess would require a similar construct), but of course the user now has to use this function if they have a ragged array. 

For example, I've tested with the `segment` function.

```python
x = [0, 1, 1, 1, 2, 2, 3, 3, 3, 3, 4]
rowsize = [3, 2, 6]
segment(x, 0.5, rowsize)
> array([1, 2, 1, 1, 1, 4, 1])
```

```python
apply_ragged(segment, x, rowsize, tolerance=0.5)
> array([1, 2, 1, 1, 1, 4, 1])
```

```python
apply_ragged_thread(segment, x, rowsize, tolerance=0.5)
> array([1, 2, 1, 1, 1, 4, 1])
```

PS: So far, I have not found an elegant way of performing the same for `n` ragged array arguments, e.g. `velocity_from_position`, which expects `x`, `y`, and `t`.

philippemiron:
Ok, I'm pretty happy because I just now managed to make it work right now for `n` parameters and arguments.

Here are the new functions. There are two versions, one is a simple map and the other one is doing the calculation on different threads. But basically both are doing the same thing.

```
from concurrent import futures
from collections.abc import Iterable
import numpy as np
from clouddrift.analysis import segment, velocity_from_position

[...]

PR #128, opened 2023-02-25
apply ragged

philippemiron:


philippemiron:
Follow up to #127.

philippemiron:
Sorry, I did not sync my main properly. Hopefully, it's good now.

selipot:
I'd like to test it. It looks different from what we discussed [here](https://github.com/Cloud-Drift/clouddrift/issues/127#issuecomment-1445181016). What is the syntax to apply in parallel?

philippemiron:
I just put the `apply_ragged_thread` version but called it `apply_ragged`. No real advantage to not process this in parallel.

clouddrift/analysis.py
```diff
@@ -1,9 +1,59 @@
...
+        res = executor.map(lambda x: func(*x, **kwargs), iter)
+    # concatenate the outputs
+    res = list(res)
+    if isinstance(res[0], tuple):  # more than 1 parameter
+        outputs = []
+        for i in range(len(res[0])):
+            outputs.append(np.concatenate([r[i] for r in res]))
+        return tuple(outputs)
+    else:
+        return np.concatenate([r for r in res])
```
milancurcic, line 54:
```suggestion
        return np.concatenate(res)
```

milancurcic:
I think it's good that this is one function rather than two separate functions for serial vs. concurrent.

The users should be able to control this behavior, though--there are scenarios in which I don't want multithreading to happen. I suggest adding an optional `max_workers` argument to `apply_ragged`, and which defaults to the `ThreadPoolExecutor`s default for `max_workers`.

Should this be in the `dataformat` (should be `raggedarray`, really, but that's for another PR) module rather than analysis? I think so.

clouddrift/analysis.py
```diff
@@ -1,9 +1,59 @@
...
+    # make sure the arrs is iterable
+    if not isinstance(arrs[0], Iterable):
+        arrs = [arrs]
+    # validate rowsize
+    for arr in arrs:
+        if not sum(rowsize) == len(arr):
+            raise ValueError("The sum of rowsize must equal the length of arr.")
+
+    # split the array(s) into trajectories
+    arrs = [np.array_split(arr, np.cumsum(rowsize[:-1])) for arr in arrs]
```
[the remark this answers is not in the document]
philippemiron: Yes this step was slow and returns a copy. Good find.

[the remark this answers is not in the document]
philippemiron: Could you add the import of `unpack_ragged` too?

clouddrift/analysis.py
```diff
@@ -1,7 +1,54 @@
 import numpy as np
+from collections.abc import Iterable
 from typing import Optional, Tuple
 import xarray as xr
+from concurrent import futures
 from clouddrift.haversine import distance, bearing
+from clouddrift.dataformat import unpack_ragged
+
+
+def apply_ragged(func, rowsize, arrs, *args, max_workers=None, **kwargs):
```
milancurcic, line 10: What do you think about re-ordering `arrs` and `rowsize` in the function signature?

```suggestion
def apply_ragged(func, arrs, rowsize, *args, max_workers=None, **kwargs):
```

I think this is a more natural way to order it: We say something like "Apply `func` to `arrs` for every row of `rowsize`". It would also be more consistent with the `map` signature (`map(f, x)`).

philippemiron: I don't have any preference.

selipot: I prefer the order `arrs, rowsize`

milancurcic:
While writing tests for inputting DataArrays, I found that even if `func` is a `DataArray -> DataArray` function, `apply_ragged` will return `np.ndarray`.

I can think of two options for us how to deal with this:

1. Do nothing and document that `apply_ragged` will always return `np.ndarrays`.
2. In `apply_ragged`, after `func` has returned, do a type check on the result (i.e. is it np.ndarray or xr.DataArray) and do a type-aware concatenation.

What do you think? I think 2 would be nice.

philippemiron:
The only thing I'm wondering is how to deal with coordinates associated with the values in the DataArray? It might not be so simple since it depends on the operation you are performing..

milancurcic:
Can you write an example?

From my understanding: Say you have a list of DataArrays that are each a result of `func(arr)` where `arr` is one row of a ragged DataArray. They all have the same coordinates, and concatenating with `xr.concat` preserves these coordinates. We do need to pass `dim` to `xr.concat`, which we can default to -1, and make an optional argument to `apply_ragged` if we choose to support that. Or do you mean something else?

philippemiron:
I was not sure it would always conserve the coordinates. I guess then we just check and concat using the appropriate method (your 2nd choice).

milancurcic:
I played a little bit with type-aware concatenation and it seems to work the way I imagined. But to do this I introduced a new function, so let's tackle that in a separate PR.

```python
def pack_ragged(arrays: list[Union[np.ndarray, xr.DataArray]]) -> Union[np.ndarray, xr.DataArray]:
    if type(arrays[0]) is np.ndarray:
        return np.concatenate(arrays)
    elif type(arrays[0]) is xr.DataArray:
        return xr.concat(arrays, dim=arrays[0].dims[-1])
    else:
        raise TypeError("arrays must be a list of np.ndarray or xr.DataArray")
```

clouddrift/analysis.py
```diff
@@ -1,7 +1,94 @@
...
+    >>> apply_ragged(func, [x, y], [5, 5])
+    array([10, 12, 14, 16, 18, 20, 22, 24, 26, 28])
+
+    Raises
+    ------
+    ValueError
+        If the sum of ``rowsize`` does not equal the length of ``arrs``.
+    """
+    # make sure the arrs is iterable
+    if not isinstance(arrs[0], Iterable):
```
milancurcic, line 64: This fails if a DataArray is passed as a scalar because an element of a DataArray is also a DataArray (annoyingly so).

milancurcic (approved):
I think this is now in great shape and good to merge. Thanks, @philippemiron!

philippemiron:
The `pack_ragged` looks good, I also thought that this should be implemented in as separate function!

PR #134, opened 2023-03-02
segments floats or timedelta

### LINKED ISSUE #133: segment function on time objects
The `segment` function is not able to handle `datetime` objects. This needs to work, as an example for the GDP dataset. The following should work, but does not:

```
import xarray as xr
from clouddrift.analysis import segment
from datetime import timedelta
ds = xr.open_dataset("gdp_v2.00.nc")
traj_idx = np.insert(np.cumsum(ds.rowsize.values), 0, 0)
j = 100
sli = slice(traj_idx[j], traj_idx[j+1])
dt = timedelta(seconds=1800)
row1 = segment(time[sli],tolerance=dt)
```

philippemiron:
Fix #133

### DIFF
--- a/clouddrift/analysis.py
+++ b/clouddrift/analysis.py
@@ -154,6 +154,7 @@ def segment(
         positive_tol = tolerance >= timedelta(seconds=0)
     else:
         positive_tol = tolerance >= 0
+
     if rowsize is None:
         if positive_tol:
             exceeds_tolerance = np.diff(x) > tolerance
--- a/clouddrift/analysis.py
+++ b/clouddrift/analysis.py
@@ -93,7 +93,8 @@ def apply_ragged(
 
 def segment(
     x: np.ndarray,
-    tolerance: Union[float, timedelta], rowsize: np.ndarray[int] = None,
+    tolerance: Union[float, timedelta],
+    rowsize: np.ndarray[int] = None,
 ) -> np.ndarray[int]:
     """Segment an array into contiguous segments.
 
--- a/clouddrift/analysis.py
+++ b/clouddrift/analysis.py
@@ -2,6 +2,7 @@ import numpy as np
 from typing import Optional, Tuple, Union
 import xarray as xr
 from concurrent import futures
+import datetime
 from clouddrift.haversine import distance, bearing
 from clouddrift.dataformat import unpack_ragged
 
@@ -91,7 +92,9 @@ def apply_ragged(
 
 
 def segment(
-    x: np.ndarray, tolerance: float, rowsize: np.ndarray[int] = None
+    x: np.ndarray,
+    tolerance: float | datetime.timedelta,
+    rowsize: np.ndarray[int] = None,
 ) -> np.ndarray[int]:
     """Segment an array into contiguous segments.
 
@@ -147,8 +150,13 @@ def segment(
     array([2, 2, 2, 2])
     """
 
+    if type(tolerance) == datetime.timedelta:
+        positive_tol = tolerance > datetime.timedelta(seconds=0)
+    else:
+        positive_tol = tolerance > 0
+
     if rowsize is None:
-        if tolerance >= 0:
+        if positive_tol:
             exceeds_tolerance = np.diff(x) > tolerance
         else:
             exceeds_tolerance = np.diff(x) < tolerance
--- a/clouddrift/analysis.py
+++ b/clouddrift/analysis.py
@@ -93,8 +93,7 @@ def apply_ragged(
 
 def segment(
     x: np.ndarray,
-    tolerance: float | timedelta,
-    rowsize: np.ndarray[int] = None,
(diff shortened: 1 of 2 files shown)


tests/analysis_tests.py
```diff
@@ -45,6 +45,17 @@ def test_segment_rowsize_raises(self):
         with self.assertRaises(ValueError):
             segment(x, tol, rowsize)
 
+    def test_segments_datetime(self):
+        x = [
+            np.datetime64("2023-01-01"),
+            np.datetime64("2023-01-02"),
+            np.datetime64("2023-01-03"),
+            np.datetime64("2023-02-01"),
+            np.datetime64("2010-02-01"),
```
selipot, line 54: I think there's an issue here. Did you mean "2023-02-02"?

selipot: Let's put a test with hour intervals. We want the tolerance to be the maximum difference allowed so that `tol = np.timedelta64(1,"H")` will split the GDP hourly ragged array, correct?

tests/analysis_tests.py
```diff
@@ -45,6 +45,17 @@ def test_segment_rowsize_raises(self):
         with self.assertRaises(ValueError):
             segment(x, tol, rowsize)
 
+    def test_segments_datetime(self):
+        x = [
+            np.datetime64("2023-01-01"),
```
selipot, line 50: the function uses `datetime.timedelta` but here you use `np.datetime64`. Does it matter?

selipot (changes requested):
I will test soon with the GDP!

tests/analysis_tests.py
philippemiron: Yes, that was a typo. I'm not sure I understand the difference with the test you want to do (apart from 1h instead of 1d).

tests/analysis_tests.py
philippemiron: I modified to datetime for consistency, but looks like both are compatible.

clouddrift/analysis.py
```diff
@@ -91,7 +92,9 @@ def apply_ragged(
 
 
 def segment(
-    x: np.ndarray, tolerance: float, rowsize: np.ndarray[int] = None
+    x: np.ndarray,
+    tolerance: float | timedelta,
```
milancurcic, line 96: This syntax was introduced in Python 3.10. If we want to support Python 3.9, use Union:

```suggestion
    tolerance: Union[float, timedelta],
```

(but I personally am fine dropping 3.9 for improved type hint syntax)

clouddrift/analysis.py
```diff
@@ -147,8 +150,13 @@ def segment(
     array([2, 2, 2, 2])
     """
 
+    if type(tolerance) == timedelta:
+        positive_tol = tolerance > timedelta(seconds=0)
+    else:
+        positive_tol = tolerance > 0
```
milancurcic, line 156: Don't you need >= here? Otherwise `tolerance=0` would produce incorrect results, it seems.

```suggestion
    if type(tolerance) == timedelta:
        positive_tol = tolerance >= timedelta(seconds=0)
    else:
        positive_tol = tolerance >= 0
```

milancurcic (approved):
Looks good pending the review getting resolved.

clouddrift/analysis.py
milancurcic: Let's also add a unit test for tolerance of positive and negative zero.

philippemiron: Positive or negative 0?

philippemiron: If tolerance is 0, it should just return `np.arange(x)` no? Maybe it should be an ArgumentError.

clouddrift/analysis.py
philippemiron: I will make the tests run with 3.9 and 3.10 for now. So we can catch this type of error.. because I'm only using 3.10 (or 3.11) currently.

clouddrift/analysis.py
milancurcic: Please ignore my mention of a negative zero. I did some reading and see that both the IEEE standard and Python define signed zero such that `-0 == 0`.

If tolerance is 0, we tolerate consecutive values equal to each other to be a segment. So, `segment([1, 1, 2, 2], tolerance=0)` should return `[2, 2]`.

selipot: That's a good example right there to form a rowsize variable if it does not exist.

philippemiron:
I'm not sure why, but I marked conversations as resolved but the PR is still blocked.

selipot:
> I'm not sure why, but I marked conversations as resolved but the PR is still blocked.

Maybe because I requested some changes?

selipot (approved):

philippemiron:
I thought the comments being resolved would change to approve, but it doesn't.

PR #135, opened 2023-03-03
Support for different time date type

philippemiron:
Modifications after #133.

### DIFF
--- a/clouddrift/analysis.py
+++ b/clouddrift/analysis.py
@@ -154,6 +154,7 @@ def segment(
         positive_tol = tolerance >= timedelta(seconds=0)
     else:
         positive_tol = tolerance >= 0
+
     if rowsize is None:
         if positive_tol:
             exceeds_tolerance = np.diff(x) > tolerance
--- a/clouddrift/analysis.py
+++ b/clouddrift/analysis.py
@@ -93,7 +93,8 @@ def apply_ragged(
 
 def segment(
     x: np.ndarray,
-    tolerance: Union[float, timedelta], rowsize: np.ndarray[int] = None,
+    tolerance: Union[float, timedelta],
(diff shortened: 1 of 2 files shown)


clouddrift/analysis.py
```diff
@@ -150,8 +151,12 @@ def segment(
     array([2, 2, 2, 2])
     """
 
-    if type(tolerance) == timedelta:
-        positive_tol = tolerance >= timedelta(seconds=0)
+    # for compatibility with datetime list or np.timedelta64 arrays
+    if type(tolerance) in [np.timedelta64, timedelta]:
```
philippemiron, line 155: Should check that if the tolerance is a type of timedelta, but the array is not, it should return `AttributeError`.

milancurcic: I think it's better to not check that and let the incompatible comparison to raise an exception.

Suppose that a library introduces a new `datetime` class, perhaps `xr.datetime` (it wouldn't be the craziest thing to happen given how many datetimes and timedeltas there already is). If `xr.datetime` is `np.diff` compatible and also comparison-compatible with `pd.Timedelta`, then it will just work. And that was the original intent of Python's duck typing.

milancurcic (approved):
Looks good, thanks. I only suggest adding an example in the docstring of using `segment` with datetimes and timedeltas.

clouddrift/analysis.py
philippemiron: Sounds good.

selipot:
Yes I will

selipot:
docstring needs updating

philippemiron:
Updating?

selipot:
Yes, the docstring does not correspond to the signature for the function segment.

selipot:
Yes! For some reasons i could not comment.

PR #144, opened 2023-03-24
6-hourly GDP adapter

milancurcic:
This is a bit bigger refactor to reuse the common functions. Now only `to_raggedarray`, `download`, and a few constants are specific to `gdp1h` and `gdp6h`. The rest is in `gdp`.

Short test with all 3:

```python
In [1]: from clouddrift.adapters import gdp1h, gdp6h

In [2]: ra_1h = gdp1h.to_raggedarray(n_random_id=100)
Downloading files: 100%|██████████████████████| 100/100 [00:11<00:00,  9.09it/s]
Retrieving the number of obs: 100%|█████████████| 98/98 [00:02<00:00, 44.88it/s]
/home/milan/Work/clouddrift/clouddrift/venv/lib/python3.9/site-packages/xarray/coding/variables.py:148: FutureWarning: elementwise comparison failed; returning scalar instead, but in the future will perform elementwise comparison
  condition |= data == fv
Filling the Ragged Array: 100%|█████████████████| 98/98 [00:09<00:00, 10.08it/s]

In [3]: ra_1h_exp = gdp1h.to_raggedarray(n_random_id=100, url=gdp1h.GDP_DATA_URL_EXPERIMENTAL)
Downloading files: 100%|██████████████████████| 100/100 [00:11<00:00,  8.74it/s]
Retrieving the number of obs: 100%|███████████| 100/100 [00:02<00:00, 43.48it/s]
...
Filling the Ragged Array: 100%|███████████████| 100/100 [00:08<00:00, 11.80it/s]

In [4]: ra_6h = gdp6h.to_raggedarray(n_random_id=100)
Downloading files: 100%|██████████████████████| 100/100 [00:07<00:00, 14.12it/s]
Retrieving the number of obs: 100%|███████████| 100/100 [00:01<00:00, 50.78it/s]
...
Filling the Ragged Array: 100%|███████████████| 100/100 [00:07<00:00, 12.85it/s]
```

Closes #104 #139

PR #146, opened 2023-03-25
142-subset

### LINKED ISSUE #142: subset
Since we are not using awkward arrays, we should redo the [subset function](https://github.com/Cloud-Drift/clouddrift/blob/main/clouddrift/select.py) so it's compatible with the xarray Dataset.

philippemiron:
Resolves #142.


selipot:
Look useful. 
1) So is it to have an easier syntax than `.where()`, `.loc()`, and `.iloc()`, especially with ranges? Such as lon>lon1 & lon<lon2?  
2) Can be applied to any xarray DataArray, not only ragged array?
3) More generally, could it be a PR for xarray?

clouddrift/select.py
```diff
@@ -1,40 +1,41 @@
-import awkward as ak
+import xarray as xr
 import numpy as np
+import warnings
 
 
-def mask_var(var: ak.Array, value):
+def mask_var(var: xr.DataArray, value):
```
milancurcic, line 6: I suggest `criteria` (consistent with `subset`), or even better, `condition`:

```suggestion
def mask_var(var: xr.DataArray, criteria):
```

Also, add type hints for arguments and result.

clouddrift/select.py
```diff
@@ -1,40 +1,41 @@
...
     """Return the mask of a subset of the data matching a test criterion.
 
     Args:
-        var (ak.Array): ak.Array
+        var (xr.DataArray): ak.Array
         value: tuple, list or scalar defining a test criterion
 
     Returns:
-        mask (Ak.Array): where values corresponding to the criterion are True
+        mask (xr.DataArray): where values corresponding to the criterion are True
```
milancurcic, line 14: Fix docstring format to NumPy style, and the description should explain how different condition types work.

clouddrift/select.py
```diff
@@ -1,40 +1,41 @@
...
     """
 
     if isinstance(value, tuple):  # min/max defining range
         mask = np.logical_and(var >= min(value), var <= max(value))
```
milancurcic, line 18: If we use `tuple` to mean expressing a range of values, I think we should constrain to a 2-tuple, and use `(min, max)` explicitly. I.e.:

```suggestion
        mask = np.logical_and(var >= value[0], var <= value[1])
```

Rationale:

1. If we let tuples have any number of values, it becomes easier for the user to confuse the `tuple` and the `list` and get two completely different behaviors.
2. There is a use-case in data analysis and modeling where a `(min, max)` pattern with `min > max` is used to return an empty subset. This is not possible with the current implementation.

clouddrift/select.py
```diff
@@ -1,40 +1,41 @@
...
         for v in value:
             mask = np.logical_or(mask, var == v)
     else:  # select one
         mask = var == value
     return mask
 
 
-def subset(ds: ak.Array, criteria: dict) -> ak.Array:
+def subset(ds: xr.Dataset, criteria: dict) -> xr.Dataset:
     """Subset the dataset as a function of one or many criteria. The criteria are passed as a dictionary, where
```
milancurcic, line 29: Docstring format;

Also we need to explain that this function expects `obs` and `traj` dimensions. That said, should we allow the user to specify some arbitrary dimension along which to subset? This would complicate the implementation though, so I'm not sure.

selipot:
So what's next before merge?

clouddrift/select.py
```diff
@@ -50,30 +51,34 @@ def subset(ds: ak.Array, criteria: dict) -> ak.Array:
...
-            mask_traj = np.logical_and(mask_traj, mask_var(ds[key], criteria[key]))
-        elif key in ds.obs.fields:
-            mask_obs = np.logical_and(mask_obs, mask_var(ds.obs[key], criteria[key]))
+        if key in ds:
+            if ds[key].dims == ("traj",):
+                mask_traj = np.logical_and(mask_traj, mask_var(ds[key], criteria[key]))
+            elif ds[key].dims == ("obs",):
+                mask_obs = np.logical_and(mask_obs, mask_var(ds[key], criteria[key]))
         else:
             raise ValueError(f"Unknown variable '{key}'.")
```
milancurcic, line 64: This errors out if there is a variable that does not have `traj` or `dims`. Should we warn and let it run through instead? It seems reasonable to me that a user could have some arbitrary data here using other dims.

clouddrift/select.py
```diff
@@ -50,30 +51,34 @@ def subset(ds: ak.Array, criteria: dict) -> ak.Array:
...
+    for i in np.where(~mask_traj)[0]:
+        mask_obs[slice(traj_idx[i], traj_idx[i + 1])] = False
+
+    # remove trajectory completely filtered in mask_obs
+    mask_traj = np.logical_and(
+        mask_traj, np.in1d(ds["ID"], np.unique(ds["ids"].isel({"obs": mask_obs})))
+    )
+
+    if not any(mask_traj):
+        warnings.warn("Empty set.")
```
milancurcic, line 77:
```suggestion
        warnings.warn("No data matches the criteria; returning an empty dataset.")
```

clouddrift/select.py
milancurcic: Never mind, I made a blunder, all good.

milancurcic, line 64:
```suggestion
            raise ValueError(f"Variable '{key}' not found in this dataset.")
```

philippemiron:
I don't mind moving this and the test to the analysis file.

milancurcic:
I'll try to resolve it now.

milancurcic (approved):
Good to go, thanks!

milancurcic:
The conflict was due to the #144 being merged into main before this, so I just had to merge main into this branch to resolve the conflict. There was also one small fix to the tests needed because of the switch of `preprocess_func` API to use `**kwargs`.

philippemiron:
Thanks!

PR #151, opened 2023-03-31
Implement `chunk`

### LINKED ISSUE #126: Implement `chunk`
JLab reference: [cellchunk](https://github.com/jonathanlilly/jLab/blob/master/jCell/cellchunk.m)

milancurcic:
In the TDD spirit, a few tests suggested by GH Copilot are still failing, but that aside this is ready for review.

Closes #126.

### DIFF
--- a/clouddrift/analysis.py
+++ b/clouddrift/analysis.py
@@ -100,21 +100,76 @@ def apply_ragged(
 
 
 def chunk(
-    x: Union[np.ndarray, xr.DataArray], length: int, overlap: int = 0
+    x: Union[list, np.ndarray, xr.DataArray],
+    length: int,
+    overlap: int = 0,
+    rowsize: Union[np.ndarray, xr.DataArray] = None,
 ) -> np.ndarray:
-    """Chunk the ragged array x.
+    """Chunk the ragged array ``x`` into equal-length chunks while respecting
+    the contiguous segments of the ragged array. The result is 2-dimensional
+    NumPy array of shape ``(num_chunks, length)``. The resulting number of chunks
+    is determined based on the length of ``x``, ``length``, ``overlap``, and
+    ``rowsize`` if provided.
+
+    Parameters
+    ----------
+    x : list or np.ndarray or xr.DataArray
+        A list or array-like to split into chunks.
+    length : int
+        The length of each chunk.
+    overlap : int, optional
+        The number of overlapping points between chunks. The default is 0.
+        Must be smaller than ``length``. For example, if ``length`` is 4 and
+        ``overlap`` is 2, the chunks of ``[0, 1, 2, 3, 4, 5]`` will be
+        ``np.array([[0, 1, 2, 3], [2, 3, 4, 5]])``.
+    rowsize : np.ndarray or xr.DataArray, optional
+        The size of rows if x is a ragged array. The default is None, in which
+        case x is assumed to be a 1-dimensional, non-ragged array.
+
+    Returns
+    -------
+    np.ndarray
+        2-dimensional array of shape ``(num_chunks, length)``.
+
+    Examples
+    --------
+
+    >>> chunk([1, 2, 3, 4, 5], 2)
+    array([[1., 2.],
+           [3., 4.]])
+
+    >>> chunk([1, 2, 3, 4, 5], 2, overlap=1)
+    array([[1., 2.],
+           [2., 3.],
+           [3., 4.],
+           [4., 5.]])
     """
-    num_chunks = len(x) // (length - overlap)
-    if overlap == 0:
-        return np.reshape(x[: num_chunks * length].values, (num_chunks, length))
-    elif 1 >= overlap < length:
-        res = np.zeros((num_chunks, length))
-        for n in range(num_chunks):
-            start = n * (length - overlap)
-            end = start + length
-            res[n] = x[start:end]
+
+    if rowsize is None:
+        num_chunks = [
+            (len(x) - length) // (length - overlap) + 1 if len(x) >= length else 0
+        ]
     else:
-        raise ValueError("Bad value for overlap")
+        num_chunks = [
+            (int(row_length) - length) // (length - overlap) + 1
+            if int(row_length) >= length
+            else 0
+            for row_length in rowsize
+        ]
+
+    res = np.empty((sum(num_chunks), length))
+
+    nchunk = 0
+    start = 0
+    for n in num_chunks:
+        # TODO Fix bug with start and end when we have varying rowsizes
+        for i in range(n):
+            end = start + length
+            res[nchunk] = x[start:end]
+            start = end - overlap
+            nchunk += 1
+
+    return res
 
 
 def segment(
--- a/clouddrift/analysis.py
+++ b/clouddrift/analysis.py
@@ -100,10 +100,10 @@ def apply_ragged(
 
 
 def chunk(
-    x: Union[list, np.ndarray, xr.DataArray],
+    x: Union[list, np.ndarray, xr.DataArray, pd.Series],
     length: int,
     overlap: int = 0,
-    rowsize: Union[np.ndarray, xr.DataArray] = None,
+    rowsize: Union[list, np.ndarray, xr.DataArray, pd.Series] = None,
 ) -> np.ndarray:
     """Chunk the ragged array ``x`` into equal-length chunks while respecting
     the contiguous segments of the ragged array. The result is 2-dimensional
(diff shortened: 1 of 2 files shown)


selipot:
Awesome, will test! What's TDD?

philippemiron:
Test driven development

milancurcic:
I think the ragged array is the main use case, but a non-ragged one is only a special case. Both seem to work.

philippemiron:
Ok, I guess it's simpler for a user then combining the function with `apply_ragged`.

milancurcic:
Initially I didn't think that `apply_ragged` would work here but on second thought I think it may. I'll compare the explicit use here and the one calling it with `apply_func` and if it works well we can just remove the `rowsize` from this PR.

philippemiron:
Sounds good!

milancurcic:
I compared the two approaches: running `chunk` with passing `rowsize` to it explicitly, and another where I pass `chunk` and `rowsize` to `apply_ragged` and let `apply_ragged` handle the ragged logic. I used 100 (well, ahem, 98) random drifters from the hourly GDP:

```python
In [1]: from clouddrift.adapters.gdp1h import to_raggedarray

In [2]: ds = to_raggedarray(n_random_id=100).to_xarray()
Downloading GDP hourly data to /tmp/clouddrift/gdp...
Downloading files: 100%|██████████████████| 100/100 [00:00<00:00, 222509.50it/s]
Retrieving the number of obs: 100%|█████████████| 98/98 [00:01<00:00, 49.62it/s]
/home/milan/.local/lib/python3.10/site-packages/xarray/coding/variables.py:147: FutureWarning: elementwise comparison failed; returning scalar instead, but in the future will perform elementwise comparison
  condition |= data == fv
Filling the Ragged Array: 100%|█████████████████| 98/98 [00:04<00:00, 22.45it/s]

In [3]: from clouddrift.analysis import apply_ragged, chunk
```

The output of `chunk` with `rowsize` directly:

```python
In [4]: chunk(ds.vn, 1000, rowsize=ds.rowsize)
Out[4]: 
array([[-0.1432, -0.1223, -0.055 , ..., -0.0648, -0.1092, -0.1271],
       [-0.1211, -0.1019, -0.0666, ..., -0.1052, -0.1046, -0.1046],
       [-0.1048, -0.1054, -0.1136, ..., -0.2873, -0.2815, -0.3201],
       ...,
       [ 0.2102,  0.239 ,  0.1996, ...,  0.5103,  0.494 ,  0.4204],
       [ 0.3241,  0.209 ,  0.1957, ...,  0.1575,  0.1275,  0.1221],
       [-0.1132, -0.1603, -0.1736, ..., -0.2045, -0.1538, -0.1137]],
      dtype=float32)
```

The output of `chunk` via `apply_ragged` is the same:

```python
In [5]: apply_ragged(chunk, [ds.vn], ds.rowsize, 1000)
Out[5]: 
array([[-0.1432, -0.1223, -0.055 , ..., -0.0648, -0.1092, -0.1271],
       [-0.1211, -0.1019, -0.0666, ..., -0.1052, -0.1046, -0.1046],
       [-0.1048, -0.1054, -0.1136, ..., -0.2873, -0.2815, -0.3201],
       ...,
       [ 0.2102,  0.239 ,  0.1996, ...,  0.5103,  0.494 ,  0.4204],
       [ 0.3241,  0.209 ,  0.1957, ...,  0.1575,  0.1275,  0.1221],
       [-0.1132, -0.1603, -0.1736, ..., -0.2045, -0.1538, -0.1137]],
      dtype=float32)

In [6]: np.all(apply_ragged(chunk, [ds.vn], ds.rowsize, 1000) == chunk(ds.vn, 1000, rowsize=ds.rowsize))
Out[6]: True
```

This means that `apply_ragged` works with `chunk`.

## API comparison

Say `x` is the array to chunk and `rowsize` defines its raggedness. The two APIs are then

```python
chunk(x, length, rowsize=rowsize)
```

and

```python
apply_ragged(chunk, [x], rowsize, length)
```

At a surface, the `apply_ragged` API is slightly more complex, but once the user learns how to use it, and knows what functions are compatible with it, it shouldn't be an issue IMO.

## Performance

```python
In [7]: %timeit chunk(ds.vn, 1000, rowsize=ds.rowsize)
77.1 ms ± 1.67 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [8]: %timeit apply_ragged(chunk, [ds.vn], ds.rowsize, 1000)
121 ms ± 3.4 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

The `apply_ragged` is about 50% slower in this specific case. We know that there is an extra copy in `apply_ragged` due to `np.concatenate` (no way around it) relative to the plain `chunk`, but I'm not sure if that could justify this large of a difference. Another thought is the threading. Let's try single-threaded:

```python
In [9]: %timeit apply_ragged(chunk, [ds.vn], ds.rowsize, 1000, max_workers=1)
85.1 ms ± 363 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

Much better and closer to plain `chunk` performance.

Another difference is that the current `chunk` implementation has some extra logic to handle `rowsize`. But if run with `apply_ragged`, that's not needed. So let's define a reduced `chunk2` function without `rowsize` to make the comparison more fair:

```python
def chunk2(
    x: Union[list, np.ndarray, xr.DataArray, pd.Series],
    length: int,
    overlap: int = 0,
) -> np.ndarray:


    num_chunks = (len(x) - length) // (length - overlap) + 1 if len(x) >= length else 0

    res = np.empty((num_chunks, length), dtype=np.array(x).dtype)

    start = 0
    for n in range(num_chunks):
        end = start + length
        res[n] = x[start:end]
        start = end - overlap

    return res
```

This shaves off a little bit:

```python
In [13]: %timeit apply_ragged(chunk2, [ds.vn], ds.rowsize, 1000, max_workers=1)
84.3 ms ± 611 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

(performance comparison when using non-zero `overlap` has similar results; not shown here).

## Preliminary takeaway

1. `apply_ragged` has slightly more complex API for the end-user, but not problematic.
2. `apply_ragged` is slightly less performant than plain `chunk` but not so much to justify not using it IMO.
3. Multithreading may be hurting performance rather than helping it. Perhaps we should consider using `max_workers=1` as the default in a separate PR, but more experiments are needed.

So, I think we should adopt `chunk2` above as the new `chunk` in this PR, and document well how to use it with `apply_ragged`.

philippemiron:
That's a great summary.

First, we have this in `apply_ragged`, so you don't have to pass as a list of you are only using one ragged array.

```
if type(arrays) not in [list, tuple]:
        arrays = [arrays]
```
still, as you said, it is slightly more complex using `ragged_array`.

```
apply_ragged(chunk, x, rowsize, length)
````

Another possibility could be to hide the `apply_ragged` from the user for those functions. We could have something like this:

```
def chunk(
    x: Union[list, np.ndarray, xr.DataArray, pd.Series],
    length: int,
    overlap: int = 0,
    rowsize: Union[list, np.ndarray, xr.DataArray, pd.Series] = None,
) -> np.ndarray:

    if rowsize is not None:
        apply_ragged(chunk, x, rowsize, length, overlap)

    num_chunks = (len(x) - length) // (length - overlap) + 1 if len(x) >= length else 0

    res = np.empty((num_chunks, length), dtype=np.array(x).dtype)

    start = 0
    for n in range(num_chunks):
        end = start + length
        res[n] = x[start:end]
        start = end - overlap

    return res
```

which is somehow a compromise. It does not require too much redundancy for the functions and makes it simpler to use for the user. And `apply_ragged` is still available if someone wants to use it for a user defined function.

milancurcic:
> ```
> if type(arrays) not in [list, tuple]:
>     arrays = [arrays]
> ```

That's great, I forgot about that.

Re: `chunk3`, I may be missing something, but to me this compromise takes the "worst of both worlds": we need to handle the `rowsize` argument in each function (so we can `apply_ragged`-it) and we keep the performance overhead of `apply_ragged` (even if only ~10% single-threaded).

If we `apply_ragged` consistently for ragged arrays operations, and keep the operation kernels (i.e. the operation on a contiguous segment) clean from ragged logic, and clearly document all this, I don't think it would bad at all for the user.

philippemiron:
Ok.

To be clear, my first option is `chunk2` (and simplify other functions in a similar fashion) and document `apply_ragged`. I was just wondering if there could be a way to make this easier for the user.

PS: I think the performance issues are mostly because you are using a quite small dataset.

milancurcic:
Thanks for the review!

PR #155, opened 2023-04-04
154 chunk vs cellchunk

### LINKED ISSUE #154: chunk vs cellchunk
The behavior of jlab cellchunk is such that:

_If there is a remainder R in how many times L goes into the length of
the time series, FLOOR(R/2) points will be thrown away from the
beginning of the time series and CEIL(R/2) points from the end._ 

but that behavior is not replicated in our `chunk`. We should discuss if this should be implemented.

philippemiron:
I think it's worth having this feature, so I just implemented it. Not sure if we need more tests to cover every usage..

Fix #154

### DIFF
--- a/clouddrift/analysis.py
+++ b/clouddrift/analysis.py
@@ -172,7 +172,11 @@ def chunk(
 
     res = np.empty((num_chunks, length), dtype=np.array(x).dtype)
 
-    start = 0 if discard_last else 1
+    if discard_last:
+        start = 0
+    else:
+        start = (len(x) + (num_chunks - 1) * overlap) - num_chunks * length
+
     for n in range(num_chunks):
         end = start + length
         res[n] = x[start:end]
--- a/clouddrift/analysis.py
+++ b/clouddrift/analysis.py
@@ -127,13 +127,13 @@ def chunk(
         to offset chunks by some number of elements. For example, if ``length``
         is 2 and ``overlap`` is -1, the chunks of ``[0, 1, 2, 3, 4, 5]`` will
         be ``np.array([[0, 1], [3, 4]])``.
-    keep : str, optional ["start", "middle", "end"]
+    align : str, optional ["start", "middle", "end"]
         If the number of chunks (including or not overlap) is not a multiple of the
         length of ``x`` and there is a reminder of N points, this parameter controls
-        which part of the array will be kept into the chunks. If ``keep="start"``, the
+        which part of the array will be kept into the chunks. If ``align="start"``, the
         points at the beginning of the array will be kept, and N points are discarded at
-        the end. If `keep="middle"`, floor(N/2) and ceil(N/2) points will be respectively
-        discarded from the beginning and the end of the array. If ``keep="end"``, the
+        the end. If `align="middle"`, floor(N/2) and ceil(N/2) points will be respectively
+        discarded from the beginning and the end of the array. If ``align="end"``, the
         points at the end of the array will be kept, and the `N` first points are discarded.
         The default is "start".
 
--- a/clouddrift/analysis.py
+++ b/clouddrift/analysis.py
@@ -103,7 +103,7 @@ def chunk(
     x: Union[list, np.ndarray, xr.DataArray, pd.Series],
     length: int,
     overlap: int = 0,
-    discard_last: bool = True,
+    keep: str = "mid",
 ) -> np.ndarray:
     """Chunk an array ``x`` into equal-length chunks while respecting
     the contiguous segments of the ragged array. The result is 2-dimensional
@@ -127,11 +127,15 @@ def chunk(
         to offset chunks by some number of elements. For example, if ``length``
         is 2 and ``overlap`` is -1, the chunks of ``[0, 1, 2, 3, 4, 5]`` will
         be ``np.array([[0, 1], [3, 4]])``.
-    discard_last : bool, optional
+    keep : str, optional ["start", "mid", "last"]
         If the number of chunks (including or not overlap) is not a multiple of the
-        length of ``x``, the points at the end of the array will be thrown away.
-        If ``discard_last`` is False, the points at the beginning of the array are
-        thrown away. The default is True.
+        length of ``x`` and there is a reminder of N points, this parameter controls
+        which part of the array will be kept into the chunks. If `keep="mid"`, floor(N/2)
+        ceil(N/2) points will be respectively discarded from the beginning and the end of
+        the array. If ``keep="start"``, the points at the beginning of the array will
+        be kept, and N points are discarded at the end. If ``keep="last"``, the points at
+        the end of the array will be kept, and the `N` first points are discarded.
+        The default is "mid".
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


clouddrift/analysis.py
```diff
@@ -126,6 +127,10 @@ def chunk(
         to offset chunks by some number of elements. For example, if ``length``
         is 2 and ``overlap`` is -1, the chunks of ``[0, 1, 2, 3, 4, 5]`` will
         be ``np.array([[0, 1], [3, 4]])``.
+    discard_last : bool, optional
+        If the number of chunks is not a multiple of the length of ``x``, the
+        points at the end will be thrown away. If ``skip_last`` is False, the
```
selipot: do you mean if ``discard_last`` is False?

philippemiron: Yes, I just fixed it. That was my previous variable name. Also open to any other parameter name by the way!

selipot:
So that's an interesting feature, but not exactly like cellchunk which throws away half at the beginning and half at the end. As stated:

_If there is a remainder R in how many times L goes into the length of
the time series, FLOOR(R/2) points will be thrown away from the
beginning of the time series and CEIL(R/2) points from the end._

Not sure what is best.

selipot:
Right now am working with cellchunk in Matlab and throwing away half at the beginning and half at the end is sort-of desirable because I am filtering the time series and then chunking them. Hence throwing away both ends minimize edge effects from filtering (but not always if my the trajectory length is exactly an integer number of the chunk length).

philippemiron:
Oh, I did not understand it properly then... I guess it could be modified to have 3 possibilities.
- start from the beginning of the array
- finish at the end of the array
- in-between case where you start at floor(R/2)

selipot:
Yes, we should have three possibilities with the default behavior being discarding half and half at beginning and end.

philippemiron:
I think this should work as expected.

philippemiron:
Wouldn't it be more logical to start from the beginning by default?

selipot:
Not sure what that logic is? I like trimming equally both ends. And it's always an option so users can decide.

milancurcic:
From a software design and the least-surprise principle point of view, the best default for a function argument is the one that doesn't introduce additional behavior. Additional behavior here is to shift the chunks relative to the start of the array. In that sense, the most logical default behavior is to not shift the array, as Philippe wrote.

On the other side, "I like trimming equally both ends" is difficult to argue against.

Whatever you decide, the default behavior should be clearly described in the main description of the function, and the additional behavior via optional args should be described in the args description.

clouddrift/analysis.py
```diff
@@ -103,6 +103,7 @@ def chunk(
     x: Union[list, np.ndarray, xr.DataArray, pd.Series],
     length: int,
     overlap: int = 0,
+    keep: str = "mid",
```
milancurcic, line 106: I think the name `keep` is confusing here. To keep the points implies that the purpose of the function is to discard points. However, discarding is merely a consequence of the input not matching the size of the output.

What do you think about naming this argument `shift` or `align`, with possible options being "left", "center", "right"?

clouddrift/analysis.py
```diff
@@ -164,10 +174,16 @@ def chunk(
         if ``length == 0``.
     """
     num_chunks = (len(x) - length) // (length - overlap) + 1 if len(x) >= length else 0
-
+    reminder = len(x) - num_chunks * length + (num_chunks - 1) * overlap
```
milancurcic, line 177: `remainder`?

clouddrift/analysis.py
```diff
@@ -164,10 +174,16 @@ def chunk(
...
     res = np.empty((num_chunks, length), dtype=np.array(x).dtype)
 
-    start = 0
+    if keep == "start":
+        start = 0
+    elif keep == "end":
+        start = reminder
+    else:
+        start = np.floor(reminder / 2).astype("int")
```
milancurcic, line 185:
```suggestion
        start = remainder // 2
```

clouddrift/analysis.py
milancurcic: Alternatively, `discard` or `trim` (meaning, opposite of `keep`) would work well too.

milancurcic (approved):
Pending comment resolution.

clouddrift/analysis.py
philippemiron: Haha yes.

clouddrift/analysis.py
philippemiron: Actually `keep=mid` does keeps the middle point (discard if needed start and end). Similarly, with `keep=start`, chunks will start at the beginning.

milancurcic: Because of the align to the middle behavior, I realize my second suggestion (`discard` or `trim`) doesn't work well.

I understand the meaning of `keep`, and it makes sense when I think through it. I just personally found it a bit more difficult to understand. Because the function is called `chunk`, and it chunks an input array into a 2-d array of equal-length chunks, the argument `keep` I couldn't help but ask "keep what?". It's not "keep the chunk", but something else. Then I had to read the description. If we call this `shift` or `align`, the verb naturally addresses the chunks, so I find it a bit more intuitive. Maybe it's just my brain.

philippemiron:
My logic was : if you have a array of length 8 and ask someone to split into 2 chunks of 3. My guess is that they would start at the beginning.

Then, they might wonder what do to with the last two numbers, but to me this is a secondary step..

clouddrift/analysis.py
selipot: I think that `keep` and `align` are both good but `left` and `right` as possible options not so much. I visualize a ragged array as being a column vector, not a row vector, but others may see it differently. As such, I prefer `(start,[mid,center],end)` to `(left,center,right)`.

milancurcic:
> My logic was : if you have a array of length 8 and ask someone to split into 2 chunks of 3 without overlapping. > My guess is that they would start at the beginning.
>
> Then, they might wonder what do to with the last two numbers, but to me this is a secondary step..

Exactly my thoughts as well..

clouddrift/analysis.py
milancurcic: That's a good point. I suggest `mid` -> `middle` for consistency with `start` and `end` (they are not abbreviated).

philippemiron: and we use "keep" or "align"?

selipot: The goal of the function is to create the chunks so let's make this option about the output, i.e. how do we want the output to be aligned with respect to the input. So let's use `align`.

selipot:
> > My logic was : if you have a array of length 8 and ask someone to split into 2 chunks of 3 without overlapping. > My guess is that they would start at the beginning.
> > Then, they might wonder what do to with the last two numbers, but to me this is a secondary step..
> 
> Exactly my thoughts as well..

I do agree as well, but I do not think it is a matter of logic. In fact I spent 20 min trying to understand a cellchunk result in Matlab until I read carefully the help and realized it mentioned the default behavior being `align="middle"`, a clearly not default expected behavior. So yes, I agree that default should be `align="start"`.

PR #164, opened 2023-04-13
Expand the OS and Python version matrix in CI

milancurcic:
* Linux, macOS, and Windows
* Python 3.9, 3.10, and 3.11
* Coverage report upload only runs from Linux+Python 3.9 job.

The CI time does go from about a minute to a few minutes, but I think it's worth it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,4 +1,4 @@
-name: test
+name: CI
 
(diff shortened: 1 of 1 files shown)


philippemiron (commented):
Looks good.

PS: Microsoft can afford to run those tests 9x more.

PR #169, opened 2023-04-28
Handle exception in rowsize func

milancurcic:
Occasionally an individual drifter NetCDF file may be corrupt.

Rather than failing, the script will continue processing and set that drifter's rowsize value to 0.

Needs to be tested, I'm not able to do that yet as my laptop is too small for the dataset.


milancurcic:
Now that I'm on a bigger computer, the upstream dataset has been fixed so I can't reproduce the problem. Let's merge nevertheless and address it later if any issues come up.

PR #186, opened 2023-06-02
Implement `plane_to_sphere`

milancurcic:


clouddrift/sphere.py
```diff
@@ -105,6 +105,66 @@ def recast_lon180(lon: np.ndarray) -> np.ndarray:
...
+    Raises
+    ------
+    AttributeError
+        If ``x`` and ``y`` are not NumPy arrays
+
+    See Also
+    --------
+    :func:`sphere_to_plane`
+    """
+    lon = np.empty(x.shape, dtype=np.float64)
```
philippemiron, line 148: You could use `np.empty_like` to keep the type of the input.

clouddrift/sphere.py
```diff
@@ -143,16 +203,20 @@ def sphere_to_plane(
...
     ------
-    TypeError
+    AttributeError
         If ``lon`` and ``lat`` are not NumPy arrays
+
+    See Also
+    --------
+    :func:`plane_to_sphere`
     """
     x = np.empty(lon.shape, dtype=np.float64)
```
philippemiron, line 221: Same here

philippemiron (approved):

clouddrift/sphere.py
```diff
@@ -105,6 +105,66 @@ def recast_lon180(lon: np.ndarray) -> np.ndarray:
...
+    ------
+    AttributeError
+        If ``x`` and ``y`` are not NumPy arrays
+
+    See Also
+    --------
+    :func:`sphere_to_plane`
+    """
+    lon = np.empty(x.shape, dtype=np.float64)
+    lat = np.empty(y.shape, dtype=np.float64)
```
philippemiron, line 149:
```suggestion
    lon = np.empty_like(x)
    lat = np.empty_like(y)
```

clouddrift/sphere.py
```diff
@@ -143,16 +203,20 @@ def sphere_to_plane(
...
-    TypeError
+    AttributeError
         If ``lon`` and ``lat`` are not NumPy arrays
+
+    See Also
+    --------
+    :func:`plane_to_sphere`
     """
     x = np.empty(lon.shape, dtype=np.float64)
     y = np.empty(lat.shape, dtype=np.float64)
```
philippemiron, line 222:
```suggestion
     x = np.empty_like(lon)
     y = np.empty_like(lat)
```

philippemiron (approved):

philippemiron:
By default in python (and most people use like this since you don't have to define type of course) everything is in `float64`. 

The reason my suggestion would be to keep `type` is that if it's `float32`, it's most likely because the user switched the type to save space, for compatibility with another app (pytorch for example), etc., so switching to 64 bits might create confusion having inputs/outputs of different types.

We can add a note that round-trip of position on the Earth are not exact if precision is decreased. But at the same time, this is almost always the case for any function. 

For example:

milancurcic:
Thanks for the review.

PR #193, opened 2023-06-22
Add module docstrings and reformat remaining old-format docstrings

### LINKED ISSUE #140: Fix docstrings format
This issue will track re-formatting all docstrings to [NumPy style](https://numpydoc.readthedocs.io/en/latest/format.html).

* [x] adapters
* [x] analysis
* [x] datasets
* [x] raggedarray
* [x] haversine
* [x] sphere

milancurcic:
Closes #140.

### DIFF
--- a/clouddrift/analysis.py
+++ b/clouddrift/analysis.py
@@ -1,5 +1,5 @@
 """
-This module provides the common Lagrangian analysis and transformation
+This module provides common Lagrangian analysis and transformation
 functions.
 """
 
--- a/clouddrift/analysis.py
+++ b/clouddrift/analysis.py
@@ -1,3 +1,8 @@
+"""
+This module provides the common Lagrangian analysis and transformation
+functions.
+"""
+
 import numpy as np
 from typing import Optional, Tuple, Union, Iterable
 import xarray as xr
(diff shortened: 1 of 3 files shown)


clouddrift/haversine.py
```diff
@@ -1,3 +1,9 @@
+"""
+This module provides various Haversine functions to compute distances and
```
selipot, line 2: kinda makes me think that this module could be folded in the sphere module?

selipot: but not now

clouddrift/analysis.py
```diff
@@ -1,3 +1,8 @@
+"""
+This module provides the common Lagrangian analysis and transformation
```
selipot, line 2: remove "the"

selipot (changes requested):

clouddrift/haversine.py
milancurcic: Yes, I think that would be appropriate.

clouddrift/analysis.py
milancurcic, line 2:
```suggestion
This module provides common Lagrangian analysis and transformation
```

Issue #180, opened 2023-05-19
How to run CloudDrift's analysis functions in parallel?


milancurcic:
* Dask
* ProcessPoolExecutor?
* Inside `apply_ragged` or some other way?

selipot:
Let's keep discussing this one and propose solutions if any.

philippemiron:
I took some time to read on `concurrent.futures` and `dask.futures`, and it looks like it's basically a drop in replacement!

I believe it shouldn't be too hard to use `dask` in `apply_ragged`. I'll try to find some time to test it out over the next weekend/week.

See, https://docs.dask.org/en/stable/futures.html.

PS: I also want to see if we can move away from threads in the default case and use processes.

philippemiron:
So with the latest update to `apply_ragged`.

I tried the following:

```
import clouddrift as cd
from clouddrift.analysis import position_from_velocity, apply_ragged

from dask.distributed import Client
client = Client("tcp://127.0.0.1:62864")
client

#ds = cd.datasets.gdp1h().load()
import xarray as xr
ds = xr.open_dataset("gdp_v2.00.nc")

u, v = apply_ragged(
    position_from_velocity,
    [ds.ve, ds.vn, ds.time],
    ds.rowsize,
    0,
    0,
    coord_system="cartesian",
    integration_scheme="centered",
    executor=client, # passing the Dask client here
)
```
and it does seem to be processing the trajectories in parallel. 

![image](https://github.com/Cloud-Drift/clouddrift/assets/387624/7a471129-3239-4196-8368-af7a74d4b424)

AS you can see on the next screenshot, it's creating 17324 tasks... it might be more efficient to be able to run a group of trajectories at the same time, but I guess that's a first step.

![image](https://github.com/Cloud-Drift/clouddrift/assets/387624/c78d8fbc-99bc-4094-bf4c-0727a2956ef6)

PS: It doesn't start when I used the AWS source.

selipot:
I think that's great! So did it complete?

philippemiron:
Yes, it completed fine.

philippemiron:
FYI, just ran this again today as a test after talking with someone at COAPS.

With the data loaded in memory:

```
ds = cd.datasets.gdp1h().load()
```

The following:
```
%%time
x, y = apply_ragged(
    position_from_velocity,
    [ds.ve, ds.vn, ds.time],
    ds.rowsize,
    0,
    0,
    coord_system="cartesian",
[...]

PR #198, opened 2023-06-24
option to use processes

philippemiron:
- ProcessPoolExecutor option in `ragged_array` (Discussed in #180).


philippemiron:
There is some limitation with the Processes. The function can not be defined in the same scope.

For example this would not work:
```
def local_mean(x):
   return np.mean(x)

apply_ragged(local_mean, ds.lon.values, ds.rowsize.values, use_threads=False)
```

but if the function is defined in another module, you can use processes.

```
apply_ragged(np.mean, ds.lon.values, ds.rowsize.values, use_threads=False)
```

By default, I guess we can leave threads, but I think it's a good option to have for I/O bound tasks.

clouddrift/analysis.py
```diff
@@ -81,14 +82,16 @@ def apply_ragged(
...
+    if use_threads:
+        executor_type = futures.ThreadPoolExecutor
+    else:
+        executor_type = futures.ProcessPoolExecutor
 
     # parallel execution
-    with futures.ThreadPoolExecutor(max_workers=max_workers) as executor:
-        res = executor.map(lambda x: func(*x, **kwargs), iter)
+    with executor_type(max_workers=max_workers) as executor:
+        res = [executor.submit(func, *x, *args, **kwargs).result() for x in iter]
```
philippemiron, line 93: I also switch to submit, which in my opinion, simplifies the syntax since we have `*args` and `**kwargs` to pass to the function.

philippemiron:
I haven't tested it just yet, but it should be possible to pass a `dask.distributed.Client()` to `apply_ragged` instead and use it as an `executor`.

selipot:
I'd like to follow but I am not. Could you summarize this PR in non technical terms? And provide an example with the GDP as an example?

philippemiron:
- process: use a defined memory space
- threads: run inside one process, so they share the memory

Because of this and some design decisions of Python, only one thread is executed at a time. So threads can be helpful in some cases, where for example when there is a lot of waiting involved (reading from files, request from website, etc.). When each thread must perform intensive CPU calculations, this is not efficient and will most likely be longer than running on a single thread.

So on the other hand, processes have a larger overhead, so they require more time to create, but they are efficient to perform parallel calculations. Problem is that we don't have functions that take a LONG time so the results are not that interesting.

But for example with `position_from_velocity`:

```
u, v = apply_ragged(
    position_from_velocity,
    [ds.ve, ds.vn, ds.time],
    ds.rowsize,
    0,
    0,
    coord_system="cartesian",
    integration_scheme="centered",
    use_threads=False,
)
```
you can see that it working in parallel on my computer.

image

selipot:
So with the PR `apply_ragged` uses *Processes*. And before it used only *Threads*, is this correct? Still don't get it but I trust you it is better!

philippemiron:
This PR adds a parameter (`use_threads: bool = True`) that you can set to `True/False` to use either threads or processes.

philippemiron:
Maybe @milancurcic can explain it better; sorry for the poor explanation.

milancurcic:
Thanks for this PR. It seems like a straightforward change.

Rather than having a `use_threads: bool = True`, what do you think about having the user pass the executor that they want to use? The argument for this is that `use_threads = False` means that we won't use threads, but it does not imply that we will use processes instead. So I suggest:

```python
def apply_ragged(
    func: callable,
    arrays: list[np.ndarray],
    rowsize: list[int],
    *args: tuple,
    executor: Executor = ThreaPoolExecutor,
    max_workers: int = None,
    **kwargs: dict,
) -> Union[tuple[np.ndarray], np.ndarray]:
```

Then, if we find an executor that works as a drop-in replacement, perhaps from Dask or somewhere else, then the user can already use it as is rather than ask us to change the function.

@selipot I'll take some time to review and test this, please don't rush to merge this in the middle of the meeting.

selipot:
It isn't my style to rush anything 😝

philippemiron:
> 

Actually much better. And yes, the `dask` executor should work that way easily.

philippemiron:
It might actually be better to pass directly: `Executor = ThreaPoolExecutor(max_workers=None)`. That way, we can also pass a `dask.Client()`, which follows the same API and can perform `client.submit()`.

philippemiron:
Let me know what you think @milancurcic.

milancurcic (approved):
Looks good, thank you!

selipot:
We need concrete example in the doc or in clouddrift-example with the GDP datasets for this. I for example don;t know how to get started with this.

PR #205, opened 2023-06-26
Haversine to sphere

philippemiron:


clouddrift/__init__.py
```diff
@@ -5,6 +5,4 @@
 from clouddrift.raggedarray import RaggedArray
 import clouddrift.adapters
 import clouddrift.analysis
-import clouddrift.datasets
```
milancurcic, line 8: Why remove `datasets` import here?

milancurcic:
We'll also need a minor version bump here (API change), so v0.17.0.

clouddrift/__init__.py
philippemiron: That's a merge error.

PR #212, opened 2023-07-03
Rename `rowsize` to `count`

### LINKED ISSUE #197: Should `rowsize` be renamed to `count`?
It occurred to me that in the first sentence of the [Usage page in docs](https://cloud-drift.github.io/clouddrift/usage.html) we link to [a section in CF convention](https://cfconventions.org/cf-conventions/cf-conventions.html#_contiguous_ragged_array_representation) about contiguous ragged arrays. This sections states:

> In this representation, the file contains a count variable, which must be an integer type and
>
> | count(i1) | count(i2) | count(i3) | count(i4) |
> |---|---|---|---|
> | 2 | 4 | 3 | 6 | 
>
> must have the instance dimension as its sole dimension. The count variable contains the number of elements that each feature has. This representation and its count variable are identifiable by the presence of an attribute, sample_dimension, found on the count variable, which names the sample dimension being counted. For indices that correspond to features, whose data have not yet been written, the count variable should  have a value of zero or a missing value.

I don't think that `count` is any better than `rowsize`. However, I do think that following some standard is better than following no standard, and CF is at least to an extent respected in the gridded weather, ocean, and climate world. So it seems to me that if we want to follow CF more closely, we should:

1. Rename `rowsize` to `count`
2. `count` variable should have an attribute `sample_dimension`, which in our case would be `obs`.

This kind of change is uncomfortable to make, but if we decide that it should be made, it would be the sooner the better.

What do you think?

milancurcic:
Closes #197.

Version bump to 0.19.0.

### DIFF
--- a/clouddrift/adapters/gdp1h.py
+++ b/clouddrift/adapters/gdp1h.py
@@ -484,7 +484,7 @@ def preprocess(index: int, **kwargs) -> xr.Dataset:
     ds.attrs = attrs
 
     # rename variables
-    ds = ds.rename_vars({"longitude": "lon", "latitude": "lat"})
+    ds = ds.rename_vars({"longitude": "lon", "latitude": "lat", "rowsize": "count"})
 
     return ds
 
@@ -566,7 +566,7 @@ def to_raggedarray(
         name_coords=gdp.GDP_COORDS,
         name_meta=gdp.GDP_METADATA,
         name_data=GDP_DATA,
-        rowsize_func=gdp.rowsize,
+        count_func=gdp.count,
         filename_pattern=filename_pattern,
         tmp_path=GDP_TMP_PATH,
     )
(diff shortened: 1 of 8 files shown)


clouddrift/adapters/gdp1h.py
```diff
@@ -484,7 +484,7 @@ def preprocess(index: int, **kwargs) -> xr.Dataset:
     ds.attrs = attrs
 
     # rename variables
-    ds = ds.rename_vars({"longitude": "lon", "latitude": "lat"})
+    ds = ds.rename_vars({"longitude": "lon", "latitude": "lat", "rowsize": "count"})
```
philippemiron, line 487: maybe put:

```
if "rowsize" in ds.variables():
    ds = ds.rename_vars({"rowsize": "count"})
else:
    print("Remove now that GPD was updated")
```

PR #213, opened 2023-07-07
Begin reworking the Usage page in the docs

### LINKED ISSUE #122: Should clouddrift adopt a shorthand `cd`?
```python
import numpy as np
import xarray as xr
import clouddrift as cd
```

Through examples, we could guide the user on how to use clouddrift. One thing that we could show/recommend is to import the `clouddrift` as `cd`, like `np`, `xr`, or `pd`.

Has `cd` been claimed by any library? I'm not aware of it.

One advantage of this style is to provide a single, short import for `clouddrift` and a 2-letter namespace for users to not have to explicitly import stuff by name.

milancurcic:
I started reworking the Usage page of the docs. It will show how to access available ragged-array datasets, how to run some analysis functions and what's available, and how can custom datasets be adapted to a ragged-array structure.

Closes #122.

### DIFF
--- a/clouddrift/analysis.py
+++ b/clouddrift/analysis.py
@@ -977,6 +977,18 @@ def subset(ds: xr.Dataset, criteria: dict) -> xr.Dataset:
     ValueError
         If one of the variable in a criterion is not found in the Dataset
     """
+    # Normally we expect the ragged-array dataset to have a "count" variable.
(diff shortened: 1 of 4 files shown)


milancurcic:
I also changed the theme to sphinx-book-theme in this PR because it was easy. Let's add graphics in a separate PR.

PR #218, opened 2023-07-28
initial commit for analytic transform function

selipot:
Draft for adding the analytic (rotary) transform function to the library in a new module called analytic. Let's discuss refinements such as extension to ND arrays? Also, I have not tested all input types: should this be added to the tests?


milancurcic:
Great! Yes, when feasible I try to test for different input types.

milancurcic:
`analytic` is to similar to the `analysis` module name.

Does this fit in `analysis`?

If not, how about naming this module `signal`?

selipot:
> `analytic` is too similar to the `analysis` module name.
> 
> Does this fit in `analysis`?
> 
> If not, how about naming this module `signal`?

Yes I realize that. `signal` could be good. or `rotary` since the applications for Lagrangian bivariate data is very much about extracting rotary components.

clouddrift/analytic.py
```diff
@@ -0,0 +1,80 @@
...
+    >>> zp = anatrans(z)
+    >>> zn = anatrans(np.conj(z))
+
+    Raises
+    ------
+    ValueError
+        If ``boundary not in ["mirror", "zeros", "periodic"]``.
+    """
+    # assume unidimensional input; add dimension option
+    m0 = len(x)
```
milancurcic, line 47: `m0` is not used until way down. Move this closer to where it's used (L76).

clouddrift/analytic.py
```diff
@@ -0,0 +1,80 @@
...
+
+    # remove mean
+    x = x - np.mean(x)
```
milancurcic, line 50: Both work but this form is more Pythonic.

```suggestion
    x -= np.mean(x)
```

clouddrift/analytic.py
```diff
@@ -0,0 +1,80 @@
...
+
+    To obtain the analytic part of a real-valued signal
+    >>> x = np.random.rand(99)
+    >>> z = anatrans(x)
+
+    To obtain the analytic and anti-analytic parts of a complex-valued signal
+    >>> z = np.random.rand(99)+1j*np.random.rand(99)
+    >>> zp = anatrans(z)
+    >>> zn = anatrans(np.conj(z))
+
```
milancurcic, line 40: An example with the `boundary` parameter would be good.

milancurcic (approved):

clouddrift/analytic.py
```diff
@@ -0,0 +1,80 @@
...
+"""
+
+import numpy as np
+from typing import Optional, Tuple, Union
+import xarray as xr
+import pandas as pd
+import warnings
+
+
+def anatrans(
```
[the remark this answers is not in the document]
selipot: Sure

clouddrift/analytic.py
selipot: I don't think so. The original length need to be recorded before the boundary condition is applied.

PR #219, opened 2023-08-01
signal module

selipot:
PR to replace the anatrans PR. Changed the name of module to signal and implemented Milan's suggestions


selipot:
Should it be expanded to multi-dimensional arrays following velocity_from_position as an example?

milancurcic:
I think yes if it's a low-hanging fruit. `np.fft.fft` accepts `axis` to specify the axis on which to run the FFT, and slices can use the `[..., :]` syntax to indicate any number of slow-moving dimensions.

selipot:
> I think yes if it's a low-hanging fruit. `np.fft.fft` accepts `axis` to specify the axis on which to run the FFT, and slices can use the `[..., :]` syntax to indicate any number of slow-moving dimensions.

Should I work on the multi-dimensional option in a subsequent PR?

milancurcic (approved):
I think this can go ahead, pending the addition of `signal` to the API docs, and also importing it in `clouddrift/__init__.py`, like the other submodules.

selipot:
Anything else? Some version bump?

PR #225, opened 2023-08-02
rotary transform function

selipot:
PR for adding rotary_transform function to derive rotary components of bivariate time series

### DIFF
--- a/clouddrift/signal.py
+++ b/clouddrift/signal.py
@@ -86,3 +86,70 @@ def analytic_transform(
     z = z[int(m0 + 1) - 1 : int(2 * m0 + 1) - 1]
 
     return z
+
+
+def rotary_transform(
+    u: Union[list, np.ndarray, xr.DataArray, pd.Series],
+    v: Union[list, np.ndarray, xr.DataArray, pd.Series],
+    boundary: Optional[str] = "mirror",
+) -> Tuple[np.ndarray, np.ndarray]:
+    """Returns time-domain rotary components time series (zp,zn) from Cartesian components time series (u,v).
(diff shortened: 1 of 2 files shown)


clouddrift/signal.py
```diff
@@ -86,3 +86,70 @@ def analytic_transform(
...
 
     return z
+
+
+def rotary_transform(
+    u: Union[list, np.ndarray, xr.DataArray, pd.Series],
+    v: Union[list, np.ndarray, xr.DataArray, pd.Series],
+    boundary: Optional[str] = "mirror",
+) -> Tuple[np.ndarray, np.ndarray]:
+    """Returns time-domain rotary components time series (zp,zn) from Cartesian components time series (u,v).
```
milancurcic, line 96:
```suggestion
    """Return time-domain rotary components time series (zp,zn) from Cartesian components time series (u,v).
```

clouddrift/signal.py
```diff
@@ -86,3 +86,70 @@ def analytic_transform(
...
+    # u and v arrays must have the same shape or list length.
+    if type(u) == list and type(v) == list:
+        if not len(u) == len(v):
+            raise ValueError("u and v must have the same length.")
+    else:
+        if not u.shape == v.shape:
+            raise ValueError("u and v must have the same shape.")
+
+    # convert to numpy in case input is xarray DataArray
+    if type(u) == xr.DataArray:
```
philippemiron, line 147: Maybe this could be done after: 

```
if muv == xr.DataArray:
   muv == muv.to_numpy()  (or muv.values())
```

so the main calculation is only once.

philippemiron (approved):

PR #228, opened 2023-08-11
Nd-signal

selipot:
extension of signal transforms to nd arrays. Test left to write.

### DIFF
--- a/clouddrift/signal.py
+++ b/clouddrift/signal.py
@@ -43,9 +43,14 @@ def analytic_transform(
     >>> zp = analytic_transform(z)
     >>> zn = analytic_transform(np.conj(z))
 
-    To specify that a periodic boundary condition should be used
+    To specify that a periodic boundary condition should be used:
     >>> x = np.random.rand(99)
     >>> z = analytic_transform(x,boundary="periodic")
+
+    To specify that the time axis is along the first axis, and apply
+    zero boundary conditions:
+    >>> x = np.random.rand(100,99)
+    >>> z = analytic_transform(x,time_axis=0,boundary="zeros")
 
     Raises
     ------
@@ -134,6 +139,8 @@ def rotary_transform(
     boundary : str, optional ["mirror", "zeros", "periodic"] optionally specifies the
     boundary condition to be imposed at the edges of the time series for the underlying analytic
     transform. Default is "mirror".
+    time_axis : int, optional)
+        Axis along which time is (default is -1)
 
     Returns
     -------
@@ -146,7 +153,15 @@ def rotary_transform(
     --------
 
     To obtain the rotary components of a real-valued signal:
+    >>> u = np.random.rand(99)
+    >>> v = np.random.rand(99)
     >>> zp, zn = rotary_transform(u,v)
+
+    To specify that the time axis is along the first axis, and apply
+    zero boundary conditions:
+    >>> u = np.random.rand(100,99)
+    >>> v = np.random.rand(100,99)
+    >>> zp, zn = rotary_transform(u,v,time_axis=0,boundary="zeros")
 
     Raises
     ------
--- a/clouddrift/signal.py
+++ b/clouddrift/signal.py
@@ -11,6 +11,7 @@ import warnings
 def analytic_transform(
     x: Union[np.ndarray, xr.DataArray],
     boundary: Optional[str] = "mirror",
+    time_axis: Optional[int] = -1,
 ) -> np.ndarray:
     """Return the analytic part of a real-valued signal or of a complex-valued
     signal. To obtain the anti-analytic part of a complex-valued signal apply analytic_transform
@@ -49,6 +50,13 @@ def analytic_transform(
     ValueError
         If ``boundary not in ["mirror", "zeros", "periodic"]``.
     """
+    # time_axis must be in valid range
+    if time_axis < -1 or time_axis > len(u.shape) - 1:
+        raise ValueError(
+            f"time_axis ({time_axis}) is outside of the valid range ([-1,"
+            f" {len(x.shape) - 1}])."
+        )
+    # to continue following Milan's code
     # assume unidimensional input; add dimension option
     m0 = len(x)
 
--- a/clouddrift/signal.py
+++ b/clouddrift/signal.py
@@ -9,7 +9,7 @@ import warnings
 
 
 def analytic_transform(
-    x: Union[list, np.ndarray, xr.DataArray],
+    x: Union[np.ndarray, xr.DataArray],
     boundary: Optional[str] = "mirror",
 ) -> np.ndarray:
     """Return the analytic part of a real-valued signal or of a complex-valued
@@ -88,8 +88,8 @@ def analytic_transform(
 
 
 def rotary_transform(
-    u: Union[list, np.ndarray, xr.DataArray],
-    v: Union[list, np.ndarray, xr.DataArray],
+    u: Union[np.ndarray, xr.DataArray],
+    v: Union[np.ndarray, xr.DataArray],
     boundary: Optional[str] = "mirror",
 ) -> Tuple[np.ndarray, np.ndarray]:
     """Return time-domain rotary components time series (zp,zn) from Cartesian components time series (u,v).
@@ -134,13 +134,9 @@ def rotary_transform(
     :func:`analytic_transform`
     """
     # to implement: input one complex argument instead of two real arguments
-    # u and v arrays must have the same shape or list length.
-    if type(u) == list and type(v) == list:
-        if not len(u) == len(v):
-            raise ValueError("u and v must have the same length.")
-    else:
-        if not u.shape == v.shape:
-            raise ValueError("u and v must have the same shape.")
+    # u and v arrays must have the same shape.
+    if not u.shape == v.shape:
+        raise ValueError("u and v must have the same shape.")
 
     muv = np.mean(u) + 1j * np.mean(v)
 
--- a/clouddrift/signal.py
+++ b/clouddrift/signal.py
@@ -78,26 +78,26 @@ def analytic_transform(
 
     # Subtract mean along time axis (-1); convert to np.array for compatibility
     # with xarray.DataArray.
-    x_ -= np.array(np.mean(x_, axis=-1, keepdims=True))
+    xa = x_ - np.array(np.mean(x_, axis=-1, keepdims=True))
 
     # apply boundary conditions
     if boundary == "mirror":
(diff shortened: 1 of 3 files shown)


clouddrift/signal.py
```diff
@@ -41,57 +43,84 @@ def analytic_transform(
...
+        raise ValueError(
+            f"time_axis ({time_axis}) is outside of the valid range ([-1,"
+            f" {len(x.shape) - 1}])."
+        )
+
+    # Reshape the inputs to ensure the time axis is last (fast-varying)
+    # should we add a condition to skip if time_axis=-1 or time_axis == len(x.shape)-1
+    # do we need to copy x here? does it matter for memory and performance?
+    if time_axis != -1 and time_axis != len(x.shape) - 1:
+        x_ = np.swapaxes(x, time_axis, -1)
```
philippemiron, line 71: if you are not using `x` I would just overwrite it.

milancurcic: Not a big deal, but I think it's a good practice to leave input arguments untouched because otherwise the code may be more difficult to read and understand. For example here, `x` may have a different shape in different code sections depending on input state. For a similar example, see [this](https://github.com/Cloud-Drift/clouddrift/pull/218#discussion_r1280847570) where I made a reasoning mistake because of modifying input arguments.

selipot:
I had to loosen the tolerance on the tests for `analytic_transform` and `rotary_transform` in `signal.py`. It also looks like I should define test classes for both functions?

clouddrift/signal.py
```diff
@@ -41,57 +43,84 @@ def analytic_transform(
...
+    # time_axis must be in valid range
+    if time_axis < -1 or time_axis > len(x.shape) - 1:
+        raise ValueError(
+            f"time_axis ({time_axis}) is outside of the valid range ([-1,"
+            f" {len(x.shape) - 1}])."
+        )
+
+    # Reshape the inputs to ensure the time axis is last (fast-varying)
+    # should we add a condition to skip if time_axis=-1 or time_axis == len(x.shape)-1
+    # do we need to copy x here? does it matter for memory and performance?
```
milancurcic, line 69: No. From `swapaxes` docs:

>        For NumPy >= 1.10.0, if `a` is an ndarray, then a view of `a` is
>        returned; otherwise a new array is created. For earlier NumPy
>        versions a view of `a` is returned only if the order of the
>        axes is changed, otherwise the input array is returned.

When a view is returned, it's the same data in memory, i.e. there's no copy. I will modify the comment to clarify this.

milancurcic:
Signal tests were actually not running; they are now with ac6ea3e. Maybe the class is needed not sure; I'm not too familiar with `unittest`.

philippemiron:
> Signal tests were actually not running; they are now with [ac6ea3e](https://github.com/Cloud-Drift/clouddrift/commit/ac6ea3e23146c2fe415c1cb8ebf3d31d596667c2). Maybe the class is needed not sure; I'm not too familiar with `unittest`.

It does have to be inside a subclass of `unittest.TestCase` to be run by `unittest.main()`.

milancurcic:
@selipot I see that you test both even and odd-sized input arrays and that the even-sized require high tolerance to pass.

```python
In [37]: x = np.random.rand(9)

In [38]: z = analytic_transform(x)

In [39]: x - np.mean(x) - z.real
Out[39]: 
array([1.38777878e-16, 5.55111512e-17, 5.55111512e-17, 1.66533454e-16,
       0.00000000e+00, 1.66533454e-16, 5.55111512e-17, 1.66533454e-16,
       5.55111512e-17])

In [40]: x = np.random.rand(10)

In [41]: z = analytic_transform(x)

In [42]: x - np.mean(x) - z.real
Out[42]: 
array([-0.01073005,  0.01073005, -0.01073005,  0.01073005, -0.01073005,
        0.01073005, -0.01073005,  0.01073005, -0.01073005,  0.01073005])
```

Is it expected that the real component `analytic_transform(x)` when `x` is even-sized has such large deviations from `x - np.mean(x)`? Compared to `1e-16` in case of odd-sized `x`.

milancurcic:
The N-d part and associated tests also seems to have some strange behaviors. Tests fail but if I step through the code by hand values check out. The output also doesn't seem to be deterministic, let me confirm.

milancurcic:
I'm confused. Which part of `analytic_transform` is not deterministic?

```python
In [42]: x = np.random.rand(9)

In [43]: z1 = analytic_transform(x)

In [44]: z2 = analytic_transform(x)

In [45]: z2 - z1
Out[45]: 
array([-2.22044605e-16+0.00000000e+00j, -2.22044605e-16-5.55111512e-17j,
       -1.24900090e-16+0.00000000e+00j, -1.66533454e-16+0.00000000e+00j,
       -2.22044605e-16+2.08166817e-17j, -1.66533454e-16+1.11022302e-16j,
       -1.66533454e-16+0.00000000e+00j, -1.66533454e-16+5.55111512e-17j,
       -2.49800181e-16+6.59194921e-17j])
```

selipot:
> I'm confused. Which part of `analytic_transform` is not deterministic?
> 
> ```python
> In [42]: x = np.random.rand(9)
> 
> In [43]: z1 = analytic_transform(x)
> 
> In [44]: z2 = analytic_transform(x)
> 
> In [45]: z2 - z1
> Out[45]: 
> array([-2.22044605e-16+0.00000000e+00j, -2.22044605e-16-5.55111512e-17j,
>        -1.24900090e-16+0.00000000e+00j, -1.66533454e-16+0.00000000e+00j,
>        -2.22044605e-16+2.08166817e-17j, -1.66533454e-16+1.11022302e-16j,
>        -1.66533454e-16+0.00000000e+00j, -1.66533454e-16+5.55111512e-17j,
>        -2.49800181e-16+6.59194921e-17j])
> ```

Isn't that numerical noise on the FFT?

milancurcic:
I hope not! :) There shouldn't be any noise.

```python
In [69]: x
Out[69]: 
array([-0.43081607,  0.33690368, -0.10336309, -0.31246727,  0.38224228,
       -0.30749581, -0.18163184,  0.3732039 ,  0.24342423])

In [70]: x1 = np.fft.ifft(np.fft.fft(x))

In [71]: x2 = np.fft.ifft(np.fft.fft(x))

In [72]: x2 - x1
Out[72]: 
array([0.+0.j, 0.+0.j, 0.+0.j, 0.+0.j, 0.+0.j, 0.+0.j, 0.+0.j, 0.+0.j,
       0.+0.j])
```

Can you check the behavior of `analytic_transform(x)` on your end?

selipot:
> ```python
>  x1 = np.fft.ifft(np.fft.fft(x))
> ```

I confirm on my end that same behavior for `analytic_transform` and `np.fft`. And I don't understand it either.

jonathanlilly:
Hi all, there was a typo in the jLab code for the even case.  It should have read

    %analytic signal has M/2 + 1 coefficients
    index=(M/2+2):size(Z,dim);
    Z=vindexinto(Z,0,index,dim);
    %divide Nyquist by two in even case
    Z=vindexinto(Z,vindex(Z,M/2+1,dim)/2,M/2+1,dim);
    
I confirmed that the original code led to an error with the even case of the Solomon Islands time series, and that this revised version passes.

What should be happening is that the point at the Nyquist, M/2+1, is getting divided by two, but in the original code the half-Nyquist value was incorrectly being mapped to the last Fourier coefficient.

I assume you are all accounting for the fact that Python is using 0-based indexing?

Sorry about that.  And thanks for finding the bug!

selipot:
> I managed to get N-d tests to pass with low tolerance (I'm still very puzzled about the behavior of previous tests).
> 
> Regarding the `rotary_transform` is it expected for this deviation to be this large?
> 
> ```python
> In [14]: u = np.random.random((9))
> 
> In [15]: v = np.random.random((9))
> 
> In [16]: zp, zn = rotary_transform(u, v)
> 
> In [17]: u + 1j * v - (zp + np.conj(zn))
> Out[17]: 
> array([-0.48965686-0.37362692j, -0.48965686-0.37362692j,
>        -0.48965686-0.37362692j, -0.48965686-0.37362692j,
>        -0.48965686-0.37362692j, -0.48965686-0.37362692j,
>        -0.48965686-0.37362692j, -0.48965686-0.37362692j,
>        -0.48965686-0.37362692j])
> ```
> 
> The behavior is repeatable with different magnitudes of course, depending on the random input sample.

ok, what the heck is going on here? While the mean value of `zp+np.conj(zn)` is the expected result, why are the mean values of `u` and `v` changing? Are `u` and `v` global variables now???

Screenshot 2023-08-22 at 4 09 21 PM

philippemiron:
All structures (numpy array, Dataset, dictionnary) are passed by reference. So the following is normal behavior.

```
import numpy as np

a = np.zeros(4)

def modify_a(a):
    a[0] = 1
    return

print(a)
modify_a(a)
print(a)
```

```
[0. 0. 0. 0.]
[1. 0. 0. 0.]
```

selipot:
> Hi all, there was a typo in the jLab code for the even case. It should have read
> 
> ```
> %analytic signal has M/2 + 1 coefficients
> index=(M/2+2):size(Z,dim);
> Z=vindexinto(Z,0,index,dim);
> %divide Nyquist by two in even case
> Z=vindexinto(Z,vindex(Z,M/2+1,dim)/2,M/2+1,dim);
> ```
> 
> I confirmed that the original code led to an error with the even case of the Solomon Islands time series, and that this revised version passes.
> 
> What should be happening is that the point at the Nyquist, M/2+1, is getting divided by two, but in the original code the half-Nyquist value was incorrectly being mapped to the last Fourier coefficient.
> 
> I assume you are all accounting for the fact that Python is using 0-based indexing?
> 
> Sorry about that. And thanks for finding the bug!

Thank you @jonathanlilly for checking this for us! Turns out we had a coding mistake on top of your typo so it's all mutually beneficial in the end

milancurcic:
@selipot @philippemiron great find, may be a subtle bug in how `x_` is assigned, I haven't looked yet. This would explain the behavior I saw earlier.

milancurcic:
> Thank you @jonathanlilly for checking this for us! Turns out we had a coding mistake on top of your typo so it's all mutually beneficial in the end

Too bad the two bugs didn't cancel each other out! ;)

selipot:
Are satisfied with the number of tests for `rotary_transforms` following [my last commit](https://github.com/Cloud-Drift/clouddrift/pull/228/commits/195db40fbf4dcd261c59fce7ab63893864187ea9) ?

milancurcic (approved):
I'm satisfied.

PR #232, opened 2023-08-12
proposed fix for #229

### LINKED ISSUE #229: velocity_from_position not behaving as expected
I think `velocity_from_position` (and similarly some other related functions) are not behaving as expected. If the time axis is not the last (-1), isn't the intension to return an output of the same size as the input? See the case below where I expect `u` to be the same size as `x`

```
x = np.random.rand(10,20,100)
y = np.random.rand(10,20,100)
t = np.tile(np.arange(0,100),(10,20,1))
u,v = velocity_from_position(x,y,t,time_axis=0)
display(np.shape(x))
display(np.shape(u))

(10, 20, 100)
(100, 10, 20)
```

Maybe a solution is to use `numpy.swapaxes` as I do in my PR #228

selipot:
fix #229

### DIFF
--- a/clouddrift/analysis.py
+++ b/clouddrift/analysis.py
@@ -612,17 +612,15 @@ def position_from_velocity(
             f" {len(x.shape) - 1}])."
         )
 
-    # Nominal order of axes on input, i.e. (0, 1, 2, ..., N-1)
-    target_axes = list(range(len(u.shape)))
-
-    # If time_axis is not the last one, transpose the inputs
+    # If time_axis is not the last one, swap axes
     if time_axis != -1 and time_axis < len(u.shape) - 1:
-        target_axes.append(target_axes.pop(target_axes.index(time_axis)))
-
-    # Reshape the inputs to ensure the time axis is last (fast-varying)
-    u_ = np.transpose(u, target_axes)
-    v_ = np.transpose(v, target_axes)
-    time_ = np.transpose(time, target_axes)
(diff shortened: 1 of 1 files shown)


selipot:
the tests on the time axis fail which is obviously related to what this PR is intending on fixing. My issue is that I don't fully understand these tests. However, I think that my proposed fix can work, e.g.:
```
x = np.random.rand(10,20,100)
y = np.random.rand(10,20,100)
t = np.tile(np.arange(0,100),(10,20,1))
u,v = analysis.position_from_velocity(x,y,t,0,0,time_axis=0)
display(np.shape(x))
display(np.shape(u))

(10, 20, 100)
(10, 20, 100)

x = np.random.rand(10,20,100)
y = np.random.rand(10,20,100)
t = np.tile(np.arange(0,100),(10,20,1))
u,v = analysis.position_from_velocity(x,y,t,0,0,time_axis=1)
display(np.shape(x))
display(np.shape(u))

(10, 20, 100)
(10, 20, 100)

x = np.random.rand(10,20,100)
y = np.random.rand(10,20,100)
t = np.tile(np.arange(0,100),(10,20,1))
u,v = analysis.position_from_velocity(x,y,t,0,0,time_axis=-1)
display(np.shape(x))
display(np.shape(u))

(10, 20, 100)
(10, 20, 100)
```

selipot:
replaced by #232

PR #234, opened 2023-08-17
count->rowsize gdp

selipot:
change `count` to `rowsize` for gdp adapter module


milancurcic:
I tested it with hourly and 6-hourly, it works, thank you.

PR #238, opened 2023-08-24
Add pandas to dependencies

### LINKED ISSUE #231: Missing dependencies
I realized that we are importing `pandas` in `analysis.py` but it is not in the dependencies of the project `pyproject.toml` or `environment.yaml`.

milancurcic:
Fixes #231.


milancurcic:
Good question, need to check, but I vaguely remember that it was an implicit dependency to something, maybe to xarray. Like how we don't use netCDF4 directly, but need it.

philippemiron:
Ok. Looks good then!

philippemiron (approved):

PR #240, opened 2023-08-26
start of wavelet module

selipot:
With this draft PR is am starting the writing of a new wavelet module (maybe to be incorporated in signal module eventually).  I will be going slowly through the Morse wavelet implementation found in jlab ... and this will take some time. I would appreciate coding comments! 

I should probably write the tests as I go but am not sure what they should be for simple formulas. Tests on sizes and shapes?


selipot:
And since I am shamelessly copying @jonathanlilly 's code from jlab, how do I properly acknowledge the source? What does your license imply @jonathanlilly ?

milancurcic:
Before adding new stuff, let's double check that the [scipy.signal](https://docs.scipy.org/doc/scipy/reference/signal.html#wavelets) doesn't already have the wavelets that clouddrift needs.

selipot:
> Before adding new stuff, let's double check that the [scipy.signal](https://docs.scipy.org/doc/scipy/reference/signal.html#wavelets) doesn't already have the wavelets that clouddrift needs.

It does not, I know that much. 

The Morse wavelet is derived from the work of @jonathanlilly and collaborators. It is an analytic wavelet which is particularly useful for Lagrangian bivariate data. It has been ported in [a python package](https://github.com/danrsc/analytic_wavelet/tree/master) but I am not in love with their implementation that I have already tested extensively.

So, yes, I want to add that stuff to clouddrift.

milancurcic:
Translated code needs to be licensed under the upstream (jLab's) license (currently [CC BY-NC-SA](http://creativecommons.org/licenses/by-nc-sa/4.0/)) unless @jonathanlilly lets us re-license. I strongly recommend relicensing because the NC part prevents commercial use.

milancurcic:
> I should probably write the tests as I go but am not sure what they should be for simple formulas. Tests on sizes and shapes?

You can start with a known small input and expected output values, based on jLab's wavelet. Make a test function that passes that input to your Python wavelet function and compares (with some small tolerance) the expected output from jLab against your Python function output. The test will of course fail, but let it fail as you work on the implementation and when it passes you'll know the implementation is correct.

milancurcic:
> It does not, I know that much.
>
> The Morse wavelet is derived from the work of @jonathanlilly and collaborators. It is an analytic wavelet which is particularly useful for Lagrangian bivariate data. It has been ported in a python package but I am not in love with their implementation that I have already tested extensively.

Great, that's all I ask--it wasn't clear from your first message. :)

clouddrift/wavelet.py
```diff
@@ -0,0 +1,184 @@
...
+        a = np.where(be == 0, 2, 2 / (np.exp(be * np.log(om) - om**ga)))
+    else:
+        raise ValueError(
+            "Normalization option (norm) must be one of 'energy' or 'bandpass'."
+        )
+
+    return a
+
+
+def _gamma(
```
milancurcic, line 148: Checkout SciPy gamma and its log counterpart that accepts array-likes: https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.gamma.html#scipy-special-gamma

selipot: I saw that but i questioned if it was worth it to have the whole scipy depency just for this?

milancurcic: 100%. I can explain in more detail on our call, but there are several benefits that outweigh the cost of increased package size.

selipot:
I feel like `morsewave` should be defined as a class because it clearly has characteristics such as parameters etc. I welcome  advice on this.

clouddrift/wavelet.py
```diff
@@ -0,0 +1,523 @@
...
+        x_ = np.concatenate((np.flip(x_, axis=-1), x_, np.flip(x_, axis=-1)), axis=-1)
+    elif boundary == "zeros":
+        x_ = np.concatenate((np.zeros_like(x_), x_, np.zeros_like(x_)), axis=-1)
+    elif boundary == "periodic":
+        x_ = np.concatenate((x_, x_, x_), axis=-1)
+    else:
+        raise ValueError("boundary must be one of 'mirror', 'align', or 'zeros'.")
+
+    # danger zone, use different letters from jlab: n is their time length
+    n = np.shape(x)[-1]
```
philippemiron, line 109: I would suggest using variable names everywhere it's possible. 

For example, `time_length = np.shape(x)[-1)`. Could be good! 

Same below, we have `k, l, w, om`, we could switch to meaningful names too.

clouddrift/wavelet.py
```diff
@@ -0,0 +1,523 @@
...
+        If the time axis is outside of the valid range ([-1, N-1]).
+        If the shape of time axis is different for input and wavelet.
+        If ``boundary not in ["mirror", "zeros", "periodic"]``.
+
+    See Also
+    --------
+    :func:`morsewave`, `morsefreq`
+    """
+    # time_axis must be in valid range
+    if time_axis < -1 or time_axis > len(x.shape) - 1:
```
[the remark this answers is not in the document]
milancurcic: Ah, true, that didn't occur to me.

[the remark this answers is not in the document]
selipot: So this test here and elsewhere should be?
```
if time_axis < - x.ndim or time_axis > x.ndim - 1 
```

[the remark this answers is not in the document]
philippemiron: I think this should work too. Or the absolute value I added.

clouddrift/wavelet.py
```diff
@@ -0,0 +1,523 @@
...
+    """
+    Continuous wavelet transform.
+
+    Parameters
+    ----------
+    x : np.ndarray
+        Real- or complex-valued signals
+    psi : np.ndarray
+        A suite of Morse wavelets as returned by function morsewave. The dimensions
+        of the suite of Morse wavelets typically are typically (f_order, f_axis, time_axis).
```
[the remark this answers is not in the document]
selipot: Thanks

clouddrift/wavelet.py
```diff
@@ -0,0 +1,523 @@
...
+    if time_axis < -1 or time_axis > len(x.shape) - 1:
+        raise ValueError(
+            f"time_axis ({time_axis}) is outside of the valid range ([-1,"
+            f" {len(x.shape) - 1}])."
+        )
+    # Positions and time arrays must have the same shape.
+    if x.shape[time_axis] != psi.shape[-1]:
+        raise ValueError("x and psi time axes must have the same length.")
+
+    psi_ = np.moveaxis(psi, [f_axis, order_axis], [-2, -3])
```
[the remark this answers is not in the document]
selipot: `psi` is a wavelet passed as an argument, generated outside of the function and I don't want this function to modify it. If I understand variable and function behaviors correctly. Does this make sense?

clouddrift/wavelet.py
```diff
@@ -0,0 +1,523 @@
...
+
+    # initialization: output will be ((x_orders),f_order, f_axis, time_axis)
+    # w = np.tile(
+    #     np.expand_dims(np.zeros_like(x), (-3, -2)),
+    #     (1, np.shape(psi)[-3], np.shape(psi)[-2], 1),
+    # )
+
+    # if x is of dimension 1 we need to expand
```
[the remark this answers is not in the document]
selipot: Not sure why it would not if used with `apply_ragged`? I need to do some algebra with the input and thus needs to add dimensions to to it.

clouddrift/wavelet.py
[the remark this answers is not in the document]
philippemiron: I'm not 100% of the `ndim` here. If I do: `np.zeros(10).ndim` I get 1... So probably safer to do `len(x)`.

Note: My `abs(time_axis)` is not correct.

clouddrift/wavelet.py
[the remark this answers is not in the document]
philippemiron: okok!

clouddrift/wavelet.py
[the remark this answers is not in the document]
philippemiron: I saw that you are adding padding after, so indeed a copy is needed.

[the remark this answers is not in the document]
selipot: ok!

milancurcic:
I replaced the hand-coded `_gamma` and `_lgamma` and imported the equivalents from `scipy.special`. I commented out their definitions for reference, but we should delete them once we're ready to merge.

`scipy.special` also provides `laguerre` which we may be able to use. However, I wasn't sure because the SciPy one only takes an order `n`, while the hand-coded one takes `k` and `c`, so maybe they are somewhat different functions.

clouddrift/wavelet.py
```diff
@@ -0,0 +1,534 @@
...
+    if np.ndim(x) < 2:
+        x_ = np.expand_dims(x, axis=0)
+    else:
+        x_ = np.moveaxis(x, time_axis, -1)
+
+    if ~np.all(np.isreal(x)):
+        if norm == "energy":
+            x_ = x_ / np.sqrt(2)
+        elif norm == "bandpass":
+            x_ = x_ / 2
```
milancurcic, line 94:
```suggestion
            x_ /= np.sqrt(2)
        elif norm == "bandpass":
            x_ /= 2
```

clouddrift/wavelet.py
```diff
@@ -0,0 +1,534 @@
...
+    time_length = np.shape(x)[-1]
+    time_length_ = np.shape(x_)[-1]
+
+    # pad wavelet with zeros?
+    # wave_ = np.concatenate((np.zeros_like(wave), wave, np.zeros_like(wave)), axis=-1)
+    order_length, freq_length, _ = np.shape(wave)
+    wave_ = np.zeros((order_length, freq_length, time_length_), dtype="cfloat")
+    # display(np.shape(wave_))
+    index = slice(
+        int(0 + np.floor(time_length_ - time_length) / 2),
```
milancurcic, line 118:
```suggestion
        int(np.floor(time_length_ - time_length) / 2),
```

clouddrift/wavelet.py
```diff
@@ -0,0 +1,534 @@
...
+    # add a test for rad_freq being a numpy array
+    # initialization
+    wave = np.zeros((n, order, len(rad_freq)), dtype="cdouble")
+    wavefft = np.zeros((n, order, len(rad_freq)), dtype="cdouble")
+
+    # call to morsewave take only ga and be as float, no array
+    fo, _, _ = morsefreq(ga, be)
+    for i in range(0, len(rad_freq)):
+        wave_tmp = np.zeros((n, order), dtype="cdouble")
+        wavefft_tmp = np.zeros((n, order), dtype="cdouble")
```
milancurcic, line 214:
```suggestion
    wave = np.zeros((n, order, len(rad_freq)), dtype=np.complex128)
    wavefft = np.zeros((n, order, len(rad_freq)), dtype=np.complex128)

    # call to morsewave take only ga and be as float, no array
    fo, _, _ = morsefreq(ga, be)
    for i in range(0, len(rad_freq)):
        wave_tmp = np.zeros((n, order), dtype=np.complex128)
        wavefft_tmp = np.zeros((n, order), dtype=np.complex128)
```

clouddrift/wavelet.py
```diff
@@ -0,0 +1,534 @@
...
+    wavezero: np.ndarray,
+    order: Optional[int] = 1,
+    norm: Optional[str] = "bandpass",
+) -> np.ndarray:
+    """
+    Derive first family wavelet
+    """
+    r = (2 * be + 1) / ga
+    c = r - 1
+    L = np.zeros_like(norm_rad_freq, dtype="float")
```
milancurcic, line 286:
```suggestion
    L = np.zeros_like(norm_rad_freq, dtype=np.float64)
```

selipot:
> General question about numerical types used in wavelet calculation: Is double precision important or would single precision suffice? We could inherit precision from the input array.

Why not double precision that seems to be the default in python? I am not sure that any gain of speed or storage is worthwhile at this level of the testing. I will nevertheless ask @jonathanlilly

milancurcic:
Less so for speed and storage, more for memory use (factor of 2). Accuracy of weather, ocean, and climate analysis and modeling generally don't benefit from double precision except for where needed for convergence. For a sufficiently large dataset, single vs. double precision makes a difference whether you can do it or not.

selipot:
I see! I suspect double precision is beneficial for the calculation of the wavelet itself (`morsewave`) but may not be for the calculation of the wavelet transform (`wavetrans`). We could enforce double precision for `morsewave` and inherit the precision of `x` for `wavetrans`. If that sounds reasonable how does one do that?

clouddrift/wavelet.py
```diff
@@ -0,0 +1,534 @@
...
+    if time_axis < -1 or time_axis > len(x.shape) - 1:
+        raise ValueError(
+            f"time_axis ({time_axis}) is outside of the valid range ([-1,"
+            f" {len(x.shape) - 1}])."
+        )
+    # Positions and time arrays must have the same shape.
+    if x.shape[time_axis] != wave.shape[-1]:
+        raise ValueError("x and wave time axes must have the same length.")
+
+    wave_ = np.moveaxis(wave, [freq_axis, order_axis], [-2, -3])
```
milancurcic, line 75: This `wave_` doesn't seem to be used, can it be removed?

milancurcic:
> We could enforce double precision for morsewave and inherit the precision of x for wavetrans. If that sounds reasonable how does one do that?

I think it would be where we initialize `wave_` and tell it to use the same `dtype` as `x`. It seems like the arrays that are then computed from `wave_` and finally `wtx` that is returned by the function would inherit that type.

clouddrift/wavelet.py
selipot: hmmm it is meant to be used at line 121 where it put into the padded wavelet which I initialized and also called `wave_` by mistake. Yet the tests passed...

selipot:
> > We could enforce double precision for morsewave and inherit the precision of x for wavetrans. If that sounds reasonable how does one do that?
> 
> I think it would be where we initialize `wave_` and tell it to use the same `dtype` as `x`. It seems like the arrays that are then computed from `wave_` and finally `wtx` that is returned by the function would inherit that type.

The one  tricky thing is that `x` could be real or complex. I'll find a way.

clouddrift/wavelet.py
selipot: I think that complex128 is an alias which is platform-dependent? cdouble seems to be norm? see https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.double

clouddrift/wavelet.py
```diff
@@ -0,0 +1,534 @@
...
+        int(np.floor(time_length_ - time_length) / 2),
+        int(time_length + np.floor(time_length_ - time_length) / 2),
+    )
+    wave_[:, :, index] = wave
+
+    # here we assume that wavelet and data have exactly the same number of data points;
+    # to do: include case wavelet is shorter or longer
+
+    # take fft along axis = -1
+    wavefft_ = np.fft.fft(wave_)
```
philippemiron, line 127: From https://numpy.org/devdocs/reference/routines.fft.html#type-promotion, it looks like automatically, the output of `numpy.fft` is converted to `float64` and `complex128`. So I don't think you need to manually define `dtype="cfloat"` for `wave_`.

selipot: ok! Then do you think we should bother to make sure we return the transform `wtx` with the same precision as the input `x`?

philippemiron:
> I see! I suspect double precision is beneficial for the calculation of the wavelet itself (`morsewave`) but may not be for the calculation of the wavelet transform (`wavetrans`). We could enforce double precision for `morsewave` and inherit the precision of `x` for `wavetrans`. If that sounds reasonable how does one do that?

I don't know if there is a better way but you could do:

```
complex_dtype = np.complex64 if x.dtype == "float32" else np.complex128
wave_ = np.zeros(1, dtype=complex_dtype)
```

selipot:
> > I see! I suspect double precision is beneficial for the calculation of the wavelet itself (`morsewave`) but may not be for the calculation of the wavelet transform (`wavetrans`). We could enforce double precision for `morsewave` and inherit the precision of `x` for `wavetrans`. If that sounds reasonable how does one do that?
> 
> I don't know if there is a better way but you could do:
> 
> ```
> complex_dtype = np.complex64 if x.dtype == "float32" else np.complex128
> wave_ = np.zeros(1, dtype=complex_dtype)
> ```

ok but should we not use `csingle` and `cdouble`???

clouddrift/wavelet.py
philippemiron: In general, I would not modify the type of the output and leave it with the same precision as the input anywhere where it's possible.

philippemiron:
It is the same thing. Single is 32 bits and Double is 64 bits, so for complex it becomes 64/128.

image

selipot:
> It is the same thing. Single is 32 bits and Double is 64 bits, so for complex it becomes 64/128.
> 
> image

But this
Screenshot 2023-08-31 at 4 18 00 PM

philippemiron:
> > It is the same thing. Single is 32 bits and Double is 64 bits, so for complex it becomes 64/128.
> > image
> 
> But this Screenshot 2023-08-31 at 4 18 00 PM

Interesting... on Macos it's working so I'm a bit confused. We run the tests on Linux/macOS/Windows so worst case it would fail later. I guess either works.

selipot:
Super interesting because tests fail on my mac. Tests pass with `np.cdouble`.

philippemiron:
Maybe the ARM version of macOS has the same alias but not the regular one.

milancurcic:
No need to mess around with aliases, let's just go with the main names / defaults from the docs (`np.cdouble` etc.). I didn't know that all this time I was using something OS-specific.

selipot:
> Maybe the ARM version of macOS has the same alias but not the regular one.

🤯

philippemiron:
> No need to mess around with aliases, let's just go with the main names / defaults from the docs (`np.cdouble` etc.). I didn't know that all this time I was using something OS-specific.

Same! We all learned something today!

milancurcic:
For this PR we still need to either:

1. Include jLab's license at the top of wavelet.py; or
2. Get permission from @jonathanlilly to license the translation under CloudDrift's license.

clouddrift/wavelet.py
```diff
@@ -0,0 +1,518 @@
...
+from typing import Optional, Tuple, Union
+import xarray as xr
+import warnings
+from scipy.special import gamma as _gamma, gammaln as _lgamma
+
+
+def wavetrans(
+    x: np.ndarray,
+    wave: np.ndarray,
+    norm: Optional[str] = "bandpass",
```
milancurcic, line 15: `norm` not in docstring below, yet.

clouddrift/wavelet.py
```diff
@@ -0,0 +1,518 @@
...
+import numpy as np
+from typing import Optional, Tuple, Union
+import xarray as xr
+import warnings
+from scipy.special import gamma as _gamma, gammaln as _lgamma
+
+
+def wavetrans(
+    x: np.ndarray,
+    wave: np.ndarray,
```
milancurcic, line 14: Would it make sense to call this `wavelets`? `wave` is very general, I think.

clouddrift/wavelet.py
```diff
@@ -0,0 +1,518 @@
...
+"""
+
+import numpy as np
+from typing import Optional, Tuple, Union
+import xarray as xr
+import warnings
+from scipy.special import gamma as _gamma, gammaln as _lgamma
+
+
+def wavetrans(
```
milancurcic, line 12: I suggest

```suggestion
def wavelet_transform(
```

clouddrift/wavelet.py
```diff
@@ -0,0 +1,518 @@
...
+    x: np.ndarray,
+    wave: np.ndarray,
+    norm: Optional[str] = "bandpass",
+    boundary: Optional[str] = "mirror",
+    time_axis: Optional[int] = -1,
+    freq_axis: Optional[int] = -2,
+    order_axis: Optional[int] = -3,
+) -> np.ndarray:
+    """
+    Continuous wavelet transform.
```
milancurcic, line 22: Should this say anything more or this is enough?

clouddrift/wavelet.py
```diff
@@ -0,0 +1,518 @@
...
+    complex_dtype = np.cdouble if x.dtype == np.single else np.csingle
+    wtx = np.fft.ifft(X_ * np.conj(_wavefft)).astype(complex_dtype)
+    wtx = wtx[..., index]
+    # remove extra dimensions
+    wtx = np.squeeze(wtx)
+
+    return wtx
+
+
+def morsewave(
```
milancurcic, line 141: What do you think about

```suggestion
def morse_wavelets(
```

clouddrift/wavelet.py
```diff
@@ -0,0 +1,518 @@
...
+    n: int,
+    ga: float,
+    be: float,
+    rad_freq: np.ndarray,
+    order: Optional[int] = 1,
+    norm: Optional[str] = "bandpass",
```
milancurcic, line 147:
```suggestion
    length: int,
    gamma: float,
    beta: float,
    radian_frequency: np.ndarray,
    order: Optional[int] = 1,
    normalization: Optional[str] = "bandpass",
```

clouddrift/wavelet.py
```diff
@@ -0,0 +1,518 @@
...
+
+        # wavelet frequencies
+        fact = np.abs(rad_freq[i]) / fo
+        # norm_rad_freq first dim is n points
+        norm_rad_freq = 2 * np.pi * np.linspace(0, 1 - 1 / n, n) / fact
+        if norm == "energy":
+            if be == 0:
+                wavezero = np.exp(-(norm_rad_freq**ga))
+            else:
+                wavezero = np.exp(be * np.log(norm_rad_freq) - norm_rad_freq**ga)
```
milancurcic, line 207: Could this not be just

```suggestion    
            wavezero = np.exp(be * np.log(norm_rad_freq) - norm_rad_freq**ga)
```

if `be == 0`, the first term in the exponent argument drops.

clouddrift/wavelet.py
```diff
@@ -0,0 +1,518 @@
...
+
+def _morsewave_first_family(
+    fact: float,
+    n: int,
+    ga: float,
+    be: float,
+    norm_rad_freq: np.ndarray,
+    wavezero: np.ndarray,
+    order: Optional[int] = 1,
+    norm: Optional[str] = "bandpass",
```
milancurcic, line 259: Here and below, I suggest renaming these as above.

clouddrift/wavelet.py
```diff
@@ -0,0 +1,518 @@
...
+#        x_ = x.flatten()
+#        y = np.zeros_like(x_, dtype="float")
+#        for i in range(0, np.size(x)):
+#            y[i] = math.lgamma(x_[i])
+#        y = np.reshape(y, np.shape(x))
+#    else:
+#        y = math.lgamma(x)
+#
+#    return y
+
```
milancurcic, line 504: If we're happy with SciPy we can remove this.

```suggestion
```

clouddrift/wavelet.py
```diff
@@ -0,0 +1,518 @@
...
+    """
+    # test common size? or could be broadcasted
+    p = np.sqrt(ga * be)
+    skew = ga - 3 / p
+    kurt = 3 - skew**2 - 2 / p**2
+
+    return p, skew, kurt
+
+
+def morseafun(
```
milancurcic, line 442: What would be a full name of this function?

clouddrift/wavelet.py
```diff
@@ -0,0 +1,518 @@
...
+    )
+
+    fe = 1 / (2 ** (1 / ga)) * _gamma((2 * be + 2) / ga) / _gamma((2 * be + 1) / ga)
+
+    fi = _gamma((be + 2) / ga) / _gamma((be + 1) / ga)
+
+    return fm, fe, fi
+
+
+def morsespace(
```
milancurcic, line 341: This name is good but may want a _

```suggestion
def morse_space(
```

clouddrift/wavelet.py
```diff
@@ -0,0 +1,518 @@
...
+                coeff = 1
+
+        index = slice(0, int(np.round(n / 2)))  # how to define indices?
+        L[index] = _laguerre(2 * norm_rad_freq[index] ** ga, i, c)
+        wavefft1[:, i] = coeff * wavezero * L
+
+    return wavefft1
+
+
+def morsefreq(
```
milancurcic, line 286:
```suggestion
def morse_frequency(
```

milancurcic (approved):
@selipot I did another review pass and left several suggestions, most of which you probably already have on the radar, but I'm including them just in case.

clouddrift/wavelet.py
selipot: Morse wavelet amplitude or a-function, kind of a mouthful.

milancurcic: Cam we call it `morse_amplitude`?

clouddrift/wavelet.py
selipot: ok

clouddrift/wavelet.py
selipot: changed to `wavelet`

clouddrift/wavelet.py
selipot: ok!

selipot:
`morse_space` is not behaving as expected which I will discuss with @jonathanlilly

selipot:
`morse_space` fixed. Need to write tests.

clouddrift/wavelet.py
selipot: done

selipot:
@milancurcic @philippemiron I would be very grateful if you could do another pass reading the docstrings for typos and comprehension. @milancurcic please add the jlab license at the top of module as we discuss.

clouddrift/wavelet.py
```diff
@@ -0,0 +1,801 @@
...
+    radian_frequency: np.ndarray,
+    complex: Optional[bool] = False,
+    order: Optional[int] = 1,
+    normalization: Optional[str] = "bandpass",
+    boundary: Optional[str] = "mirror",
+    time_axis: Optional[int] = -1,
+) -> Union[Tuple[np.ndarray], np.ndarray]:
+    """
+    Apply a continuous wavelet transform to an input signal using the generalized Morse
+    wavelets of Olhede and Walden (2002). The wavelet transform is normalized differently
```
philippemiron, line 26: I think we should have the full reference below. See in numpy for example they have ([link](https://github.com/scipy/scipy/blob/v1.11.2/scipy/integrate/_ode.py#L103-L571)):

```
    "dopri5"

        This is an explicit runge-kutta method of order (4)5 due to Dormand &
        Prince (with stepsize control and dense output).

        Authors:

            E. Hairer and G. Wanner
            Universite de Geneve, Dept. de Mathematiques
            CH-1211 Geneve 24, Switzerland
            e-mail:  ernst.hairer@math.unige.ch, gerhard.wanner@math.unige.ch
```

clouddrift/wavelet.py
```diff
@@ -0,0 +1,801 @@
...
+    ----------
+    x : np.ndarray
+        Real- or complex-valued signals.
+    gamma: float
+       Gamma parameter of the Morse wavelets.
+    beta: float
+       Beta parameter of the Morse wavelets.
+    radian_frequency: np.ndarray
+       An array of radian frequencies at which the Fourier transform of the wavelets
+       reach their maximum amplitudes. ``radian_frequency`` is between 0 and 2 * np.pi * 0.5,
```
philippemiron, line 40: should this range be validated?

clouddrift/wavelet.py
```diff
@@ -0,0 +1,801 @@
...
+        np.shape(x)[time_axis],
+        gamma,
+        beta,
+        radian_frequency,
+        normalization=normalization,
+        order=order,
+    )
+
+    # apply the wavelet transform, distinguish complex and real cases
+    if complex:  # np.all(np.isreal(x))
```
philippemiron, line 156: Why is there the `# np.all(np.isreal(x))` comment here?

philippemiron: Also, is there cases where we would want to pass a `complex` signal but use `complex=False` here?

clouddrift/wavelet.py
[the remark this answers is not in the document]
philippemiron: (you can mark conversion as resolved when they are done)

clouddrift/wavelet.py
```diff
@@ -0,0 +1,801 @@
...
+    See Also
+    --------
+    :func:`morse_wavelet`, `morse_wavelet_transform`, `morse_freq`
+    """
+    # time_axis must be in valid range
+    if time_axis < -1 or time_axis > len(x.shape) - 1:
+        raise ValueError(
+            f"time_axis ({time_axis}) is outside of the valid range ([-1,"
+            f" {len(x.shape) - 1}])."
+        )
```
philippemiron, line 251: I think it needs a `raise ValueError` here.

clouddrift/wavelet.py
```diff
@@ -0,0 +1,801 @@
...
+    wavelet_ = np.moveaxis(wavelet, [freq_axis, order_axis], [-2, -3])
+
+    # if x is of dimension 1 we need to expand
+    # otherwise make sure time axis is last
+    if np.ndim(x) < 2:
+        x_ = np.expand_dims(x, axis=0)
+    else:
+        x_ = np.moveaxis(x, time_axis, -1)
+
+    # add detrending option?
```
philippemiron, line 265: remove?

philippemiron:
```suggestion
```

clouddrift/wavelet.py
```diff
@@ -0,0 +1,801 @@
...
+
+    wavelet_ = np.moveaxis(wavelet, [freq_axis, order_axis], [-2, -3])
+
+    # if x is of dimension 1 we need to expand
+    # otherwise make sure time axis is last
+    if np.ndim(x) < 2:
+        x_ = np.expand_dims(x, axis=0)
+    else:
+        x_ = np.moveaxis(x, time_axis, -1)
+
```
philippemiron, line 264:
```suggestion
```

clouddrift/wavelet.py
```diff
@@ -0,0 +1,801 @@
...
+    # add detrending option?
```
philippemiron, line 265: remove?

```suggestion
```

philippemiron (approved):
awesome work @selipot  !

clouddrift/wavelet.py
selipot: That should not be necessary because radian frequencies are cyclic, i.e pi = 3pi etc. I modified the docstring to state that radian frequency is typically between 0 and 2 np.pi * 0.5

clouddrift/wavelet.py
selipot: ok removed the comment which was a previous implementation. Yes, the savy/knowledgeable user might want to do that, and this is why it is not automatically tested but passed as an argument.

clouddrift/wavelet.py
selipot: Don't get it?

clouddrift/wavelet.py
```diff
@@ -0,0 +1,801 @@
...
+
+    # pad wavelet with zeros: JML ok
+    order_length, freq_length, _ = np.shape(wavelet)
+    _wavelet = np.zeros((order_length, freq_length, time_length_), dtype=np.cdouble)
+
+    index = slice(
+        int(np.floor(time_length_ - time_length) / 2),
+        int(time_length + np.floor(time_length_ - time_length) / 2),
+    )
+    _wavelet[:, :, index] = wavelet_
```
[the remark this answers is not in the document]
selipot: I think so.

clouddrift/wavelet.py
```diff
@@ -0,0 +1,801 @@
...
+
+    # take fft along axis = -1
+    _wavelet_fft = np.fft.fft(_wavelet)
```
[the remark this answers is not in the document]
selipot: Sorry, I'd like to keep this to keep track of the variables.

tests/wavelet_tests.py
```diff
@@ -0,0 +1,357 @@
...
+import numpy as np
+import unittest
+
+if __name__ == "__main__":
+    unittest.main()
+
+
+class morse_wavelet_transform_tests(unittest.TestCase):
+    def test_morse_wavelet_transform_real(self):
+        length = 1023
```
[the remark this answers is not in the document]
selipot: ok!

clouddrift/wavelet.py
selipot: I'll leave it here to be a future feature.

milancurcic:
I'll provide a final review in the next hour.

clouddrift/wavelet.py
```diff
@@ -0,0 +1,813 @@
...
+    # add detrending option eventually
+
+    # apply boundary conditions
+    if boundary == "mirror":
+        x_ = np.concatenate((np.flip(x_, axis=-1), x_, np.flip(x_, axis=-1)), axis=-1)
+    elif boundary == "zeros":
+        x_ = np.concatenate((np.zeros_like(x_), x_, np.zeros_like(x_)), axis=-1)
+    elif boundary == "periodic":
+        # x_ = np.concatenate((x_, x_, x_), axis=-1) # JML: this not needed
+        x_ = x_
```
milancurcic, line 284:
```suggestion
        continue
```

clouddrift/wavelet.py
```diff
@@ -0,0 +1,813 @@
...
+    fo, _, _ = morse_freq(gamma, beta)
+    for i in range(len(radian_frequency)):
+        wavelet_tmp = np.zeros((length, order), dtype=np.cdouble)
+        waveletfft_tmp = np.zeros((length, order), dtype=np.cdouble)
+
+        # wavelet frequencies
+        fact = np.abs(radian_frequency[i]) / fo
+        # norm_radian_frequency first dim is n points
+        norm_radian_frequency = (
+            2 * np.pi * np.linspace(0, 1 - 1 / length, length) / fact
```
milancurcic, line 413: Is it intended that the first element is 0, given that on L417 we make a log(0)?

clouddrift/wavelet.py
```diff
@@ -0,0 +1,808 @@
...
+        )
+        if normalization == "energy":
+            waveletzero = np.exp(
+                beta * np.log(norm_radian_frequency) - norm_radian_frequency**gamma
```
milancurcic, line 417: This always raises a RuntimeWarning because `norm_radian_frequency[0]` is always 0. Not the best UX. Should we assign the first element on a separate line?

milancurcic (approved):
Great! Please merge when ready!

milancurcic:
Let's release v0.21.0 after merging this.

clouddrift/wavelet.py
selipot: Yes because it then np.exp(np.log(0)) = np.exp(=inf) = 0. I am suggesting to add a warning suppressor as below.

PR #242, opened 2023-08-27
Unittest actions

### LINKED ISSUE #241: Unittest failing
I have been able to reproduce the issue with the tests locally.

```
coverage run -m unittest tests/raggedarray_tests.py
```

I sometimes (~half of the time) gets the following:
```
...
Ran 8 tests in 2.495s

OK
python(87689,0x1de2e1e00) malloc: Double free of object 0x1315fa2b0
python(87689,0x1de2e1e00) malloc: *** set a breakpoint in malloc_error_break to debug
zsh: abort      coverage run -m unittest tests/raggedarray_tests.py
```

and this seems to match the error that we see on the GitHub actions.

philippemiron:
Trying to fix #241.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,7 +22,7 @@ jobs:
         python-version: ["3.9", "3.10", "3.11"]
     steps:
       - uses: actions/checkout@v3
-      - uses: mamba-org/provision-with-micromamba@main
+      - uses: mamba-org/setup-micromamba@v1
         with:
           environment-file: environment.yml
           environment-name: clouddrift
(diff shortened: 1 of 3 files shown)


philippemiron:
Now the tests are passing, but I'm still able to get the same error locally. It seems to be related with awkward, because if I remove:
```
self.ra.to_parquet(PARQUET_ARCHIVE)
```
and all the tests related, I don't get the double free...

milancurcic:
Hmm, I can't reproduce the error locally, but I think this is a good change. Please merge when you're happy with it.

philippemiron:
We will go with this and then we can adapt if this break again down the road.

PR #243, opened 2023-08-30
Fix N-d velocity and position functions with `time_axis`

### LINKED ISSUE #229: velocity_from_position not behaving as expected

milancurcic:
Fixes #229.

* Now using `np.swapaxes` as the most straightforward way to do this.
* Expanded `time_axis` tests to do the differentiation/integration along each of the 3 axes of a 3-d array.

### DIFF
--- a/clouddrift/analysis.py
+++ b/clouddrift/analysis.py
@@ -612,17 +612,15 @@ def position_from_velocity(
             f" {len(x.shape) - 1}])."
         )
 
-    # Nominal order of axes on input, i.e. (0, 1, 2, ..., N-1)
-    target_axes = list(range(len(u.shape)))
-
-    # If time_axis is not the last one, transpose the inputs
-    if time_axis != -1 and time_axis < len(u.shape) - 1:
-        target_axes.append(target_axes.pop(target_axes.index(time_axis)))
-
-    # Reshape the inputs to ensure the time axis is last (fast-varying)
-    u_ = np.transpose(u, target_axes)
-    v_ = np.transpose(v, target_axes)
-    time_ = np.transpose(time, target_axes)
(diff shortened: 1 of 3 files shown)


clouddrift/analysis.py
```diff
@@ -612,17 +612,15 @@ def position_from_velocity(
...
-    time_ = np.transpose(time, target_axes)
+    # Swap axes so that we can differentiate along the last axis.
+    # This is a syntax convenience rather than memory access optimization:
+    # np.swapaxes returns a view of the array, not a copy, if the input is a
+    # NumPy array. Otherwise, it returns a copy. For readability, introduce new
+    # variable names so that we can more easily differentiate between the
+    # original arrays and those with swapped axes.
+    u_ = np.swapaxes(u, time_axis, -1)
+    v_ = np.swapaxes(v, time_axis, -1)
+    time_ = np.swapaxes(time, time_axis, -1)
```
milancurcic, line 623: Since `np.swapaxes` returns a view, we no longer check if `time_axis` is other than the last; if it is, this is a no-op.

selipot (approved):
Looks good to me

PR #246, opened 2023-08-31
Rowsize everywhere

### LINKED ISSUE #221: switch back to rowsize
I would like to propose to switch back to using `rowsize` as the default name for the variable indicating the length of each row of a ragged array.
Pros:
- the term `rowsize` makes sense for an array with rows (think of the ragged array representation/schematic)
- does not interfere with the xarray method `.count` (I just spent 10 minutes trying to figure out why I had an error writing `ds.count` as the input argument of `apply_ragged` whereas I should have written `ds["count"]`)
- the recommendation (from?) to use `count` is not very strong given that the ragged array format is so little used.

milancurcic:
Fixes #221.


selipot (approved):
All good to me. Thanks!

PR #247, opened 2023-09-01
Sphere-to-3d

selipot:
Adding two functions that invert each other to the sphere module: sphere_to_threedim and threedim_to_sphere

These functions will be used by spheretrans which is the wavelet transform on a sphere which is eventually needed for the upcoming extraction of inertial oscillations from geographical positions.

Open to name change etc.

### DIFF
--- a/clouddrift/sphere.py
+++ b/clouddrift/sphere.py
@@ -205,7 +205,7 @@ def recast_lon(lon: np.ndarray, lon0: Optional[float] = -180) -> np.ndarray:
 
     >>> recast_lon(200)
     -160
-    
+
     >>> recast_lon(180)
     -180
 
--- a/clouddrift/sphere.py
+++ b/clouddrift/sphere.py
@@ -435,3 +435,111 @@ def sphere_to_plane(
     y[..., :] = np.cumsum(dy, axis=-1)
 
     return x, y
+
+
+def sphere_to_threedim(
+    lon: np.ndarray,
+    lat: np.ndarray,
+    earth_radius: Optional[float] = EARTH_RADIUS_METERS,
+) -> Tuple[np.ndarray, np.ndarray, np.ndarray]:
+    """Converts latitude and longitude into 3D Cartesian coordinates.
+
+    If converting multiple trajectories, use
+    :func:`apply_ragged` for highest accuracy.
+
+    Parameters
+    ----------
+    lon : np.ndarray
+        An N-d array of longitudes in degrees
+    lat : np.ndarray
+        An N-d array of latitudes in degrees
+    earth_radius: float, optional
+        The radius of the Earth in meters. Default is EARTH_RADIUS_METERS = 6.3781e6
+
+    Returns
+    -------
+    x : np.ndarray
+        x-coordinates in 3D
+    y : np.ndarray
+        y-coordinates in 3D
+    z : np.ndarray
+        z-coordinates in 3D
+
+    Examples
+    --------
+
+    Raises
+    ------
+    AttributeError
+        If ``lon`` and ``lat`` are not NumPy arrays
+
+    See Also
+    --------
+    :func:`threedim_to_sphere`
+    """
+
+    x = earth_radius * np.cos(lat * np.pi / 180) * np.cos(lon * np.pi / 180)
+    y = earth_radius * np.cos(lat * np.pi / 180) * np.sin(lon * np.pi / 180)
+    z = earth_radius * np.sin(lat * np.pi / 180)
+
+    x = np.nan_to_num(x, posinf=np.nan, neginf=np.nan)
+    y = np.nan_to_num(y, posinf=np.nan, neginf=np.nan)
+    z = np.nan_to_num(z, posinf=np.nan, neginf=np.nan)
+
+    return x, y, z
+
+
+def threedim_to_sphere(
+    lon: np.ndarray,
+    lat: np.ndarray,
+    earth_radius: Optional[float] = EARTH_RADIUS_METERS,
+) -> Tuple[np.ndarray, np.ndarray]:
+    """Converts 3D Cartesian coordinates to latitude and longitude.
+
+    If converting multiple trajectories, use
+    :func:`apply_ragged` for highest accuracy.
+
+    Parameters
+    -------
+    x : np.ndarray
+        x-coordinates in 3D
+    y : np.ndarray
+        y-coordinates in 3D
+    z : np.ndarray
+        z-coordinates in 3D
+    earth_radius: float, optional
+        The radius of the Earth in meters. Default is EARTH_RADIUS_METERS = 6.3781e6
+
+    Returns
+    ----------
+    lon : np.ndarray
+        An N-d array of longitudes in degrees
+    lat : np.ndarray
+        An N-d array of latitudes in degrees
+
+    Examples
+    --------
+
+    To write.
+
+    Raises
+    ------
+    AttributeError
+        If ``lon`` and ``lat`` are not NumPy arrays
+
+    See Also
+    --------
+    :func:`sphere_to_threedim`
+    """
+    R = np.sqrt(np.abs(x) ** 2 + np.abs(y) ** 2 + np.abs(z) ** 2)
+    x /= R
+    y /= R
+    z /= R
+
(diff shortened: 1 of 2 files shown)


philippemiron:
If we are discussing double/single precision, maybe this could be done on an ellipsoid, the equations do not look much more complex, see [here](https://gssc.esa.int/navipedia/index.php/Ellipsoidal_and_Cartesian_Coordinates_Conversion). I don't know how much difference that would make.

image

selipot:
That could be a good thing to implement. However, in equation (1) on that page, if the excentricity `e = 0` then the coefficient in front is `2h` which is not what our current equations predict. What am I missing?

selipot:
I think I was missing that `h` is maybe not the Earth radius but the third ellipsoidal coordinate.

philippemiron:
Yes `h` is the height above/below the ellipsoid.

selipot:
I think @philippemiron 's suggestion to implement the ellipsoidal case is a good one but maybe one for another PR/set of functions.

milancurcic:
Ellipsoid sounds great.

clouddrift/sphere.py
```diff
@@ -435,3 +435,133 @@ def sphere_to_plane(
     y[..., :] = np.cumsum(dy, axis=-1)
 
     return x, y
+
+
+def earthsphere_to_cartesian(
```
milancurcic, line 440: Why not `sphere_to_cartesian`? Unless the ellipsoid stuff will be Earth-specific. But it seems to me that these functions will be useful on other planets too, e.g. Mars.

clouddrift/sphere.py
```diff
@@ -435,3 +435,133 @@ def sphere_to_plane(
...
+    lon: np.ndarray,
+    lat: np.ndarray,
+    earth_radius: Optional[float] = EARTH_RADIUS_METERS,
```
milancurcic, line 443: Same here, can be just `radius` that defaults to `EARTH_RADIUS_METERS`.

clouddrift/sphere.py
```diff
@@ -435,3 +435,133 @@ def sphere_to_plane(
...
+        If ``lon`` and ``lat`` are not NumPy arrays.
+
+    See Also
+    --------
+    :func:`cartesian_to_earthsphere`
+    """
+
+    x = earth_radius * np.cos(lat * np.pi / 180) * np.cos(lon * np.pi / 180)
+    y = earth_radius * np.cos(lat * np.pi / 180) * np.sin(lon * np.pi / 180)
+    z = earth_radius * np.sin(lat * np.pi / 180)
```
milancurcic, line 497:
```suggestion
    lonr, latr = np.deg2rad(lon), np.deg2rad(lat)

    x = earth_radius * np.cos(latr) * np.cos(lonr)
    y = earth_radius * np.cos(latr) * np.sin(lonr)
    z = earth_radius * np.sin(latr)
```

clouddrift/sphere.py
```diff
@@ -435,3 +435,133 @@ def sphere_to_plane(
...
+
+    x = np.nan_to_num(x, posinf=np.nan, neginf=np.nan)
+    y = np.nan_to_num(y, posinf=np.nan, neginf=np.nan)
+    z = np.nan_to_num(z, posinf=np.nan, neginf=np.nan)
```
milancurcic, line 501: I don't think this is necessary. For what values of lon, lat do we expect +/- inf in x, y, z? It seems to me that the result values are always bounded by the radius.

clouddrift/sphere.py
```diff
@@ -435,3 +435,133 @@ def sphere_to_plane(
...
+    :func:`earthsphere_to_cartesian`
+    """
+
+    R = np.sqrt(np.abs(x) ** 2 + np.abs(y) ** 2 + np.abs(z) ** 2)
+    x /= R
+    y /= R
+    z /= R
+
+    lon = recast_lon180(np.imag(np.log((x + 1j * y)))) * 180 / np.pi
+    lat = np.arcsin(z) * 180 / np.pi
```
milancurcic, line 565: As above

```suggestion
    lon = np.rad2deg(recast_lon180(np.imag(np.log((x + 1j * y)))))
    lat = np.rad2deg(np.arcsin(z))
```

clouddrift/sphere.py
```diff
@@ -435,3 +435,133 @@ def sphere_to_plane(
...
+    ------
+    AttributeError
+        If ``x``, ``y``, and ``z`` are not NumPy arrays.
+
+    See Also
+    --------
+    :func:`earthsphere_to_cartesian`
+    """
+
+    R = np.sqrt(np.abs(x) ** 2 + np.abs(y) ** 2 + np.abs(z) ** 2)
```
milancurcic, line 559: Are `abs()` needed? It seems to me they make a difference only when inputs are complex; is that expected?

clouddrift/sphere.py
milancurcic: If we remove this, the function is compatible with `xr.DataArray` as well.

milancurcic (approved):

clouddrift/sphere.py
selipot: ok!

clouddrift/sphere.py
selipot: ok!

clouddrift/sphere.py
philippemiron, line 497: I think using numpy `np.radians` makes things cleaner.

```suggestion
    x = earth_radius * np.cos(np.radians(lat)) * np.cos(np.radians(lat))
    y = earth_radius * np.cos(np.radians(lat)) * np.sin(np.radians(lon))
    z = earth_radius * np.sin(np.radians(lat))
```

clouddrift/sphere.py
```diff
@@ -435,3 +435,133 @@ def sphere_to_plane(
...
+    x /= R
+    y /= R
+    z /= R
+
+    lon = recast_lon180(np.imag(np.log((x + 1j * y)))) * 180 / np.pi
```
philippemiron, line 564:
```suggestion
    lon = np.radians(recast_lon180(np.imag(np.log((x + 1j * y)))))
```

philippemiron, line 564:
```suggestion
    lon = np.degrees(recast_lon180(np.imag(np.log((x + 1j * y)))))
```

clouddrift/sphere.py
philippemiron, line 565:
```suggestion
    lat = np.degrees(np.arcsin(z))
```

philippemiron (approved):

clouddrift/sphere.py
selipot: ok!

philippemiron:
I guess we suggested the same thing ... Ps: Personally, I prefer `np.radians` and `np.degrees` than the `np.deg2rad` `np.rad2deg` weird name style.

clouddrift/sphere.py
selipot: You need to convert to degree before recasting?

clouddrift/sphere.py
selipot: ok

milancurcic:
@philippemiron Cool! I didn't know about `np.radians` and `np.degrees`.

clouddrift/sphere.py
selipot: Let's use `np.deg2rad` and `np.rad2deg` as it is already used in other functions.

clouddrift/sphere.py
```diff
@@ -195,20 +195,23 @@ def recast_lon(lon: np.ndarray, lon0: Optional[float] = -180) -> np.ndarray:
     Returns
     -------
     np.ndarray or float
-        Converted longitudes in the range `[lon0, lon0+360]`
+        Converted longitudes in the range `[lon0, lon0+360[`
```
milancurcic, line 198: Should it not be `[lon0, lon0+360)` ?

selipot: Both seem accepted https://en.wikipedia.org/wiki/ISO_31-11

milancurcic: Interesting, I didn't know.

PR #248, opened 2023-09-01
More general subset

### LINKED ISSUE #233: Make `subset` more general
`subset` makes some assumptions about the dataset, which I learned after playing with a new small dataset:

* Dataset must have an `ID(traj)` variable; in MOSAiC, I named this `id(traj)` (forgetting about the `subset` requirement)
* Dataset must have `ids(obs)`; this should not be necessary, we can reconstruct `ids(obs)` via `np.repeat(id, rowsize)`.
* It assumes that the dimensions are `obs` and `traj`; this is mostly fine because we're adopting these as a convention, but we should provide a mechanism to override this, e.g. via optional parameters `obs_dim_name` and `traj_dim_name` or similar.

milancurcic:
Fixes #233.


selipot (approved):
I have not tested but the docstring looks great.

milancurcic:
This PR enables `subset` to work with ragged Spotter data in [this notebook](https://github.com/sofarocean/spotter-data-archive/blob/main/examples/plot_spotter_drift_latest.ipynb). I'll go ahead and merge.

PR #249, opened 2023-09-06
Relicense to MIT

milancurcic:
Both licenses are the permissive open-source kind. Apache is longer and more complex, so for users for whom licenses are important, it may be more difficult to read and understand. Apache-licensed software is slightly more difficult to distribute when modified because it requires documenting state changes. Patent use permission makes it incompatible with GPLv2. These are all minor differences, but for us the cost to relicense is minimal (10 minutes of work that I already spent).

Summary below (I highlighted the differences in Apache relative to MIT)

MIT (proposed):

Permissions: Commercial use, Modification, Distribution, Private use
Limitations: Liability, Warranty
Conditions: License and copyright notice

Apache 2.0 (current):

Permissions: Commercial use, Modification, Distribution, **Patent use**, Private use
Limitations: **Trademark use**, Liability, Warranty
Conditions: License and copyright notice, **State changes**

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,201 +1,21 @@
-                                 Apache License
-                           Version 2.0, January 2004
-                        http://www.apache.org/licenses/
+MIT License
 
-   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+Copyright (c) 2022-2023 CloudDrift contributors
 
-   1. Definitions.
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
 
-      "License" shall mean the terms and conditions for use, reproduction,
-      and distribution as defined by Sections 1 through 9 of this document.
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
 
-      "Licensor" shall mean the copyright owner or entity authorized by
-      the copyright owner that is granting the License.
-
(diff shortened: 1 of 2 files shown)


philippemiron:
I think I chose reading this ([source](https://www.exygy.com/blog/which-license-should-i-use-mit-vs-apache-vs-gpl)):

```
What Are You Afraid Of? Another way of looking at it is that you’re picking a license based on what you are afraid of. All of these licenses assume you’re afraid of being sued. The MIT license is if you’re afraid no one will use your code; you’re making the licensing as short and non-intimidating as possible. The Apache License you are somewhat afraid of no one using your code, but you are also afraid of legal ambiguity and patent trolls. With the GPL licenses, you are afraid of someone else profiting from your work (and ambiguity, and patent trolls). This is a radical simplification, but if nothing else it can be a helpful framework in discussing with your attorney what license makes sense for your software.
```

I wouldn't go with GPL as it's too restrictive, but MIT/Apache are fine with me.

philippemiron:
I remember I chose after reading this ([source](https://www.exygy.com/blog/which-license-should-i-use-mit-vs-apache-vs-gpl)):

`
What Are You Afraid Of? Another way of looking at it is that you’re picking a license based on what you are afraid of. All of these licenses assume you’re afraid of being sued. The MIT license is if you’re afraid no one will use your code; you’re making the licensing as short and non-intimidating as possible. The Apache License you are somewhat afraid of no one using your code, but you are also afraid of legal ambiguity and patent trolls. With the GPL licenses, you are afraid of someone else profiting from your work (and ambiguity, and patent trolls). This is a radical simplification, but if nothing else it can be a helpful framework in discussing with your attorney what license makes sense for your software.
`

I wouldn't go with GPL has it's too restrictive but MIT/Apache are fine with me.

philippemiron (approved):

milancurcic:
Thank you.

PR #252, opened 2023-09-07
Allow time to be 1-d if u,v or lon,lat are n-d

### LINKED ISSUE #230: allow different shape for velocity_from_position
I think that `velocity_from_position` and related functions should allow for a `time` variable input of different shape than the `x` and `y` input. It would suffice for `time` to be the length of the `time_axis` of `x` and `y`.

milancurcic:
Fixes #230.

### DIFF
--- a/clouddrift/analysis.py
+++ b/clouddrift/analysis.py
@@ -592,8 +592,9 @@ def position_from_velocity(
     Raises
     ------
     ValueError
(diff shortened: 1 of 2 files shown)


selipot (commented):
So there is no algebra change because of the broadcasting?

milancurcic:
I initially thought it would be needed but apparently not.

Issue #251, opened 2023-09-07
Add diagrams to API docs


milancurcic:
3 diagrams are available here: https://cloud-drift.github.io/earthcube-community-talk-2023/

PR #257, opened 2023-09-13
Diagrams_in_doc

philippemiron:
I will let you add some text and/or move those around. But here is a good start for #251.


philippemiron:
You need to build the docs locally, the instructions are here.

```
cd clouddrift/docs

conda install sphinx
pip install sphinx_book_theme
pip install sphinx-copybutton

make html
```

Then you can open `_build/html/index.html` in a browser.

selipot:
Thanks @philippemiron I feel you told me that before

philippemiron:
No worries.

docs/api.rst
```diff
@@ -38,7 +38,22 @@ Analysis
...
+      :width: 800
+      :align: center
+      :alt: chunk schematic
+    
+    Combined with :func:`clouddrift.analysis.apply_ragged`, :func:`clouddrift.analysis.chunk` can be used to divide a ragged array into equal chunks.
+    
+    .. image:: img/chunk_ragged.png
+      :width: 800
+      :align: center
+      :alt: ragged array chunk schematic
```
milancurcic, line 56: On what page do these render? I can't seem to find them and they don't appear in the API docs for chunk. (I'm a Sphinx n00b)

milancurcic: OK, it seems like they need to be placed in the docstrings.

milancurcic (approved):
Thanks for this PR, great addition.

milancurcic:
It has not occurred to me that, as @selipot just pointed out on our call, that having the RST code to insert an image in the Python docstring itself is unfriendly to users reading the docstring from the Python prompt. This is probably why Philippe attempted to place them in api.rst, though that didn't render the images for me. Let's find a way to include graphics in the function docs without polluting the docstring itself, perhaps by looking at how other projects do this (maybe Xarray? not sure).

philippemiron:
I'm not sure why it's not rendering on your end, but it's working fine for me. See,

image

philippemiron:
and I agree (I missed the previous commits) that we don't want to have this in the docstrings of the functions itself.

milancurcic:
> I'm not sure why it's not rendering on your end, but it's working fine for me. See,

I either initially did something wrong or I didn't scroll down enough to see them because they appear at the bottom of the docs. All good, I reverted, this is the way to go.

Issue #200, opened 2023-06-25
How to do a global spaghetti plot with trajectories crossing the dateline or edges on longitude range?


selipot:
How to do a global spaghetti plot with trajectories crossing the dateline or edges on longitude range? In jLab, this can be done simply with `cellplot(360,...)`.

milancurcic:
I haven't tried, but I think `segment` could be useful for this:

```python
rowsize = segment(ds.lon, 359, segment(ds.lon, -359))  # or some sensibly large tolerance
# use rowsize to plot contiguous segments or build LineCollections
```

philippemiron:
Yes, something like this should work with `ds.lon360`, otherwise setting the limit to +/-180°.

selipot:
This would create another segment and therefore a line of a different color if the color is randomly assigned by segment. How would the plotting keep the plotting characteristics for each interrupted trajectory?

philippemiron:
Like this:
- retrieve a trajectory
- choose a random color
- split into segments
- add to the segments list
- add len(segments) times the same color

milancurcic:
Another possible approach is to insert a NaN into the array between the two elements that are across the dateline. `plt.plot()` can't draw a line through a NaN.

selipot:
Let's implement this

milancurcic:
Which one?

selipot:
whichever works or makes sense

milancurcic:
OK. I think a function to insert a NaN based on a criterion is reasonable, and it stays away from plotting functionality which I think clouddrift should not deal with. I believe that jLab has a similar function to return a cell-array as a 1-d array with NaN elements as separators, and we discussed implementing it. This specific function could be a special case of that. I'll think about the API.

philippemiron:
I did this example today, we can insert the `nan` using `segments`. Do we need this as a function really?

```
import clouddrift as cd
import matplotlib.pyplot as plt
import numpy as np
import cartopy.crs as ccrs
import cartopy.feature as cfeature
from cartopy.mpl.ticker import LongitudeFormatter, LatitudeFormatter

# example of trajectory
lon = np.array([100, 170, -175, -123, -100, -80, -50, -30, 0, 30, 90, 160, -160, -140], dtype="float")[::-1]
lat = np.array([10, 20, 30, 40, 50, 60, 70, 80, 80, 80, 60, 60, 60, 50], dtype="float")

# find the indices to insert nan using segment
[...]

PR #258, opened 2023-09-13
start of plotting "help" module

### LINKED ISSUE #227: apply_ragged with cartopy
There seems to be an issue with using `apply_ragged` with `cartopy`. Please review and have a look at the notebook below where in the last cell I attempt to do just that:

https://github.com/selipot/mosaic-get-started/blob/main/get_started_with_mosaic_icedrift.ipynb

philippemiron:
- [x] clean up codes
- [x] wraps the import to matplotlib and cartopy to include an Optional dependencies warning
- [x] validate that a projection is passed if using a geoaxes
- [x] add docstring
- [x] create examples
- [x] tests

As of right now, the goals are:
- implement #200
- create a function to return a LineCollection to avoid using `apply_ragged()` with `matplotlib` functions (to fix #227).

### DIFF
--- a/clouddrift/__init__.py
+++ b/clouddrift/__init__.py
@@ -6,12 +6,8 @@ from clouddrift.raggedarray import RaggedArray
 import clouddrift.adapters
 import clouddrift.datasets
 import clouddrift.kinematics
+import clouddrift.plotting
 import clouddrift.ragged
 import clouddrift.signal
 import clouddrift.sphere
 import clouddrift.wavelet
-
-try:
-    import clouddrift.plotting
-except:
-    pass
--- a/clouddrift/__init__.py
+++ b/clouddrift/__init__.py
@@ -10,3 +10,8 @@ import clouddrift.ragged
 import clouddrift.signal
 import clouddrift.sphere
 import clouddrift.wavelet
+
+try:
+    import clouddrift.plotting
+except:
+    pass
--- a/clouddrift/__init__.py
+++ b/clouddrift/__init__.py
@@ -9,5 +9,4 @@ import clouddrift.kinematics
 import clouddrift.ragged
 import clouddrift.signal
 import clouddrift.sphere
-import clouddrift.plotting
 import clouddrift.wavelet
--- a/clouddrift/__init__.py
+++ b/clouddrift/__init__.py
@@ -9,5 +9,5 @@ import clouddrift.kinematics
 import clouddrift.ragged
 import clouddrift.signal
 import clouddrift.sphere
-import clouddrift.utility
+import clouddrift.plotting
 import clouddrift.wavelet
--- a/clouddrift/__init__.py
+++ b/clouddrift/__init__.py
@@ -9,4 +9,5 @@ import clouddrift.kinematics
 import clouddrift.ragged
 import clouddrift.signal
 import clouddrift.sphere
+import clouddrift.utility
 import clouddrift.wavelet
--- a/clouddrift/plotting.py
+++ b/clouddrift/plotting.py
@@ -53,8 +53,8 @@ def plot_ragged(
     *args : tuple
         Additional arguments to pass to ``ax.plot``.
     tolerance : float
-        Tolerance gap between data points (in degrees) for segmenting trajectories. For periodic
-        domains, the tolerance parameter should be set to the maximum allowed gap
+        Longitude tolerance gap between data points (in degrees) for segmenting trajectories.
+        For periodic domains, the tolerance parameter should be set to the maximum allowed gap
         between data points. Defaults to 180.
     **kwargs : dict
         Additional keyword arguments to pass to ``ax.plot``.
@@ -77,7 +77,6 @@ def plot_ragged(
     >>> fig = plt.figure()
     >>> ax = fig.add_subplot(1, 1, 1)
 
-    >>> time = [v.astype(np.int64) / 86400 / 1e9 for v in ds.time.values]
     >>> plot_ragged(
     >>>     ax,
     >>>     ds.lon,
@@ -89,7 +88,6 @@ def plot_ragged(
     To plot the same trajectories, but assigning a different color to each
     observation and specifying a colormap:
 
-    >>> time = [v.astype(np.int64) / 86400 / 1e9 for v in ds.time.values]
     >>> fig = plt.figure()
     >>> ax = fig.add_subplot(1, 1, 1)
     >>> lc = plot_ragged(
--- a/clouddrift/plotting.py
+++ b/clouddrift/plotting.py
@@ -66,6 +66,7 @@ def plot_ragged(
     a different color to each trajectory:
 
     >>> from clouddrift import datasets
+    >>> import matplotlib.pyplot as plt
     >>> ds = datasets.gdp1h()
     >>> ds = subset(ds, {"ID": ds.ID[:100].values}).load()
     >>> fig = plt.figure()
@@ -100,6 +101,7 @@ def plot_ragged(
     Finally, to plot the same trajectories, but using a cartopy
     projection:
 
+    >>> import cartopy.crs as ccrs
     >>> fig = plt.figure()
     >>> ax = fig.add_subplot(1, 1, 1, projection=ccrs.Mollweide())
     >>> time = [v.astype(np.int64) / 86400 / 1e9 for v in ds.time.values]
--- a/clouddrift/plotting.py
+++ b/clouddrift/plotting.py
@@ -84,6 +84,7 @@ def plot_ragged(
 
     >>> fig = plt.figure()
     >>> ax = fig.add_subplot(1, 1, 1)
+    >>> time = [v.astype(np.int64) / 86400 / 1e9 for v in ds.time.values]
     >>> lc = plot_ragged(
     >>>     ax,
     >>>     ds.lon,
--- a/clouddrift/plotting.py
+++ b/clouddrift/plotting.py
@@ -11,22 +11,14 @@ import pandas as pd
 from typing import Optional, Union
 from clouddrift.ragged import segment, rowsize_to_index
 
-try:
-    import matplotlib.pyplot as plt
-    import matplotlib.colors as mcolors
-    from matplotlib.collections import LineCollection
-    from matplotlib import cm
-except ImportError:
-    raise ImportError("missing optional dependency 'matplotlib'")
-
 
 def plot_ragged(
-    ax: plt.Axes,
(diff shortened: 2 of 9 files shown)


selipot:
Great start! maybe call the module `plothelp` or `helper` instead of `help` which is a bit too general?

Maybe the two goals could be combined? i.e. the function that returns the LineCollection returns extra lines when crossing the dateline to accommodate? How do we make sure that the assigned colors of the lines for the drifters that cross the dateline are the same?

milancurcic:
What do you think about `utility`? For the function that inserts NaN at dateline, I suggest making a function that inserts NaNs between segments, and let the user segment the dateline. That could fit in the `analysis` module. I don't quite see the need for a separate module here. I know @selipot suggested it and I didn't say anything, but I don't see it as compelling.

milancurcic:
> and let the user segment the dateline

We can then have a separate function that segments the dateline specifically, so the desired functionality here would be by chaining the two functions.

philippemiron:
I have no preference about the name of the module and/or the locations of those functions. 

I was going to make the first function to insert `nan` general enough to deal with any periodic domain, expanding on this:
```
nan_indices = np.cumsum(cd.analysis.segment(lon, 180, rowsize=cd.analysis.segment(lon, -180)))[:-1]
```

I can switch it to `utility.py` for now and we see where we put the functions once ready.

philippemiron:
Not sure exactly how to proceed now. `segment` only supports 1-d array... 😭

philippemiron:
@selipot want to test? I guess I might have some time next week during Thanksgiving to play with this.

philippemiron:
I'm abandoning the idea of _rebasing_ this branch. Many files have been renamed, deleted and functions moved around, making this very annoying. I might just reopened another PR once we are closer to merging this.

selipot:
> I'm abandoning the idea of _rebasing_ this branch. Many files have been renamed, deleted and functions moved around, making this very annoying. I might just reopen another PR once we are closer to merging this.

Sorry about that, that happens when a PR is open for too long for a version 0.x software I guess. Do you want me to test before you open another PR or after?

philippemiron:
No worries, the easiest would be just to copy the function.

philippemiron:
One question, how can I test something that requires _optional_ dependencies... 🤔 I guess we would need to have a different environment for the test, which would include all the _mandatory_ and _optional_ packages.

milancurcic:
How about installing all dependencies (core and optional) for testing?

selipot:
Testing your function. I think there is an issue with your first test lines 132-150. It raises a valueError if ax is a simple plt.Axes which it should as it is written. As an example this does not work:
```
fig = plt.figure()
ax = fig.add_subplot(1, 1, 1)
plot_ragged(ax, ds.lon, ds.lat, ds.rowsize, colors=np.arange(len(ds.rowsize)));
```
 
Screenshot 2023-11-16 at 8 18 03 AM

selipot:
The examples need editing which I can do once you submit a new PR.

philippemiron:
Did you try the latest version? You can use this branch, I fix it yesterday.

selipot:
I copied the code for the function from the PR branch as you instructed. Again, lines 132-150 don't make sense to me or rather the error I get makes sense. Why would you test for ax to have the attribute coastline if it is only plt.Axes?
```
if isinstance(ax, plt.Axes) and hasattr(
         ax, "coastlines"
     )
...
```

philippemiron:
If it has attributes coastline, it is a geoAxe and we have to also try to load cartopy. I'm on the plane now. But it was working for me yesterday. I will check again in ~2h.

selipot:
What's next with this @philippemiron ?

philippemiron:
Did you have chance to give it a try?

selipot:
Yes! Works for me. But did not test with cartopy.

philippemiron:
Ok, it was working fine with me. I will generate some tests later today. I think I should be able to test that it is wokring without `cartopy` and returning proper exception without `matplotlib`.

clouddrift/utility.py
```diff
@@ -0,0 +1,207 @@
...
+    latitude : array-like
+        Latitude sequence. Unidimensional array input.
+    rowsize : list
+        List of integers specifying the number of data points in each row.
+    colors : array-like
+        Colors to use for plotting. If colors is the same shape as longitude and latitude,
+        the trajectories are splitted into segments and each segment is colored according
+        to the corresponding color value. If colors is the same shape as rowsize, the
+        trajectories are uniformly colored according to the corresponding color value.
+    *args : tuple
```
selipot, line 49: Isn't it *kwargs?

clouddrift/utility.py
```diff
@@ -0,0 +1,207 @@
...
+        Additional arguments to pass to `plt.plot`.
+    tolerance : float
+        Tolerance gap between data points (in degrees) for segmenting trajectories. For periodic
```
selipot, line 52: Tolerance for longitude?

clouddrift/utility.py
```diff
@@ -0,0 +1,207 @@
...
+    Plot the first 100 trajectories from the gdp1h dataset, assigning
+    a different color to each trajectory:
+
+    >>> from clouddrift import datasets
+    >>> ds = datasets.gdp1h()
+    >>> ds = subset(ds, {"ID": ds.ID[:100].values}).load()
+    >>> fig = plt.figure()
+    >>> ax = fig.add_subplot(1, 1, 1)
+
+    >>> time = [v.astype(np.int64) / 86400 / 1e9 for v in ds.time.values]
```
selipot, line 76: Why is time defined here? What the use?

clouddrift/utility.py
```diff
@@ -0,0 +1,207 @@
...
+    >>> fig.colorbar(lc[0])
+    >>> ax.set_xlim([-180, 180])
+    >>> ax.set_ylim([-90, 90])
+
+    Finally, to plot the same trajectories, but using a cartopy
+    projection:
+
+    >>> fig = plt.figure()
+    >>> ax = fig.add_subplot(1, 1, 1, projection=ccrs.Mollweide())
+    >>> time = [v.astype(np.int64) / 86400 / 1e9 for v in ds.time.values]
```
selipot, line 108: Why is time defined here?

selipot (changes requested):

clouddrift/utility.py
```diff
@@ -0,0 +1,211 @@
+"""
+This module contains various utility functions.
```
milancurcic, line 2: I get the long-term vision but while there's only one function let's be specific and precise.

```suggestion
This module provides a function to easily and efficiently plot trajectories stored in a ragged array.
```

clouddrift/utility.py
milancurcic: `**kwargs` are below. LGTM.

tests/utility_tests.py
```diff
@@ -0,0 +1,149 @@
+import cartopy.crs as ccrs
+from clouddrift.utility import plot_ragged
+import matplotlib.pyplot as plt
```
milancurcic, line 3: Is there a mechanism to skip these tests if optional dependencies are not installed?

milancurcic (approved):
This looks good to go to me. I installed it, plotted some stuff, looks good. Tests pass after installing matplotlib and cartopy.

`utility` as a module name is quite generic. Should we call this `plotting`?

tests/utility_tests.py
philippemiron: True, I assumed that it was installed in the ci/cd, but I see that this wouldn't work if you don't have those packages locally.

clouddrift/utility.py
philippemiron: What is the question?

milancurcic: "Longitude tolerance gap between..." or "Tolerance gap for longitude between..."

milancurcic:
I forgot one important thing: we should add a note to the installation instructions (in README.md and docs/install.rst) about installing optional dependencies.

milancurcic:
I think we discussed and established that matplotlib-core is already implicitly installed by depending on xarray, correct? In that case, I think only cartopy should be an optional dependency.

tests/utility_tests.py
philippemiron: Can you try again with the latest modification?

philippemiron:
If I installed the environment I don't get matplotlib.

philippemiron:
> I think we discussed and established that matplotlib-core is already implicitly installed by depending on xarray, correct? In that case, I think only cartopy should be an optional dependency.

do we really need this? If someone tries to import the module it already say: 
```
raise ImportError("missing optional dependency 'matplotlib'")
```

clouddrift/__init__.py
```diff
@@ -9,5 +9,4 @@
 import clouddrift.ragged
 import clouddrift.signal
 import clouddrift.sphere
-import clouddrift.plotting
```
milancurcic: Is it not required so that we can do

```
import clouddrift as cd
cd.plotting.plot_ragged(...)
```

?

philippemiron: I added a `try: except:` ...

milancurcic:
@selipot please check on your end if all good, and merge when ready.

clouddrift/plotting.py
```diff
@@ -0,0 +1,212 @@
...
+    observation and specifying a colormap:
+
+    >>> fig = plt.figure()
+    >>> ax = fig.add_subplot(1, 1, 1)
+    >>> lc = plot_ragged(
+    >>>     ax,
+    >>>     ds.lon,
+    >>>     ds.lat,
+    >>>     ds.rowsize,
+    >>>     colors=np.floor(time),
```
selipot, line 92: `time` is not defined for this example. @philippemiron what did you intend it to be?

clouddrift/plotting.py
```diff
@@ -0,0 +1,212 @@
...
+    >>> )
+    >>> fig.colorbar(lc[0])
+    >>> ax.set_xlim([-180, 180])
+    >>> ax.set_ylim([-90, 90])
+
+    Finally, to plot the same trajectories, but using a cartopy
+    projection:
+
+    >>> fig = plt.figure()
+    >>> ax = fig.add_subplot(1, 1, 1, projection=ccrs.Mollweide())
```
selipot, line 103: Should not there be a `import ccrs` command in this example?

selipot (changes requested):
just small editing of the docstring examples needed.

selipot (approved):

PR #268, opened 2023-09-17
add wavelet module to doc?

selipot:
Is that the way to add automatically the wavelet module to the documentation? It is currently missing.


selipot:
oops doing PR from my main branch but i think that is ok?

philippemiron:
I guess it's fine since it's only doc stuff.

philippemiron (approved):

philippemiron:
PS: the doc is only regenerated when we push a new release. So merging this will not change the website right away.

PR #269, opened 2023-09-18
Cast GDP float64 data to float32 as an option (default)

milancurcic:
This casts all float64 variables to float32 before returning the Dataset in the `RaggedArray.to_xarray()` method, but give an option to not do it if desired. I tested it with a small gdp-2.01 file (10 random drifters) and it works.

Can you confirm that float32 is OK for longitude and latitude variables? We get precision of 7 digits. For a drifter in the 3-digit longitudes, precision is down to O(10 m). Is it sufficient?

### DIFF
--- a/clouddrift/adapters/gdp1h.py
+++ b/clouddrift/adapters/gdp1h.py
@@ -510,6 +510,9 @@ def preprocess(index: int, **kwargs) -> xr.Dataset:
     # rename variables
     ds = ds.rename_vars({"longitude": "lon", "latitude": "lat"})
 
+    # Cast float64 variables to float32 to reduce memory footprint.
+    ds = gdp.cast_float64_variables_to_float32(ds)
+
     return ds
 
 
--- a/clouddrift/raggedarray.py
+++ b/clouddrift/raggedarray.py
@@ -386,7 +386,8 @@ class RaggedArray:
         Parameters
         ----------
         cast_to_float32 : bool, optional
-            Cast all variables to float32 (default is True)
+            Cast all float64 variables to float32 (default is True). This option aims at
+            minimizing the size of the xarray dataset.
 
         Returns
         -------
--- a/clouddrift/raggedarray.py
+++ b/clouddrift/raggedarray.py
@@ -380,8 +380,13 @@ class RaggedArray:
             if key not in self.attrs_variables:
                 self.attrs_variables[key] = {}
 
-    def to_xarray(self):
+    def to_xarray(self, cast_to_float32: bool = True):
         """Convert ragged array object to a xarray Dataset.
+
+        Parameters
+        ----------
+        cast_to_float32 : bool, optional
+            Cast all variables to float32 (default is True)
 
         Returns
         -------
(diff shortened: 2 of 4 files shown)


clouddrift/raggedarray.py
```diff
@@ -380,9 +380,14 @@ def validate_attributes(self):
...
                 self.attrs_variables[key] = {}
 
-    def to_xarray(self):
+    def to_xarray(self, cast_to_float32: bool = True):
         """Convert ragged array object to a xarray Dataset.
 
+        Parameters
+        ----------
+        cast_to_float32 : bool, optional
+            Cast all variables to float32 (default is True)
```
selipot, line 389: Perhaps change to: "Cast all float64 variables to float32 (Default is True). This option aims at minimizing the size of the output."

selipot: Does this change the size of the result in the active memory or only when written to disk or both?

selipot (changes requested):
Good to go for me for now. A user who would want to keep some variables float64 would have to pass False and do the conversion manually. I think this is acceptable.

clouddrift/raggedarray.py
milancurcic: Can you use the suggestion feature so we can commit the change here directly?

Yes, unless there is bug with xarray's memory management, both size in memory and on disk will be reduced.

selipot:
I was able to discuss this a bit more with @JeffreyEarly : we should keep double precision for latitude and velocity estimates but we can convert to single precision for all other variables. (Even for uncertainty of longitude and latitude.) 

Since this PR applies to the whole raggedarray class I think it is not a good idea in the end. We should keep double precision.

milancurcic:
Sounds good, we can exclude specific variables from the cast. Do you know why double for velocity?

selipot:
> Sounds good, we can exclude specific variables from the cast. Do you know why double for velocity?

That was a typo: latitude and longitude, now corrected.

selipot:
So in conclusion, despite my commit, I don't think we should implement this PR.

milancurcic:
I'm confused. Why not cast variables other than lat and lon?

selipot:
> I'm confused. Why not cast variables other than lat and lon?

If I understand correctly, this PR proposes to convert all variables to float32 (not only in the GDP cases!). Some users might want to retain the double precision for their specific variables. So I am not in favor of this behavior.

milancurcic:
OK, if you prefer manually casting that's fine. Better than adding code that won't be useful.

As an aside as since it came up here, the RaggedArray class is currently only used for GDP. I can't imagine using it for other datasets over simple functions like I did in MOSAiC. Long term I'd like to remove it altogether; nothing to do for now.

philippemiron:
I'm confused by this. So we are doing nothing and the GDP ragged array will be 35 GB? I don't think it makes much sense to have all this precision.

selipot:
26GB not 35GB. We are doing the conversion by hand for the GDP, not by a default behavior of `raggedarray.to_netcdf()`. The manual conversion leads to ~16GB.

philippemiron:
But it could be done in `adapters.gdp1h` and `adapters.gdp6h` so we don't have to do anything manually (and limit possible divergence) when regenerating the GDP ragged array.

milancurcic:
It sounds like the current direction is:

* Move casting to `adapters.gdp1h` and `adapters.gdp6h`
* Exclude `time`, `latitude`, and `longitude` from the cast

Please confirm.

selipot:
Sounds like a plan.

clouddrift/adapters/gdp1h.py
```diff
@@ -510,6 +510,9 @@ def preprocess(index: int, **kwargs) -> xr.Dataset:
     # rename variables
     ds = ds.rename_vars({"longitude": "lon", "latitude": "lat"})
 
+    # Cast float64 variables to float32 to reduce memory footprint.
+    ds = gdp.cast_float64_variables_to_float32(ds)
```
selipot, line 514: Is this where you should add variables_to_skip = ["lon", "lat", "time"]?`

selipot (changes requested):

selipot:
Once this is done I will be able to get the final parquet file.

clouddrift/adapters/gdp1h.py
milancurcic: That's the default, see the function definition.

milancurcic:
@selipot Can this be merged?

selipot (approved):

selipot:
yes!

PR #272, opened 2023-09-20
Add optional `rows` parameter to `unpack_ragged` and `apply_ragged`

### LINKED ISSUE #222: indexing of segments
### Discussed in https://github.com/Cloud-Drift/clouddrift/discussions/208

<div type='discussions-op-text'>

<sup>Originally posted by **selipot** June 28, 2023</sup>
I am wondering if we could improve indexing within rows. Let's say we start from a ragged array dataset `ds` with a given `rowsize`, we first need to generate a row index, define a slice, then index that slice:

```
row_index = np.insert(np.cumsum(ds.rowsize.values), 0, 0) # define row index
j = 1
row = slice(row_index[j], row_index[j+1]) # define a slice for row 1
ds.lon[row][n] # access the first data point of that row
```

How can we simply this to something like `ds.lon[j][n]` which is basically a syntax/functionality that is supported by awkward arrays but not xarray. 

Is one long haul solution to define new classes such `xarray.RaggedDataArray` and `xarray.RaggedDataset` for which we could have `.isel` and `.sel` methods for double indexing (`ds.lon[j][n]`)?
</div>

milancurcic:
Closes #222.

### DIFF
--- a/clouddrift/analysis.py
+++ b/clouddrift/analysis.py
@@ -18,6 +18,7 @@ def apply_ragged(
     arrays: list[np.ndarray],
     rowsize: list[int],
     *args: tuple,
+    rows: Union[int, Iterable[int]] = None,
     executor: futures.Executor = futures.ThreadPoolExecutor(max_workers=None),
     **kwargs: dict,
 ) -> Union[tuple[np.ndarray], np.ndarray]:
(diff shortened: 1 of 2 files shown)


clouddrift/analysis.py
```diff
@@ -88,7 +101,7 @@ def apply_ragged(
             raise ValueError("The sum of rowsize must equal the length of arr.")
 
     # split the array(s) into trajectories
-    arrays = [unpack_ragged(arr, rowsize) for arr in arrays]
+    arrays = [unpack_ragged(arr, rowsize, rows) for arr in arrays]
```
philippemiron, line 104: oh, very nice. I was wondering how this will integrate with `apply_ragged`.

selipot (approved):

milancurcic:
Thank you!

PR #278, opened 2023-09-22
subpages in the doc

philippemiron:
I played (way too long..) with autosummary this evening. It does generate a doc similar to `scipy` much easier to browse. Only thing I haven't try yet to include the figure after `chunk`.

There are also a lot of warnings when building the docs and I believe it's due to the way we import in `__init__.py` that conflicts with the `::recursive::` option.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -67,6 +67,7 @@ coverage.xml
 
 # Sphinx documentation
 docs/_build/
+docs/_autosummary/
 
 # PyBuilder
 target/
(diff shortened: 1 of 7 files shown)


philippemiron:
image
image
image

selipot:
WOW @philippemiron !!!

milancurcic:
It looks awesome.

philippemiron:
I think we should be able to create a special `_templates` for the `analysis` module and inject the `chunk` documentation, but it was not working when I tried.

milancurcic (approved):
Awesome, thank you. Two critical warnings had to do with wavelet.py docstrings formatting. They're gone now. The remaining warnings don't seem two concerning to me, they all seem to have to do with double definition of functions, one of them being autogenerated by Sphinx. Let's figure out where to put chunk diagram in a future PR. I built and tested the docs locally and the autosummaries work and look great.

selipot:
What else need to be done to merge this?

PR #281, opened 2023-09-22
Add `axis` to `apply_ragged` and `unpack_ragged`

### LINKED ISSUE #274: Specify concatenation `axis` to `apply_ragged`

milancurcic:
This PR:

* Adds `axis` to `apply_ragged` and `unpack_ragged` to specify unpacking/concatenation axis; default is 0, which is consistent with `np.concatenate` and `np.split`.
* Fixes a few edge cases that didn't work before:
  - If you use a reduction function over a non-ragged axis, passing `xr.DataArray`s now works.
  - If you use a reduction function over a non-ragged axis and return more than one result (tuple), concatenation now works.
* Unrelated to this PR, there was an outstanding issue with a test for `coriolis_frequency` that tested for a Warning that shouldn't be there. I don't know why tests were passing on `main` but this was failing for me. It's fixed.
* Minor version bump to 0.23.0

Closes #274.

### DIFF
--- a/clouddrift/analysis.py
+++ b/clouddrift/analysis.py
@@ -19,6 +19,7 @@ def apply_ragged(
     rowsize: list[int],
     *args: tuple,
     rows: Union[int, Iterable[int]] = None,
+    axis: int = 0,
     executor: futures.Executor = futures.ThreadPoolExecutor(max_workers=None),
     **kwargs: dict,
 ) -> Union[tuple[np.ndarray], np.ndarray]:
@@ -49,6 +50,9 @@ def apply_ragged(
     rows : int or Iterable[int], optional
         The row(s) of the ragged array to apply ``func`` to. If ``rows`` is
         ``None`` (default), then ``func`` will be applied to all rows.
+    axis : int, optional
+        The axis along which to unpack the input ragged array(s) and concatenate
+        the outputs of ``func`` before returning the result. Default is 0.
     executor : concurrent.futures.Executor, optional
         Executor to use for concurrent execution. Default is ``ThreadPoolExecutor``
         with the default number of ``max_workers``.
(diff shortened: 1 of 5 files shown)


philippemiron (approved):

clouddrift/analysis.py
```diff
@@ -40,15 +44,18 @@ def apply_ragged(
...
-    rowsize : list
+    rowsize : list[int] or np.ndarray[int] or xr.DataArray[int]
         List of integers specifying the number of data points in each row.
     *args : tuple
         Additional arguments to pass to ``func``.
     rows : int or Iterable[int], optional
         The row(s) of the ragged array to apply ``func`` to. If ``rows`` is
         ``None`` (default), then ``func`` will be applied to all rows.
+    axis : int, optional
+        The axis along which to unpack the input ragged array(s) and concatenate
```
selipot, line 57: The axis of the input ragged arrays that is ragged?

clouddrift/analysis.py
```diff
@@ -40,15 +44,18 @@ def apply_ragged(
...
-    arrays : list[np.ndarray] or np.ndarray
+    arrays : list[np.ndarray] or np.ndarray or xr.DataArray
         An array or a list of arrays to apply ``func`` to.
-    rowsize : list
+    rowsize : list[int] or np.ndarray[int] or xr.DataArray[int]
         List of integers specifying the number of data points in each row.
     *args : tuple
         Additional arguments to pass to ``func``.
     rows : int or Iterable[int], optional
         The row(s) of the ragged array to apply ``func`` to. If ``rows`` is
```
selipot, line 54: The fact that the default `None` means apply to all rows is very counter-intuitive to me.

clouddrift/analysis.py
```diff
@@ -28,6 +29,9 @@ def apply_ragged(
     indicated by row sizes ``rowsize``. The output of ``func`` will be
     concatenated into a single ragged array.
 
+    You can pass ``arrays`` as NumPy arrays or xarray DataArrays, however,
+    the result will always be a NumPy array.
+
```
selipot, line 34: Could the main docstring mention the rows argument?

selipot (changes requested):

clouddrift/analysis.py
milancurcic: I can see it being counter-intuitive because "None" for "rows" implies "apply to no rows". However, this is a Python quirk--`None` is a special word that in this case means that the `rows` argument. Though at the risk of being counter-intuitive at first, it is _the_ Pythonic solution.

clouddrift/analysis.py
milancurcic: Can you expand? I don't understand. You can also use the suggest feature to apply a specific edit.

selipot: I suggest to rephrase the definition of this argument to something like: "The axis of the input array that is ragged". The user does not necessarily know what unpack means in this context or why you mention concatenate etc.

milancurcic:
@selipot Please check the docstring and merge when ready.

selipot (approved):

PR #282, opened 2023-09-26
Gdp-update

selipot:
Modification of the GDP adapters function to reflect the hourly database update.

To merge once the DAC update completely the 2.01/ FTP folder.

Should the global variable be moved from gdp.py to gdp1h.py since it applies only to that product?

### DIFF
--- a/clouddrift/adapters/gdp1h.py
+++ b/clouddrift/adapters/gdp1h.py
@@ -71,7 +71,7 @@ def download(
     Returns
     -------
     out : list
-        List of retrived drifters
+        List of retrieved drifters
     """
 
     print(f"Downloading GDP hourly data from {url} to {tmp_path}...")
--- a/clouddrift/adapters/gdp1h.py
+++ b/clouddrift/adapters/gdp1h.py
@@ -81,7 +81,7 @@ def download(
 
     if url == GDP_DATA_URL:
         pattern = "drifter_hourly_[0-9]*.nc"
-        filename_pattern = "drifter_{id}.nc"
+        filename_pattern = "drifter_hourly_{id}.nc"
     elif url == GDP_DATA_URL_EXPERIMENTAL:
         pattern = "drifter_hourly_[0-9]*.nc"
         filename_pattern = "drifter_hourly_{id}.nc"
@@ -586,7 +586,7 @@ def to_raggedarray(
     ids = download(drifter_ids, n_random_id, url, tmp_path)
 
     if url == GDP_DATA_URL:
-        filename_pattern = "drifter_{id}.nc"
+        filename_pattern = "drifter_hourly_{id}.nc"
     elif url == GDP_DATA_URL_EXPERIMENTAL:
         filename_pattern = "drifter_hourly_{id}.nc"
     else:
--- a/clouddrift/adapters/gdp1h.py
+++ b/clouddrift/adapters/gdp1h.py
@@ -25,6 +25,7 @@ GDP_DATA_URL_EXPERIMENTAL = (
     "https://www.aoml.noaa.gov/ftp/pub/phod/lumpkin/hourly/experimental/"
 )
 GDP_TMP_PATH = os.path.join(tempfile.gettempdir(), "clouddrift", "gdp")
+GDP_TMP_PATH_EXPERIMENTAL = os.path.join(tempfile.gettempdir(), "clouddrift", "gdp_exp")
 GDP_DATA = [
     "lon",
     "lat",
@@ -52,7 +53,7 @@ def download(
     drifter_ids: list = None,
     n_random_id: int = None,
     url: str = GDP_DATA_URL,
-    tmp_path: str = GDP_TMP_PATH,
+    tmp_path: str = None,
 ):
     """Download individual NetCDF files from the AOML server.
 
@@ -75,8 +76,8 @@ def download(
     """
 
     # adjust the tmp_path if using the experimental source
-    if url == GDP_DATA_URL_EXPERIMENTAL and tmp_path == GDP_TMP_PATH:
-        tmp_path += "_experimental"
+    if tmp_path is None:
+        tmp_path = GDP_TMP_PATH if url == GDP_DATA_URL else GDP_TMP_PATH_EXPERIMENTAL
 
     print(f"Downloading GDP hourly data from {url} to {tmp_path}...")
 
@@ -589,8 +590,8 @@ def to_raggedarray(
     """
 
     # adjust the tmp_path if using the experimental source
-    if url == GDP_DATA_URL_EXPERIMENTAL and tmp_path == GDP_TMP_PATH:
-        tmp_path += "_experimental"
+    if tmp_path is None:
+        tmp_path = GDP_TMP_PATH if url == GDP_DATA_URL else GDP_TMP_PATH_EXPERIMENTAL
 
     ids = download(drifter_ids, n_random_id, url, tmp_path)
 
--- a/clouddrift/adapters/gdp1h.py
+++ b/clouddrift/adapters/gdp1h.py
@@ -18,6 +18,7 @@ import os
 import warnings
 import xarray as xr
 
+GDP_VERSION = "2.01"
 
 GDP_DATA_URL = "https://www.aoml.noaa.gov/ftp/pub/phod/lumpkin/hourly/v2.01/netcdf/"
 GDP_DATA_URL_EXPERIMENTAL = (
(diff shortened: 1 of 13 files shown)


selipot:
Not sure why this PR contains old commits?

milancurcic:
> Should the global variable be moved from gdp.py to gdp1h.py since it applies only to that product?

Yes. So the 6-hourly GDP is a completely different version track, i.e. we don't expect the 6-hourly GDP to follow the hourly one in terms of dataset updates/versions?

The 2.01 upstream data URL seems to contain NetCDF files. Is that still a generation in progress or we can actually test and (if no problems) merge?

selipot:
> > Should the global variable be moved from gdp.py to gdp1h.py since it applies only to that product?
> 
> Yes. So the 6-hourly GDP is a completely different version track, i.e. we don't expect the 6-hourly GDP to follow the hourly one in terms of dataset updates/versions?

Correct, 6-hourly is on a different version track.

> The 2.01 upstream data URL seems to contain NetCDF files. Is that still a generation in progress or we can actually test and (if no problems) merge?

I sent email to @RickLumpkin with @milancurcic copied. He is working on organizing the 2.01/ folder just like the 2.00/ folder so hold on.

milancurcic:
Do you think it's worth it to test processing it just in case? (I won't be able to do it before Monday) Otherwise, looks good to me!

milancurcic (approved):

milancurcic:
We should also soon discuss how do we want to treat different dataset versions from the `clouddrift.datasets` point of view.

selipot:
> Do you think it's worth it to test processing it just in case? (I won't be able to do it before Monday) Otherwise, looks good to me!

Yes, we need to test.

selipot:
> We should also soon discuss how do we want to treat different dataset versions from the `clouddrift.datasets` point of view.

My proposition is to supply the latest by default but optionally supply previous ones.

selipot:
For `clouddrift.datasets.gdp1h()` should an option be to open the dataset from the zarr or nc file?
- netcdf dependencies can be problematic from my experience so this favors zarr
- the zarr file has shown issues for low bandwidth so this favors netcdf

Thoughts?

milancurcic:
nc access was not yet supported by the NetCDF-C version that shipped with netCDF4 on PyPI, so Zarr was the only option when I introduced the datasets module.

philippemiron (commented):
I've tested it and this fails:
```
import clouddrift as cd
ra = cd.adapters.gdp1h.to_raggedarray(n_random_id=10)
```

Adding the experimental URL works.
```
import clouddrift as cd
from clouddrift.adapters.gdp1h import GDP_DATA_URL_EXPERIMENTAL
ra = cd.adapters.gdp1h.to_raggedarray(n_random_id=10, url=GDP_DATA_URL_EXPERIMENTAL)
```

Should they update the metadata and remove the experimental URL if we are going to switch to v2.01?

selipot:
Do you understand why it fails? 

@RickLumpkin is now populating the experimental folder with the files for the 2.02 update. So moving forward the final files for each update/version will be in 2.xx/ and the files for the on-going update will be in experimental/

The idea is that the clouddrift library will update its code to reflect the data updates.

philippemiron:
It's failing because of the `filename_pattern` (I think) that's in `download()` and `to_raggedarray()`.

```
    if url == GDP_DATA_URL:
        pattern = "drifter_hourly_[0-9]*.nc"
        filename_pattern = "drifter_{id}.nc"
    elif url == GDP_DATA_URL_EXPERIMENTAL:
        pattern = "drifter_hourly_[0-9]*.nc"
        filename_pattern = "drifter_hourly_{id}.nc"
```

I don't have time to check now, but I guess it's just a matter of `"drifter_{id}.nc"` to `"drifter_hourly_{id}.nc"`. And actually, maybe we don't need that condition anymore since both situations the pattern will be the same.

selipot:
> It's failing because of the `filename_pattern` (I think) that's in `download()` and `to_raggedarray()`.
> 
> ```
>     if url == GDP_DATA_URL:
>         pattern = "drifter_hourly_[0-9]*.nc"
>         filename_pattern = "drifter_{id}.nc"
>     elif url == GDP_DATA_URL_EXPERIMENTAL:
>         pattern = "drifter_hourly_[0-9]*.nc"
>         filename_pattern = "drifter_hourly_{id}.nc"
> ```
> 
> I don't have time to check now, but I guess it's just a matter of `"drifter_{id}.nc"` to `"drifter_hourly_{id}.nc"`. And actually, maybe we don't need that condition anymore since for both `v2.01` and `v2.02` the pattern is the same.

Yes, my last commit fixes that. But down the line there are more issues which I do not understand. I have no time either to pursue this today and will look next when i can.

philippemiron:
I fix the fix. You missed two more patterns.

clouddrift/adapters/gdp1h.py
```diff
@@ -79,8 +80,8 @@ def download(
     os.makedirs(tmp_path, exist_ok=True)
 
     if url == GDP_DATA_URL:
```
selipot, line 82: This is not useful anymore but should we keep it to bring awareness of some possible changes upstream?

clouddrift/adapters/gdp1h.py
```diff
@@ -547,7 +548,7 @@ def to_raggedarray(
     --------
 
     Invoke `to_raggedarray` without any arguments to download all drifter data
-    from the 2.00 GDP feed:
+    from the 2.01 GDP feed:
```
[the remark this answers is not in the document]
philippemiron: I don't think that's possible. The  doc is also generated from the source code and it would have to be interpreted to get the value..

clouddrift/adapters/gdp1h.py
philippemiron: I kept it in case eventually there is a difference with the two datasets. If we are sure it is not going to change I would remove it.

philippemiron:
PS: there might be a more elegant way of doing this, but this should do it.

selipot:
I have now tested the generation of ragged arrays for the entire content of the standard folder (2.01/) and the experimental folder for this PR. I have also tested the creation of ragged arrays for only 100 random ids in both cases and it seems that the creation of two different download folders for these cases is the way to go. 

The one thing I have noticed is that the coordinate `time` is not as before a datetime object but a float variables which is the number of seconds since 1970-01-01. Is this intentional? Not sure what I prefer.

philippemiron:
Humm. It might not be converted at the time of the ragged array creation. Although if you save it to netCDF and reload it, does it work?

philippemiron:
So I think there is no bug.
```
import clouddrift as cd
import xarray as xr
ds = cd.adapters.gdp1h.to_raggedarray(n_random_id=10).to_xarray()
```

if you do this, the values are returned in floats. But if you just do:

```
ds.to_netcdf("test.nc")
ds2 = xr.open_dataset("test.nc")
```

You can use this,
```
ds = xr.decode_cf(ds, decode_times=True)
```
to convert the time variable after we create the ragged arrays. We could add this to our `.to_xarray()` method.

milancurcic:
As @philippemiron hinted, the difference between float and datetime64 times for GDP has to do with data representation after the dataset is loaded. In both scenarios the data in file is stored as float seconds.

So the question is really what should be the default (decode or don't decod) for `RaggedArray.to_xarray()` and for `datasets.gdp*h()` (It's decoded there by default).

I personally like simple float seconds because they're simple, but I know many people who like `np.datetime64[ns]`.

Ideally, we should be consistent between adapters and datasets modules.

An important gotcha in context of clouddrift is that if you pass time in float seconds to `velocity_from_position`, you get velocities in m/s. If you pass time as `np.datetime64[ns]` you get velocities in m/ns. There won't be a warning because both are valid and expected. But the user may be surprised about the result (I was).

selipot:
> As @philippemiron hinted, the difference between float and datetime64 times for GDP has to do with data representation after the dataset is loaded. In both scenarios the data in file is stored as float seconds.
> 
> So the question is really what should be the default (decode or don't decod) for `RaggedArray.to_xarray()` and for `datasets.gdp*h()` (It's decoded there by default).
> 
> I personally like simple float seconds because they're simple, but I know many people who like `np.datetime64[ns]`.
> 
> Ideally, we should be consistent between adapters and datasets modules.

Ok I confirm that I can reload also the ragged array I created with or without decoding the times!

I am not a fan of the datetime objects so my suggestion is to keep seconds as floats when creating the ragged arrays, if I understand correctly. And make it consistent across the data adapters and the data accessors if possible.
 
> An important gotcha in context of clouddrift is that if you pass time in float seconds to `velocity_from_position`, you get velocities in m/s. If you pass time as `np.datetime64[ns]` you get velocities in m/ns. There won't be a warning because both are valid and expected. But the user may be surprised about the result (I was).

I am fine with this. I think it is reasonable to expect the user to 1) read the doc and 2) know what they input as argument.

philippemiron:
If you don't like the `datetime` object, you can always just do: `xr.open_dataset("file.nc", decode_times=False)` to get them in the original `float` values.

philippemiron:
What's holding this?

PR #286, opened 2023-09-30
extract inertial oscillations

selipot:
Draft PR for a new function `extract_inertial_from_position` that estimates inertial oscillations from longitude and latitude time series. Python translation of matlab method by Jonathan Lilly. 

This PR also contains another function `correct_position_for_displacement` which could be placed in sphere module? 

Tests for `extract_inertial_from_position` need to be devised.

Open to name change for functions!


clouddrift/analysis.py
```diff
@@ -1159,3 +1169,219 @@ def unpack_ragged(
         rows = [rows]
 
     return [ragged_array[indices[n] : indices[n + 1]] for n in rows]
+
+
+def extract_inertial_from_position(
+    longitude: np.ndarray,
+    latitude: np.ndarray,
+    p: int,
```
milancurcic, line 1177: You already know what I'm going to say about this `p` here.

Needs a descriptive name. Docstring says duration of wavelet (so, some kind of time?) but then it says it must be one or greater and it's an integer. Is it some kind of order or number of wavelets to use?

clouddrift/analysis.py
```diff
@@ -1159,3 +1169,219 @@ def unpack_ragged(
...
+                "omega must be a float or the same shape as time, lon, and lat."
+            )
+
+    # convert time interval of data in seconds
+    # dt = np.diff(time)[0] / np.timedelta64(1000000000, "ns")  # seconds no unit
+
+    # wavelet parameters
+    gamma = 3
+    density = 16  # JML uses 16
+    beta = p**2 / gamma
```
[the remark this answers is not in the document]
milancurcic: Of course, it's OK for an initial information to just use sane defaults and not provide all the knobs for the users to tune, and wait to see if requests for flexibility come up.

clouddrift/analysis.py
```diff
@@ -1159,3 +1169,219 @@ def unpack_ragged(
         rows = [rows]
 
     return [ragged_array[indices[n] : indices[n + 1]] for n in rows]
+
+
+def extract_inertial_from_position(
+    longitude: np.ndarray,
+    latitude: np.ndarray,
+    p: int,
+    delta_time: Optional[float] = 3600.0,
```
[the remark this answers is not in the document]
selipot: Because the method is based on wavelet transform and fft, it works only for data evenly distributed in time. I do not think that the method should incorporate a linear interpolation in case of uneven data because I think that should be left to the user to prepare the data as they want/need/can.
 
I would like to offer the flexibility for the method to take timedelta objects for the argument `time_delta` but I do not know how to best deal with this, any advice?

clouddrift/analysis.py
```diff
@@ -1159,3 +1169,219 @@ def unpack_ragged(
         rows = [rows]
 
     return [ragged_array[indices[n] : indices[n + 1]] for n in rows]
+
+
+def extract_inertial_from_position(
```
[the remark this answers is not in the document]
selipot: Great! I did not know that convention. `inertial_oscillations_from_positions` is good. Should it be: oscillation/position or oscillationsS/positionS?

clouddrift/analysis.py
selipot: Sure, I can change that one.

clouddrift/analysis.py
```diff
@@ -1159,3 +1169,219 @@ def unpack_ragged(
...
+        The constant time interval between data points in seconds. Default is 3600 s.
+    omega: Optional, float or array-like
+        Relative vorticity adding to the local Coriolis frequency. If "f" is the Coriolis
+        frequency then "f" + omega will be the effective Coriolis frequency as defined by Kunze (1985).
+        Positive values correspond to cyclonic vorticity, irrespectively of the latitudes of the data
+        points.
+
+    Returns
+    -------
+    xy : array-like
```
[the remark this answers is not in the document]
selipot: yep, forgot to change that.

clouddrift/analysis.py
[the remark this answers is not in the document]
selipot: Or we make these parameters optional?

clouddrift/analysis.py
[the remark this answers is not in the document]
milancurcic: In that case, let's just call it `time_step` and document clearly that it must be in units of seconds. Depending on what arithmetic is used inside the function, it's possible that it will support timedelta objects out of the box. We need to test it and find out.

clouddrift/analysis.py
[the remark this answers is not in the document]
milancurcic: I'm in favor of just hardcoding sane defaults in the implementation until we learn more about broader user needs.

selipot:
I think with little modification these new functions could accept unidimensional xarray.Dataarray as input, is it worth it? Not sure it is.

clouddrift/analysis.py
```diff
@@ -1198,3 +1208,239 @@ def unpack_ragged(
...
+        raise ValueError("longitude and latitude arrays must have the same shape.")
+
+    # length of data sequence
+    data_length = longitude.shape[0]
+
+    # initialization: what shall we return?
+    # inertialextract returns lon and lat after subtraction of the inertial oscillations
+    # and the horizontal inertial displacement in kilometers
+    if isinstance(relative_vorticity, float):
+        relative_vorticity = relative_vorticity * np.ones_like(longitude)
```
philippemiron, line 1274:
```suggestion
        relative_vorticity = np.full_like(longitude, relative_vorticity)
```

clouddrift/analysis.py
```diff
@@ -1198,3 +1208,239 @@ def unpack_ragged(
...
+
+    # convert transforms to horizontal displacements on tangent plane
+    wxh, wyh = cartesian_to_tangentplane(wx, wy, wz, longitude_new, latitude_new)
+
+    # rotary wavelet transforms to select inertial component; need to divide by sqrt(2)
+    wp = (wxh + 1j * wyh) / np.sqrt(2)
+    wn = (wxh - 1j * wyh) / np.sqrt(2)
+
+    # find the values of radian_frequency/dt that most closely match cor_freq
+    # is there a better way to do this?
```
philippemiron, line 1351: This looks good.

clouddrift/analysis.py
```diff
@@ -1198,3 +1208,239 @@ def unpack_ragged(
...
+    frequency_bins = [
+        np.argmin(np.abs(cor_freq[i] - radian_frequency / time_step))
+        for i in np.arange(0, data_length)
```
philippemiron, line 1354: I think this can be:

```suggestion
        for i in range(data_length)
```

clouddrift/analysis.py
```diff
@@ -1198,3 +1208,239 @@ def unpack_ragged(
...
+    wp = wp[frequency_bins, list(range(0, data_length))]
+    wn = wn[frequency_bins, list(range(0, data_length))]
+
+    # need an index of northen latitude points and index of
+    # southern latitude points
+    north = latitude >= 0
+
+    # initialize the zonal and meridional components of inertial displacements
+    wxhat = np.zeros_like(latitude).astype(np.complex64)
+    wyhat = np.zeros_like(latitude).astype(np.complex64)
```
philippemiron, line 1370:
```suggestion
    wxhat = np.zeros_like(latitude, dtype=np.complex64)
    wyhat = np.zeros_like(latitude, dtype=np.complex64)
```

clouddrift/analysis.py
```diff
@@ -1198,3 +1208,239 @@ def unpack_ragged(
...
+    Parameters
+    ----------
+    longitude : float or np.ndarray
+        Longitude in degrees.
+    latitude : float or np.ndarray
+        Latitude in degrees.
+    x : float or np.ndarray
+        Zonal displacement in kilometers.
+    y : float or np.ndarray
+        Meridional displacement in kilometers.
```
philippemiron, line 1414: Do we have to put this in km? I would keep everything in SI if possible. 

If we keep km, I would create a variable `METERS_PER_KM` to perform the conversion instead of using 1e-3. Or, just a `EARTH_RADIUS_KM` defined once at the top.

philippemiron (approved):

clouddrift/analysis.py
selipot: I agree with the SI argument but I am curious of the following: why is it that having an obvious variable `METERS_PER_KM` is better than 1e-3?

philippemiron: Because you have to guess what is happening to have to multiply by 1e-3. I had to go to the docstring and saw that it was in km. In general, we should avoid multiplying by any numbers and create either a variable to explain what it does. Exception to very obvious things like `1/2 * (a+b)`).

selipot: That makes sense. However I think we should stick to SI as a principle and return meters ...

PR #301, opened 2023-10-14
inertial extraction, take 2

selipot:


selipot:
Recreated the PR after recloning etc.

clouddrift/kinematics.py
```diff
@@ -3,9 +3,246 @@
...
+    relative_vorticity: Optional, float or array-like
+        Relative vorticity adding to the local Coriolis frequency. If "f" is the Coriolis
+        frequency then "f" + `relative_vorticity` will be the effective Coriolis frequency as defined by Kunze (1985).
+        Positive values correspond to cyclonic vorticity, irrespectively of the latitudes of the data
+        points.
+
+    Returns
+    -------
+    xy : array-like
+        Relative displacement in kilometers from inertial oscillations. The real and
```
philippemiron, line 57: Didn't we agree to switch to meters?

clouddrift/kinematics.py
```diff
@@ -3,9 +3,246 @@
...
+    # length of data sequence
+    data_length = longitude.shape[0]
+
+    if isinstance(relative_vorticity, float):
+        relative_vorticity = np.full_like(longitude, relative_vorticity)
+    elif isinstance(relative_vorticity, np.ndarray):
+        if not relative_vorticity.shape == longitude.shape:
+            raise ValueError(
+                "relative_vorticity must be a float or the same shape as time, lon, and lat."
+            )
```
philippemiron, line 88: Add the Exception section here

clouddrift/kinematics.py
```diff
@@ -3,9 +3,246 @@
...
+    x, y, z = spherical_to_cartesian(longitude_unwrapped, latitude)
+
+    # wavelet transform of x, y, z
+    wavelet, _ = morse_wavelet(data_length, gamma, beta, radian_frequency)
+    wx = wavelet_transform(x, wavelet)
+    wy = wavelet_transform(y, wavelet)
+    wz = wavelet_transform(z, wavelet)
+
+    # x, y, z minus the transforms converted back to lat and lon
+    # dum = np.broadcast(x,wx)
```
philippemiron, line 134: We can remove all those comments I guess?

clouddrift/kinematics.py
```diff
@@ -3,9 +3,246 @@
...
+    frequency_bins = [
+        np.argmin(np.abs(cor_freq[i] - radian_frequency / time_step))
+        for i in range(data_length)
+    ]
+
+    # get the transform at the inertial and "anti-inertial" frequencies
+    # extract the values of wp and wn at the calculated index as a function of time
+    # positive is anticyclonic (inertial) in the southern hemisphere
+    # negative is anticyclonic (inertial) in the northern hemisphere
+    wp = wp[frequency_bins, list(range(0, data_length))]
```
philippemiron, line 168:
```suggestion
    wp = wp[frequency_bins, np.arange(0, data_length)]
```

clouddrift/kinematics.py
```diff
@@ -3,9 +3,246 @@
...
+
+def residual_positions_from_displacements(
+    longitude: Union[float, np.ndarray],
+    latitude: Union[float, np.ndarray],
+    x: Union[float, np.ndarray],
+    y: Union[float, np.ndarray],
+) -> Union[Tuple[float], Tuple[np.ndarray]]:
+    """
+    Return residual longitudes and latitudes along a trajectory on the spherical Earth
+    after correcting for zonal and meridional displacements x and y in kilometers.
```
philippemiron, line 202: Meters?

clouddrift/kinematics.py
selipot: I do not know what you mean here.

philippemiron: In other docstrings, we add a Exception section that states: this function throws a ValueError if ..., etc.

philippemiron (approved):

clouddrift/kinematics.py
```diff
@@ -3,9 +3,256 @@
...
+        if not relative_vorticity.shape == longitude.shape:
+            raise ValueError(
+                "relative_vorticity must be a float or the same shape as longitude and latitude."
+            )
+
+    if not isinstance(relative_bandwidth, float):
+        raise ValueError("relative_bandwidth must be a float.")
+
+    if not isinstance(time_step, float):
+        raise ValueError("time_step must be a float.")
```
philippemiron, line 102: I don't think this is needed. It's multiply or divided by floats before so it will get cast.

clouddrift/kinematics.py
```diff
@@ -3,9 +3,256 @@
...
+
+    # inertial displacement in meters
+    xhat = np.real(wxhat)
+    yhat = np.real(wyhat)
+
+    return xhat, yhat
+
+
+def residual_positions_from_displacements(
+    longitude: Union[float, np.ndarray],
```
philippemiron, line 195: Here you put `float` or `nd.array`, but the first line validates the shape:
```
if not longitude.shape == latitude.shape == x.shape == y.shape:
```
which doesn't exist for a `float`.

clouddrift/kinematics.py
```diff
@@ -3,9 +3,245 @@
...
+        if not relative_vorticity.shape == longitude.shape:
+            raise ValueError(
+                "relative_vorticity must be a float or the same shape as longitude and latitude."
+            )
+
+    if not isinstance(relative_bandwidth, float):
+        raise ValueError("relative_bandwidth must be a float.")
+
+    if not relative_bandwidth <= 1:
+        raise ValueError("relative_bandwidth must be less or equal to one.")
```
milancurcic, line 104: @selipot why check for `float` here if it needs to be <= 1?

Below we use `np.abs(relative_bandwidth)`, so are negative values OK?

If not, I suggest just checking for the value:

```suggestion
    if not 0 <= relative_bandwidth <= 1:
        raise ValueError("relative_bandwidth must be in the [0, 1] range.")
```

clouddrift/kinematics.py
milancurcic: Yes, I removed this check. The tests pass now.

clouddrift/kinematics.py
```diff
@@ -3,9 +3,256 @@
...
+    (0.0, 0.0)
+
+    Raises
+    ------
+    ValueError
+        If longitude, latitude, x, and y do not have the same shape.
+
+    """
+
+    if not longitude.shape == latitude.shape == x.shape == y.shape:
```
philippemiron:
```suggestion
    if (type(longitude) == np.ndarray or type(latitude) == np.ndarray or type(x) == np.ndarray or type(y) == np.ndarray) and not longitude.shape == latitude.shape == x.shape == y.shape:
```

this is ugly, but would work if all inputs are floats. Otherwise, it validates that they all have the same `shape`.

milancurcic: Let's just let NumPy catch this:

```python
In [1]: from clouddrift.kinematics import residual_positions_from_displacements

In [2]: residual_positions_from_displacements(np.arange(10), np.arange(5), 10, 10)
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[2], line 1
----> 1 residual_positions_from_displacements(np.arange(10), np.arange(5), 10, 10)

File ~/Work/clouddrift/clouddrift_selipot/clouddrift/kinematics.py:241, in residual_positions_from_displacements(longitude, latitude, x, y)
    235 longitudehat = (
    236     180 / np.pi * x / (EARTH_RADIUS_METERS * np.cos(np.radians(latitude)))
    237 )
    239 residual_latitude = latitude - latitudehat
    240 residual_longitude = recast_lon360(
--> 241     np.degrees(np.angle(np.exp(1j * np.radians(longitude - longitudehat))))
    242 )
    244 return residual_longitude, residual_latitude

ValueError: operands could not be broadcast together with shapes (10,) (5,) 
```

(the above is from the latest code without explicit raise)

clouddrift/kinematics.py
selipot: Blame GitHub copilot :)

selipot: 0 will not work for a number of reasons.

selipot:
I want to test on the global GDP before merging I think ...

PR #307, opened 2023-10-24
`kinematics.spin` and `kinematics.eddy_kinetic_energy`

milancurcic:
* [x] `eddy_kinetic_energy`:
  - [x] implementation
  - [x] docs
  - [x] tests
* [x] `spin`:
  - [x] implementation
  - [x] docs
  - [x] tests

### DIFF
--- a/clouddrift/kinematics.py
+++ b/clouddrift/kinematics.py
@@ -733,7 +733,7 @@ def spin(
     difference_scheme: Optional[str] = "forward",
     time_axis: Optional[int] = -1,
 ) -> Union[float, np.ndarray]:
-    """Compute spin from velocities and times.
+    """Compute spin continuously from velocities and times.
 
     Spin is defined as (u'dv' - v'du') / (2 dt EKE) where u' and v' are
     eddy-perturbations of the velocity field, EKE is eddy kinetic energy, dt is
--- a/clouddrift/kinematics.py
+++ b/clouddrift/kinematics.py
@@ -756,10 +756,19 @@ def spin(
 ) -> Union[float, np.ndarray]:
     """Compute spin continuously from velocities and times.
 
-    Spin is defined as (u'dv' - v'du') / (2 dt EKE) where u' and v' are
-    eddy-perturbations of the velocity field, EKE is eddy kinetic energy, dt is
-    the time step, and du' and dv' are velocity component increments during dt.
-    The result has a rank reduced by one relative to the input arrays.
+    Spin is traditionally (Sawford, 1999; Veneziani et al., 2005) defined as
+    (<u'dv' - v'du'>) / (2 dt EKE) where u' and v' are eddy-perturbations of the
+    velocity field, EKE is eddy kinetic energy, dt is the time step, and du' and
+    dv' are velocity component increments during dt, and < > denotes time
+    average.
+
+    To allow computing spin based on full velocity fields, this function does
+    not do any demeaning of the velocity fields. If you need the spin based on
+    velocity anomalies, ensure to demean the velocity fields before passing
+    them to this function.
+
+    Averaging is performed along ``time_axis`` so the result has a rank reduced
+    by one relative to the input arrays.
 
     u, v, and time can be multi-dimensional arrays. If the time axis, along
     which the finite differencing is performed, is not the last one (i.e.
@@ -916,6 +925,6 @@ def spin(
         2 * np.mean(kinetic_energy(u, v), axis=-1)
     )
 
-    #TODO we reduced along time_axis but do we need to swap back and how?
+    # TODO we reduced along time_axis but do we need to swap back and how?
 
     return s
--- a/clouddrift/kinematics.py
+++ b/clouddrift/kinematics.py
@@ -916,4 +916,6 @@ def spin(
         2 * np.mean(kinetic_energy(u, v), axis=-1)
     )
 
+    #TODO we reduced along time_axis but do we need to swap back and how?
+
     return s
--- a/clouddrift/kinematics.py
+++ b/clouddrift/kinematics.py
@@ -765,13 +765,15 @@ def spin(
     To allow computing spin based on full velocity fields, this function does
     not do any demeaning of the velocity fields. If you need the spin based on
     velocity anomalies, ensure to demean the velocity fields before passing
-    them to this function.
+    them to this function. This function also returns instantaneous spin values,
+    so the rank of the result is not reduced relative to the input.
 
-    u, v, and time can be multi-dimensional arrays. If the time axis, along
-    which the finite differencing is performed, is not the last one (i.e.
-    u.shape[-1]), use the time_axis optional argument to specify along which the
-    spin should be calculated. u, v, and time must either have the same shape,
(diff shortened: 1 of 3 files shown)


selipot:
The formula reference should be in [Griffa et al. 2008](https://agupubs.onlinelibrary.wiley.com/doi/full/10.1029/2007GL032100)

milancurcic:
@selipot take a look at the current implementation.

clouddrift/kinematics.py
```diff
@@ -19,6 +20,62 @@
...
+    """Compute eddy kinetic energy from zonal and meridional velocities.
+
+    Eddy kinetic energy is defined as one half the variance of the velocity field.
+
+    To obtain the result, variances along ``time_axis`` are computed for each
+    velocity component. The rank of the result is thus reduced by one relative
+    to the input arrays.
+
+    The eddy perturbations are relative to the time-mean velocity, so if you
+    need to apply a shorter time scale, you need to first low-pass filter
```
selipot, line 37: I don't really understand this paragraph. Could you rephrase?

clouddrift/kinematics.py
```diff
@@ -667,3 +724,147 @@ def velocity_from_position(
     dy /= dt
 
     return np.swapaxes(dx, time_axis, -1), np.swapaxes(dy, time_axis, -1)
+
+
+def spin(
```
selipot, line 729: Are we sorting alphabetically functions by convention? Just wondering.

clouddrift/kinematics.py
```diff
@@ -667,3 +724,147 @@ def velocity_from_position(
...
+
+    Forward and backward schemes are effectively the same except that the
+    position at which the velocity is evaluated is shifted one element down in
+    the backward scheme relative to the forward scheme. In the case of a
+    forward or backward difference scheme, the last or first element of the
+    velocity, respectively, is extrapolated from its neighboring point. In the
+    case of a centered difference scheme, the start and end boundary points are
+    evaluated using the forward and backward difference scheme, respectively.
+
+    Reference for the definition of spin is:
```
selipot, line 763: Put instead in a References section as in:

References
----------
.. [1] Veneziani, M., Griffa, A., Reynolds, A.M., Garraffo, Z.D. and Chassignet,
    E.P., 2005. Parameterizations of Lagrangian spin statistics and particle
    dispersion in the presence of coherent vortices. Journal of Marine Research,
    63(6), pp.1057-1083.
.. [2] if there is another one

clouddrift/kinematics.py
```diff
@@ -667,3 +724,147 @@ def velocity_from_position(
...
+
+    Veneziani, M., Griffa, A., Reynolds, A.M., Garraffo, Z.D. and Chassignet,
+    E.P., 2005. Parameterizations of Lagrangian spin statistics and particle
+    dispersion in the presence of coherent vortices. Journal of Marine Research,
+    63(6), pp.1057-1083.
+
+    Parameters
+    ----------
+    u : np.ndarray
+        Zonal velocity
```
selipot, line 773: I always end with a period as in
    Zonal velocities.

Should we have a convention?

clouddrift/kinematics.py
```diff
@@ -667,3 +724,147 @@ def velocity_from_position(
...
+    # np.swapaxes returns a view of the array, not a copy, if the input is a
+    # NumPy array. Otherwise, it returns a copy.
+    u = np.swapaxes(u, time_axis, -1)
+    v = np.swapaxes(v, time_axis, -1)
+    time = np.swapaxes(time, time_axis, -1)
+
+    if not time.shape == u.shape:
+        # time is 1-d array; broadcast to u.shape.
+        time = np.broadcast_to(time, u.shape)
+
```
selipot, line 825: Make this an option with a keyword argument set to False by default such as `anomaly=False` i.e. by default the input velocities are not anomalies. I am thinking of an implementation for which a mean spatial field is removed first from the velocities. In that case the user passes `anomaly = True`.

selipot (changes requested):

clouddrift/kinematics.py
milancurcic: The pedantically correct thing to do (as I learned from my book editor) is to use a period when we have a full sentence (e.g. "This is a zonal velocity.") and to omit it when it's not a sentence ("Zonal velocities").

We don't need to do that because almost nobody will care. We can use periods always if you prefer it.

clouddrift/kinematics.py
milancurcic: Does `anomaly=False` mean that we compute KE and spin based on full velocities (as input)? Likewise, if `anomaly=True`, we compute EKE by demeaning the input velocities first (current implementation)?

Or, do you mean that `anomaly=False` is a way for the user to communicate whether the input velocity is a full velocity or an anomaly and we demean or not based on that information?

selipot: `anomaly=False` means the input velocities are not anomalies and hence we estimate the mean first as the trajectory mean in order to calculate the velocity anomalies and thus the spin and the eke.

milancurcic: OK. And what do we do when `anomaly=True`?

selipot: `anomalies=True` means the input velocity components are already the u' and v' of your formula. 

Which makes me think that the `eddy_kinetic_energy` function should probably be replaced by a `kinetic_energy` function which calculates `np.mean(u**2) + np.mean(v**2)`. If the input is an anomaly it is an EKE, if the input is not an anomaly it is a total kinetic energy.

selipot: That's an option which I support. In this case the code should not remove the Lagrangian mean from the input.

selipot:
Add DOIs to references?

clouddrift/kinematics.py
```diff
@@ -704,3 +745,186 @@ def velocity_from_position(
...
+
+def spin(
+    u: np.ndarray,
+    v: np.ndarray,
+    time: np.ndarray,
+    difference_scheme: Optional[str] = "forward",
+    time_axis: Optional[int] = -1,
+) -> Union[float, np.ndarray]:
+    """Compute spin continuously from velocities and times.
```
milancurcic, line 757: @selipot what do you mean by "continuously" here? I find it more confusing than without this word.

selipot: Continuous because it does not reduce the dimension of the input. Spin studies typically calculate an average of the spin over a segment length. So it is usually a single number for a trajectory or a segment. Could be removed.

milancurcic: In the current implementation we reduce over `time_axis`. Should we not?

milancurcic: On second read of Veneziani et al. (2005), eq. 2, there should be no time averaging. I confused the $< >$ brackets for a time average (it's a trajectory ensemble average). This will further simplify the code.

selipot:
Another implementation @RickLumpkin reminded me of https://agupubs.onlinelibrary.wiley.com/doi/10.1002/2015JC011435

milancurcic:
This is now ready for another review. It bumps the version to 0.26.0.

selipot (approved):

PR #310, opened 2023-10-26
Inertial modifications

### LINKED ISSUE #309: output types not consistent for residual_positions_from_displacements
I have an issue where the outputs of `residual_positions_from_displacements` are not of consistent type.

As an example:

`lon_res, lat_res = residual_positions_from_displacements(ds["lon"],ds["lat"],x,y)`

returns `lon_res` as a numpy.array but `lat_res` as a xarray dataarray. This makes sense if one looks at the code. I'd like to know how you think this should be handled code-wise.  As an example make sure that everything returned is of type numpy array.

selipot:
This PR does three things

1. slightly change the names of the functions to remove plural forms to be consistent
2. fix #309 
3. implement a new feature of `inertial_oscillation_from_position` whereas the user can input the wavelet duration instead of the bandwidth.

### DIFF
--- a/clouddrift/kinematics.py
+++ b/clouddrift/kinematics.py
@@ -72,7 +72,11 @@ def inertial_oscillation_from_position(
     To extract displacements from inertial oscillations from sequences of longitude
     and latitude values, equivalent to bandpass around 20 percent of the local inertial frequency:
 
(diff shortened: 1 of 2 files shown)


philippemiron (approved):

milancurcic (approved):
Good changes, thank you.

Issue #319, opened 2023-11-10
ANDRO: An Argo-based deep displacement dataset adapter


philippemiron:

philippemiron:
or YoMaHa, available here http://apdrc.soest.hawaii.edu/projects/Argo/data/trjctry/0-Near-Real_Time/. Both have a very similar format but use slightly different methodologies to approximate the at-depth locations of Argo floats..

philippemiron:
I will work on this next. I'm thinking of implementing "Andro" and "YoMaHa" in the same adapter as their format is very similar. We can default to Andro which is newer, but let user use YoMaHa if they want to.

PR #321, opened 2023-11-10
subsurface floats dataset

### LINKED ISSUE #320: Subsurface float trajectories dataset adapters

philippemiron:
resolves #320

- [x] quick implementation
- [x] clean up variable names
- [x] create variables: rowsize, ids
- [x] variable attributes (add coordinates)
- [x] docstring
- [x] tests to at least validate that the url is valid (to identify update)
- [x] wrap this in datasets
- [x] fix output to netcdf
- [x] references (?)

Comments?


philippemiron:
@milancurcic, I don't remember exactly how this works, but do we always create a wrapper in the `datasets` module that just calls the `to_xarray()` function?

philippemiron:
Someone asked me for the subsurface floats dataset recently, so I decided that instead of updating my code, I would do it here. I'm thinking of adding Andro (#319) or YoMaHa after this, based on Argo floats.

milancurcic:
Adapter is needed whenever a dataset needs to be re-organized to be a CloudDrift conformable ragged-array dataset (`obs` and `traj` dims, `rowsize` variable; can have different names but they must be defined). For example, adapter is not needed for the upcoming Spotter open dataset because we coordinated with them to ensure it's a CloudDrift conformable ragged array.

Dataset function is merely a convenience accessor to an existing ragged-array dataset, so that the user doesn't need to find and manage data URLs. For example, in case of GDP, we simply `xr.open_dataset(url)` and return `ds`. In case of MOSAiC, the dataset is relatively small so we invoke the adapter under the hood.

So, I expect that every `clouddrift.adapter.*` will have a corresponding `clouddrift.datasets.*`, but not all `clouddrift.datasets.*` need to have an adapter.

From the end-user (as in consumer of data for analysis; not a contributor to the library) point of view, they should only ever have to access `clouddrift.datasets`.

philippemiron:
Thanks, make sense.

clouddrift/adapters/subsurface.py
```diff
@@ -92,7 +92,9 @@ def to_xarray(
             "index_exp": (["traj"], metadata["indexExp"]),
             "float_type": (["traj"], metadata["fltType"]),
             "id": (["traj"], metadata["indexFlt"]),
-            "date": (["obs"], data["dtnum"]),
+            "rowsize": (["traj"], rowsize),
+            "datenum": (["obs"], data["dtnum"]),
+            "ids": (["obs"], np.repeat(metadata["indexFlt"], rowsize)),
```
milancurcic, line 97: I suggest not including `ids` in the output dataset as it's not needed by `cd.ragged.subset` anymore and is readily computable.

philippemiron: 🫡 it's gone.

clouddrift/adapters/subsurface.py
```diff
@@ -0,0 +1,170 @@
+"""
+This module defines functions used to adapt the subsurface float trajectories as
+a ragged-array dataset.
```
selipot, line 3: Tiny bit more info here about the dataset perhaps?

selipot: rename to subsurfacefloat.py?

clouddrift/adapters/subsurface.py
```diff
@@ -0,0 +1,170 @@
...
+import os
+import pandas as pd
+import scipy.io
+import tempfile
+import urllib.request
+import xarray as xr
+import warnings
+
+SUBSURFACE_FLOAT_DATA_URL = (
+    "https://www.aoml.noaa.gov/phod/float_traj/files/allFloats_12122017.mat"
```
selipot, line 24: a .mat file is downloaded? How do you read this in python?

selipot (changes requested):

clouddrift/adapters/subsurface.py
philippemiron: With scipy. It's there a few lines below.

clouddrift/adapters/subsurface.py
milancurcic: Does this dataset have a proper name? It's not clear to me from its website. If not, I suggest `subsurface_floats`.

clouddrift/adapters/subsurface.py
```diff
@@ -0,0 +1,170 @@
...
+SUBSURFACE_FLOAT_VERSION = "December 2017 (version 2)"
+SUBSURFACE_FLOAT_TMP_PATH = os.path.join(
+    tempfile.gettempdir(), "clouddrift", "subsurface_float"
+)
+
+
+def download(file: str):
+    print(
+        f"Downloading Subsurface float trajectories from {SUBSURFACE_FLOAT_DATA_URL} to {file}..."
+    )
```
milancurcic, line 35: I'd put this in the first branch of the `if` block below so that we don't say "Downloading..." if we're gonna skip it.

clouddrift/adapters/subsurface.py
philippemiron: That was also my hesitation. I don't see any real defined name for it..

clouddrift/adapters/subsurface.py
```diff
@@ -0,0 +1,170 @@
...
+        "lat": {"long_name": "Latitude", "units": "degrees_north"},
+        "rowsize": {
+            "long_name": "Number of observations per trajectory",
+            "sample_dimension": "obs",
+            "units": "-",
+        },
+        "u": {"long_name": "Eastward velocity", "units": "m/s"},
+        "v": {"long_name": "Northward velocity", "units": "m/s"},
+        "t": {"long_name": "Temperature", "units": "Celsius"},
+        "p": {"long_name": "Pressure", "units": "Millibar"},
```
milancurcic, line 145: Do we care about CF standard names and units here? E.g. `sea_water_temperature` and `degree_C`. I think lon, lat, u, v are clear but at least "t" and "p" could be expanded to full names.

philippemiron: Do we do this in other datasets? I don't see this in `gdp1h.py/gdp6h.py/mosaic.py` for example. I can rename the variables if this is preferred.

clouddrift/adapters/subsurface.py
```diff
@@ -0,0 +1,170 @@
...
+    # set attributes
+    for var in vars_attrs.keys():
+        if var in ds.keys():
+            ds[var].attrs = vars_attrs[var]
+        else:
+            warnings.warn(f"Variable {var} not found in upstream data; skipping.")
+    ds.attrs = attrs
+
+    # set coordinates
+    ds = ds.set_coords(["time", "id"])
```
milancurcic, line 168: Let's decide on a convention to follow. Do we want "time" and "id" or "obs" and "traj"? Do we want coordinate names to match dimension names? I think even between the datasets I implemented (MOSAiC and GLAD) I haven't been 100% consistent with this. Having a set convention would help. I'll open a separate issue for this.

clouddrift/adapters/subsurface.py
milancurcic: I think it is for MOSAIC:

```python
    ds["longitude"].attrs = {
        "long_name": "longitude",
        "standard_name": "longitude",
        "units": "degrees_east",
    }

    ds["latitude"].attrs = {
        "long_name": "latitude",
        "standard_name": "latitude",
        "units": "degrees_north",
    }
```

but that one is easy because it only has longitude and latitude.

I don't know what's best to do. On one hand I think established conventions are good (our datasets are more likely to load correctly with NetCDF viewers for example; I don't know if that's a realistic use case), and on the other hand sometimes these standard names are unwieldy. I don't know if "sea_water_temperature" is necessary, but at the very least "temperature" (and perhaps "water_temperature") is more clear than "t". Yes, we see if from "long_name", but it helps to be able to see it from the variable name itself. Likewise for "p".

clouddrift/adapters/subsurface.py
philippemiron: so if we follow the CF standard name, are those the variables name or just an extra attributes `"standard_name": sea_water_temperature `

clouddrift/adapters/subsurface.py
philippemiron: I put `temp` and `pres` but added the correct corresponding `standard_name`.. kinda following what I see here, https://cf-xarray.readthedocs.io/en/latest/examples/introduction.html. As long as the variables are _readable_, and we assigned the corresponding `"standard_name"` and `"units"`.

image

philippemiron: I think I fixed the issues with the units too. Let me know if you see anything else!

clouddrift/adapters/subsurface.py
milancurcic: I think you meant to put this comment in the other thread about CF attributes. As I understand it, the variable name can be anything, and attributes help that variable be identified by consumers. So, having variable named `temp` with an attribute `standard_name` of `sea_water_temperature` works.

clouddrift/adapters/subsurface.py
philippemiron: I added _some_ details. Let me know if you think this should be further improved.

selipot:
Looks good but it does not seem like the dataset can be used with `subset` for variables with dimension `exp`. Should not this be possible?

selipot (approved):

milancurcic:
@philippemiron I will test this locally today. Can you also cast all float variables except time, lat, lon to float32?

philippemiron:
Yes. Doesn't this in the glad adapters converts all variables?
```
    # Cast double floats to singles
    for var in ds.variables:
        if ds[var].dtype == "float64":
            ds[var] = ds[var].astype("float32")
```

milancurcic:
Yes. I special-cased GLAD because all longitudes are in the Gulf (rather than global) and the position error included in the dataset exceeds the precision of the single-precision ulp.

Since subsurface floats are global (as I understand it), it's safest to keep lat, lon as float64.

philippemiron:
Ok good! I will add this for the 3 datasets I'm working on.

```
    # Cast double floats to singles
    double_vars = ["lat", "lon"]
    for var in [v for v in ds.variables if v not in double_vars]:
        if ds[var].dtype == "float64":
            ds[var] = ds[var].astype("float32")
```

milancurcic (approved):
This is great; thank you. Will merge.

PR #322, opened 2023-11-14
yomaha and andro datasets

### LINKED ISSUE #319: ANDRO: An Argo-based deep displacement dataset adapter

philippemiron:
Resolves #319 

- [x] working implementation
- [x] look at ANDRO and see if this could be combined here or should just be another datasets
- [x] YoMaHa: include WMO id, DAC, and argo type (3 "traj" variables)
- [x] clean up code
- [x] add docstring and references
- [x] rename variables
- [x] add attributes
- [x] wrap in clouddrift.datasets
- [x] add tests


philippemiron:
Not sure what's the strategy to _test_ those datasets since it involved downloading a ~300-400Mb ASCII file.

milancurcic:
We test for GLAD which involves a ~150MB download and it takes a few seconds in CI. So it may be OK to try doing the same here and see how it goes. I don't know for certain but I expect that frequent file downloads in CI are cached in their local CDN.

selipot (approved):
Looks good to me! We need example notebooks!

PR #324, opened 2023-11-14
Remove redundant coords assignment; test for dims and coords

### LINKED ISSUE #323: GLAD dataset structure?
I do not understand the organization of the GLAD dataset: 3 dimensions `(obs, index, traj)` and 2 coordinates `obs(index)` and `traj(traj)`?  Why isn’t it organized like let’s say the gdp1h() datasets with 2 dimensions `(traj,obs)`  and coordinates `(ids(obs),time(obs))`?

milancurcic:
I don't know why this fixes this `index` dim on some systems, but it seems like it does. I added tests to ensure that only `obs` and `traj` are in dims and coords.

@selipot @philippemiron please check on your ends.

Closes #323.

### DIFF
--- a/clouddrift/adapters/glad.py
+++ b/clouddrift/adapters/glad.py
@@ -63,9 +63,10 @@ def to_xarray() -> xr.Dataset:
     # Make the dataset compatible with clouddrift functions.
     ds = (
         ds.swap_dims({"index": "obs"})
+        .drop_vars(["id", "index"])
         .assign_coords(traj=traj)
         .assign({"rowsize": ("traj", rowsize)})
-        .drop_vars(["id", "index"])
+        .rename_vars({"obs": "time", "traj": "id"})
     )
 
     # Cast double floats to singles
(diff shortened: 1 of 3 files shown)


selipot:
Ok `index` is gone.

But why not having like for the GDP the coordinates `time(obs)` and `ids(traj)` instead of `obs(obs)` and `traj(traj)`? I find this not really comprehensible. The coordinate you have currently called `obs` is effectively a time variable and the `traj` is a unique identification string. 

Screenshot 2023-11-14 at 6 54 42 PM

philippemiron:
Works for me too. Only comment: should we put a message to notify the user when the file already exists in `~/.clouddrift/data/`?

milancurcic:
Yes, let's discuss the naming convention.

My preference may be `time(time)` and `id(id)`. Unclear to me why name coords differently from dims, as they map 1-to-1 in our case.

philippemiron:
And I agree with Shane. I'm not sure I understand the addition of `obs` and `traj` as coordinates either.

milancurcic:
As discussed on the call, we will rename `obs(obs)` and `traj(traj)` coordinates to `time(obs)` and `id(traj)` respectively. Dimension names remain the same.

PR #330, opened 2023-11-21
Pairs module

### LINKED ISSUE #297: Chance pairs?
Could we write a function that would efficiently find chance pairs in a Lagrangian dataset? @dhruvbalwada how would you like such a function to look like in terms of its input/output arguments? 

Input arguments could be:
```
def chance_pairs(lon,lat,time,rowsize,max_spatial_distance,max_time_distance):

```
where `rowsize` is a list of integers specifying the number of data points in each row of the input ragged arrays `lon`, `lat`, `time` and the other arguments specify the maximum spatial and temporal distances defining the chance encounters? What should be the output? Pairs of ids that would need to be passed? double indices of rows and columns in rows?

milancurcic:
This PR introduces the `pairs` submodule.

* [x] Implementation
* [x] Docs
* [x] Tests

Closes #297

### DIFF
--- a/clouddrift/pairs.py
+++ b/clouddrift/pairs.py
@@ -93,6 +93,11 @@ def chance_pair(
     Coordinates:
         time     (obs) datetime64[ns] 2012-07-21T21:30:00.524160 ... 2012-07-22T0...
     Dimensions without coordinates: obs
+
+    Raises
+    ------
+    ValueError
+        If ``time1`` and ``time2`` are not both provided or both omitted.
     """
     if (time1 is None and time2 is not None) or (time1 is not None and time2 is None):
         raise ValueError(
@@ -236,8 +241,14 @@ def chance_pairs_from_ragged(
               175, 178])))]
 
     The result above shows that 2 chance pairs were found.
+
+    Raises
+    ------
+    ValueError
+        If ``rowsize`` has fewer than two elements.
     """
-    # TODO check len(rowsize) >= set_size
+    if len(rowsize) < 2:
+        raise ValueError("rowsize must have at least two elements.")
     pairs = list(itertools.combinations(np.arange(rowsize.size), 2))
     i = ragged.rowsize_to_index(rowsize)
     results = []
--- a/clouddrift/pairs.py
+++ b/clouddrift/pairs.py
@@ -153,21 +153,18 @@ def chance_pair(
     return overlap1, overlap2
 
 
-def chance_pairs(
+def chance_pairs_from_ragged(
     lon: array_like,
     lat: array_like,
     rowsize: array_like,
     space_distance: Optional[float] = 0,
     time: Optional[array_like] = None,
     time_distance: Optional[float] = 0,
-) -> List[
-    Tuple[
-        Tuple[int, int],
-        Tuple[np.ndarray, np.ndarray, np.ndarray, np.ndarray, np.ndarray, np.ndarray],
-    ]
-]:
-    """Return all longitudes, latitudes, and (optionally) times, at which pairs
-    of contiguous trajectories satisfy the chance pair criteria.
+) -> List[Tuple[Tuple[int, int], Tuple[np.ndarray, np.ndarray]]]:
+    """Return all chance pairs of contiguous trajectories in a ragged array,
+    and their collocated points in space and (optionally) time, given input
+    ragged arrays of longitude, latitude, and (optionally) time, and chance
+    pair criteria as maximum allowable distances in space and time.
 
     If ``time`` and ``time_distance`` are omitted, the search will be done
     only on the spatial criteria, and the result will not include the time
@@ -200,69 +197,45 @@ def chance_pairs(
 
     Returns
     -------
-    pairs : List[Tuple[Tuple[int, int], Tuple[np.ndarray, np.ndarray, np.ndarray, np.ndarray, np.ndarray, np.ndarray]]]
-        List of tuples, each tuple being a tuple of pair indices and longitudes,
-        latitudes, and times that satisfy the chance pair criteria, for each
-        trajectory. The arrays are returned in the following order:
-        longitude and latitude for the first trajectory, longitude and latitude
-        for the second trajectory, and then times for the first and second
-        trajectories, if time was provided as the input.
+    pairs : List[Tuple[Tuple[int, int], Tuple[np.ndarray, np.ndarray]]]
+        List of tuples, each tuple containing a Tuple of integer indices that
+        corresponds to the trajectory rows in the ragged array, indicating the
+        pair of trajectories that satisfy the chance pair criteria, and a Tuple
+        of arrays containing the indices of the collocated points for each
+        trajectory in the chance pair.
 
     Examples
     --------
     In the following example, we load GLAD dataset as a ragged array dataset,
-    subset the result to retain the first five trajectories, and finally find all longitudes,
-    latitudes, and times that satisfy the chance pair criteria of 6 km separation distance
-    and no time separation:
+    subset the result to retain the first five trajectories, and finally find
+    all trajectories that satisfy the chance pair criteria of 12 km separation
+    distance and no time separation, as well as the indices of the collocated
+    points for each pair.
 
     >>> from clouddrift.datasets import glad
-    >>> from clouddrift.datasets import glad
-    >>> from clouddrift.pairs import chance_pairs
+    >>> from clouddrift.pairs import chance_pairs_from_ragged
     >>> from clouddrift.ragged import subset
     >>> ds = subset(glad(), {"id": ["CARTHE_001", "CARTHE_002", "CARTHE_003", "CARTHE_004", "CARTHE_005"]}, id_var_name="id")
-    >>> pairs = chance_pairs(
+    >>> pairs = chance_pairs_from_ragged(
         ds["longitude"].values,
         ds["latitude"].values,
         ds["rowsize"].values,
-        space_distance=6000,
-        time=time,
+        space_distance=12000,
+        time=ds["time"].values,
         time_distance=np.timedelta64(0)
     )
     [((0, 1),
-      (array([-87.05255 , -87.04974 , -87.04612 , -87.04155 , -87.03718 ,
-              -87.033165], dtype=float32),
-       array([28.23184 , 28.227087, 28.222498, 28.217508, 28.2118  , 28.205938],
-             dtype=float32),
-       array([-87.111305, -87.10078 , -87.090675, -87.08157 , -87.07363 ,
-              -87.067375], dtype=float32),
-       array([28.217918, 28.20882 , 28.198603, 28.187078, 28.174644, 28.161713],
-             dtype=float32),
-       array(['2012-07-21T21:30:00.524160000', '2012-07-21T22:15:00.532800000',
-              '2012-07-21T23:00:00.541440000', '2012-07-21T23:45:00.550080000',
-              '2012-07-22T00:30:00.558720000', '2012-07-22T01:15:00.567360000'],
-             dtype='datetime64[ns]'),
-       array(['2012-07-21T21:30:00.524160000', '2012-07-21T22:15:00.532800000',
-              '2012-07-21T23:00:00.541440000', '2012-07-21T23:45:00.550080000',
-              '2012-07-22T00:30:00.558720000', '2012-07-22T01:15:00.567360000'],
-             dtype='datetime64[ns]')))]
+      (array([153, 156, 159, 162, 165, 168, 171, 174, 177, 180, 183, 186, 189,
+              192, 195, 198, 201, 204, 207, 210, 213, 216]),
+       array([142, 145, 148, 151, 154, 157, 160, 163, 166, 169, 172, 175, 178,
+              181, 184, 187, 190, 193, 196, 199, 202, 205]))),
+     ((3, 4),
+      (array([141, 144, 147, 150, 153, 156, 159, 162, 165, 168, 171, 174, 177,
+              180, 183]),
+       array([136, 139, 142, 145, 148, 151, 154, 157, 160, 163, 166, 169, 172,
+              175, 178])))]
 
-    The result above shows that 6 chance pairs were found between the first and
-    second trajectories only.
-
-    To query only the trajectory pair indices that satisfy the chance pair
-    criteria, use a list comprehension. The following example returns the
-    trajectory pair indices that satisfy the chance pair criteria of 60 km:
-
-    >>> res = chance_pairs(
-        ds["longitude"].values,
-        ds["latitude"].values,
-        ds["rowsize"].values,
-        space_distance=60000,
-        time=time,
-        time_distance=np.timedelta64(0)
-    )
-    >>> [p[0] for p in res]
-    [(1, 4), (1, 3), (2, 3), (0, 1), (2, 4), (1, 2), (0, 2), (0, 3), (3, 4)]
+    The result above shows that 2 chance pairs were found.
     """
     # TODO check len(rowsize) >= set_size
     pairs = list(itertools.combinations(np.arange(rowsize.size), 2))
--- a/clouddrift/pairs.py
+++ b/clouddrift/pairs.py
@@ -93,49 +93,48 @@ def chance_pair(
     """
     time_present = time1 is not None and time2 is not None
 
-    # If time is provided, subset the trajectories to the overlapping times.
     if time_present:
+        # If time is provided, subset the trajectories to the overlapping times.
         overlap1, overlap2 = pair_time_overlap(time1, time2, time_distance)
-        lon1 = lon1[overlap1]
-        lat1 = lat1[overlap1]
-        time1 = time1[overlap1]
-        lon2 = lon2[overlap2]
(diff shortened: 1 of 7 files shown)


clouddrift/pairs.py
```diff
@@ -0,0 +1,229 @@
+"""
+Functions to analyze pairs of contiguous data segments.
+"""
+from clouddrift import sphere
+import numpy as np
+import pandas as pd
+import xarray as xr
+from typing import Tuple, Union
+
+array_like = Union[list[float], np.ndarray[float], pd.Series, xr.DataArray]
```
selipot, line 10: I understand this being useful as a shortcut but will this be represented in the API/docstring when the user look these up?

clouddrift/pairs.py
```diff
@@ -0,0 +1,229 @@
...
+    Returns
+    -------
+    overlap1 : np.ndarray[bool]
+        Boolean mask for the overlapping times in `lon1` and `lat1`.
+    overlap2 : np.ndarray[bool]
+        Boolean mask for the overlapping times in `lon2` and `lat2`.
+
+    Examples
+    --------
+    >>> lon1 = [0, 0, 1, 1]
```
selipot, line 47: I don't understand what this function does and these examples a puzzling to me.

selipot (commented):
I am looking forward to hear more about the direction for this module.

clouddrift/pairs.py
milancurcic: Not sure yet, will find out.

clouddrift/pairs.py
milancurcic: I see that this docstring is a bad case of copy/paste (needs editing).

In short, given an input pair of lon, lat arrays, return masks that denote which elements of input arrays are inside the overlap/intersection of their bounding boxes.

milancurcic:
Tentative algorithm to determine chance pairs:

1. For each pair of contiguous trajectories with (lon, lat, time);
2. If time tolerance is provided, subset (lon, lat) for each trajectory where their times are overlapping; if subsets are empty, exit;
3. Subset (lon, lat) where their bounding boxes are overlapping; if subsets are empty, exit;
4. Compute spatial distance between all points in the pair;
5. If time tolerance is provided, compute time distance between all points in the pair (may be able to use sorted assumption to optimize)
6. Return masks for each trajectory where they satisfy the chance pair requirements.

milancurcic:
You can now review the `chance_pair` API and implementation. The answer it returns is not correct yet because the indices returned are for the subsetted arrays and not the original ones (and the user needs to know where and when the trajectories were a chance pair). I need to figure out how to back-track the indices to the original arrays.

milancurcic:
There are at least a few optimizations that could be further made to the implementation.

In an example running this with 2 GLAD trajectories (~several 1000 elements each), chance pair indices are found in approx ~5ms on my laptop.

Without any pair group algorithmic optimizations, and assuming sequential (single-core) execution (and recall that this is an embarrassingly parallel problem), I get approx. 7 minutes to get all chance indices for all chance pairs of a GLAD dataset. I have no sense for whether this is small or large, but it seems reasonable given that this kind of search is done once or a few times.

milancurcic:
Chance pair search for trajectories that cross the dateline is not implemented yet.

philippemiron:
It looks fine, but I think we also need a function that would take N trajectories and identify pairs. That function would need the ones implemented here so that could be a future addition too.

philippemiron:
I was thinking of something like this:

```
import numpy as np
from scipy.spatial import KDTree

np.random.seed(42)
x = np.random.random(100)
y = np.random.random(100)

ID = [1,2,3]
rowsize = [20, 30, 50]

ids = np.repeat(ID, rowsize)
traj_idx = np.cumsum(np.insert(np.array(rowsize), 0, 0))
tree = KDTree(np.column_stack((x,y)))

# test for 1 trajectory
i = 0
max_dist = 0.01 # with 0.02, it returns both [2,3] as close

x_i = x[traj_idx[i] : traj_idx[i + 1]]
y_i = y[traj_idx[i] : traj_idx[i + 1]]

close_traj = np.unique(np.concatenate([ids[tree.query_ball_point(p, max_dist)] for p in zip(x_i, y_i)]))
close_traj = close_traj[close_traj != ID[i]]
print(close_traj)
```

milancurcic:
Yes, I think that can work in place of the current dense matrix approach (`pair_space_distance` and `pair_time_distance`). Do you know how would scipy's kdtree work with great circle distances? Maybe the space would need to be cast to Cartesian first.

philippemiron:
I read fast and it doesn't seem to support setting a custom distance/metric.

But I guess if we set the distance in meters, then we could convert it to degrees at the current point before using it in the `tree.query_ball_points` function. But we can also simply pass the distance with the same units as `x` and `y`.

milancurcic:
On my brief playing with SciPy's kdtree on a GLAD-scale dataset, it appeared to be both slow and memory heavy (> 10GB RAM use while searching 1 trajectory on the remaining 296).

I'm very close to a working naive implementation and when that's ready we can compare alternative approaches.

Unfortunately, it seems like the use of `np.datetime64` for time prevents the use of `ProcessPoolExecutor`. Using `ThreadPoolExecutor` for now but a workaround may be to cast to floating-point seconds first, then casting back before returning the result.

milancurcic:
I am now returning actual lon, lat, time values where chance pair conditions are satisfied, rather than indices.

philippemiron:
Maybe one easy thing we could do is split the lat/lon initially before launching all the e.submit and creating batches to decrease the required memory. But like you said, better get something working first!

clouddrift/pairs.py
```diff
@@ -0,0 +1,407 @@
...
+    --------
+    In the following example, we load the first two trajectories from the GLAD
+    dataset and find all longitudes, latitudes, and times that satisfy the
+    chance pair criteria of 6 km separation distance and no time separation:
+
+    >>> from clouddrift.datasets import glad
+    >>> from clouddrift.pairs import chance_pair
+    >>> from clouddrift.ragged import unpack
+    >>> ds = glad()
+    >>> lon1 = unpack(ds["longitude"], ds["rowsize"], rows=0).pop()
```
selipot, line 68: This I never use, it is clever. Could this be wrapped in a function eventually?

clouddrift/pairs.py
```diff
@@ -0,0 +1,407 @@
...
+            lon2[indices2],
+            lat2[indices2],
+            time1[indices1],
+            time2[indices2],
+        )
+    else:
+        return lon1[indices1], lat1[indices1], lon2[indices2], lat2[indices2]
+
+
+def chance_pairs(
```
selipot, line 140: chance_pairs vs chance_pair above. Could this lead to confusion?

clouddrift/pairs.py
milancurcic: I often forget about it. It could be useful as a convenience function to wrap this.

clouddrift/pairs.py
milancurcic: It could. One letter difference is too little. This is just prototype naming. Let's discuss better names.

clouddrift/pairs.py
```diff
@@ -0,0 +1,508 @@
...
+
+    return distance
+
+
+def pair_time_distance(
+    time1: array_like,
+    time2: array_like,
+) -> np.ndarray[float]:
+    """Given two arrays of times (or any other monotonically increasing
+    quantity), return the distance between all pairs of times.
```
selipot, line 431: return the temporal distance or temporal separation?

milancurcic:
With the latest commit I have (what I think is) a working implementation + docstrings. Please see example in `chance_pairs` function to get familiar with what it returns (it's complex). The example is for searching 5 GLAD trajectories. This is a reduced example so that the output in the docstring is manageable. I haven't run it yet on the full set as I need to move to other projects for today. Let's discuss in detail tomorrow on the call.

clouddrift/pairs.py
```diff
@@ -0,0 +1,407 @@
...
+        Tolerance in meters for the spatial overlap. If the overlap is within
+        this tolerance, the pair is considered to be a chance pair. Default is
+        0, or no tolerance.
+    time_tolerance : float, optional
+        Tolerance for the temporal overlap. If the overlap is within this
+        tolerance, and the space_tolerance condition is satisfied, the pair is
+        considered a chance pair.. Default is 0, or no tolerance.
+
+    Returns
+    -------
```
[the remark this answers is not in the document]
milancurcic: Currently the user would need to search back from indices, but we could also return that from the function. This current approach where we return lat, lon, time is more due to implementation convenience, but with some more effort we could return indices.

milancurcic:
For docstring edits/verbiage, may be easiest if you just edit and commit changes directly.

clouddrift/pairs.py
```diff
@@ -0,0 +1,508 @@
...
+
+    >>> from clouddrift.datasets import glad
+    >>> from clouddrift.datasets import glad
+    >>> from clouddrift.pairs import chance_pairs
+    >>> from clouddrift.ragged import subset
+    >>> ds = subset(glad(), {"id": ["CARTHE_001", "CARTHE_002", "CARTHE_003", "CARTHE_004", "CARTHE_005"]}, id_var_name="id")
+    >>> res = chance_pairs(
+        ds["longitude"].values,
+        ds["latitude"].values,
+        ds["rowsize"].values,
```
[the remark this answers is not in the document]
milancurcic: This would require a change in the algorithm, let's bookmark it for a follow-up PR, as this one is already quite large.

selipot:
I used comments because I'm not 100% sure what the functions do

clouddrift/pairs.py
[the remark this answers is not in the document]
philippemiron: Agree with Shane. And it is also annoying to find indices of `float` number.

clouddrift/pairs.py
```diff
@@ -0,0 +1,508 @@
...
+           dtype='datetime64[ns]'),
+     array(['2012-07-21T21:30:00.524160000', '2012-07-21T22:15:00.532800000',
+            '2012-07-21T23:00:00.541440000', '2012-07-21T23:45:00.550080000',
+            '2012-07-22T00:30:00.558720000', '2012-07-22T01:15:00.567360000'],
+           dtype='datetime64[ns]'))
+    """
+    time_present = time1 is not None and time2 is not None
+
+    # If time is provided, subset the trajectories to the overlapping times.
+    if time_present:
```
philippemiron, line 95: I guess this could be:
```suggestion
    if time_tolerance:
```
and you could remove the other one. Maybe [verifying](url) if we want to add a time tolerance, we need to have time1 and time2 parameters defined. As of right now, all of them are Optional so you would have to validate time_tolerance somehow.

clouddrift/pairs.py
```diff
@@ -0,0 +1,508 @@
...
+     array(['2012-07-21T21:30:00.524160000', '2012-07-21T22:15:00.532800000',
+            '2012-07-21T23:00:00.541440000', '2012-07-21T23:45:00.550080000',
+            '2012-07-22T00:30:00.558720000', '2012-07-22T01:15:00.567360000'],
+           dtype='datetime64[ns]'),
+     array(['2012-07-21T21:30:00.524160000', '2012-07-21T22:15:00.532800000',
+            '2012-07-21T23:00:00.541440000', '2012-07-21T23:45:00.550080000',
+            '2012-07-22T00:30:00.558720000', '2012-07-22T01:15:00.567360000'],
+           dtype='datetime64[ns]'))
+    """
+    time_present = time1 is not None and time2 is not None
```
philippemiron, line 92:
```suggestion
    if time_tolerance and (time1 is None or time2 is None):
        raise ValueError("arguments time1 and time2 must be defined to use time_tolerance."
```

clouddrift/pairs.py
```diff
@@ -0,0 +1,508 @@
...
+    lat1 = lat1[bbox_overlap1]
+    lon2 = lon2[bbox_overlap2]
+    lat2 = lat2[bbox_overlap2]
+    time1 = time1[bbox_overlap1] if time_present else None
+    time2 = time2[bbox_overlap2] if time_present else None
+
+    space_distance = pair_space_distance(lon1, lat1, lon2, lat2)
+    chance_mask = space_distance <= space_tolerance
+
+    if time_present:
```
philippemiron, line 120:
```suggestion
    if time_tolerance:
```

clouddrift/pairs.py
```diff
@@ -0,0 +1,508 @@
...
+    bbox_overlap1, bbox_overlap2 = pair_bounding_box_overlap(
+        lon1, lat1, lon2, lat2, space_tolerance_degrees
+    )
+
+    lon1 = lon1[bbox_overlap1]
+    lat1 = lat1[bbox_overlap1]
+    lon2 = lon2[bbox_overlap2]
+    lat2 = lat2[bbox_overlap2]
+    time1 = time1[bbox_overlap1] if time_present else None
+    time2 = time2[bbox_overlap2] if time_present else None
```
philippemiron, line 115:
```suggestion
    time1 = time1[bbox_overlap1] if time_tolerance else None
    time2 = time2[bbox_overlap2] if time_tolerance else None
```

clouddrift/pairs.py
```diff
@@ -0,0 +1,508 @@
...
+    chance_mask = space_distance <= space_tolerance
+
+    if time_present:
+        time_distance = pair_time_distance(time1, time2)
+        chance_mask_time = time_distance <= time_tolerance
+        chance_mask = chance_mask & chance_mask_time
+
+    indices2, indices1 = np.where(chance_mask)
+
+    if time_present:
```
philippemiron, line 127:
```suggestion
    if time_tolerance:
```

clouddrift/pairs.py
```diff
@@ -0,0 +1,508 @@
...
+    >>> pair_bounding_box_overlap(lon1, lat1, lon2, lat2, 0.5)
+    (array([False, False,  True,  True]), array([ True,  True, False, False]))
+    """
+    # First get the bounding box of each trajectory.
+    lon1_min, lon1_max = np.min(lon1), np.max(lon1)
+    lat1_min, lat1_max = np.min(lat1), np.max(lat1)
+    lon2_min, lon2_max = np.min(lon2), np.max(lon2)
+    lat2_min, lat2_max = np.min(lat2), np.max(lat2)
+
+    # TODO handle trajectories that cross the dateline
```
philippemiron, line 339: I think you could do: 

```
lon1 = np.unwrap(lon1, period=360)
lon2 = np.unwrap(lon2, period=360)
```

It does the following:
```
lon = np.array([175, 179, -179, -170])
np.unwrap(lon, period=360)
>> array([175, 179, 181, 190])
```

Then at the end we could use:
```
def wrap(s, period):
     return np.mod(s + period/2, period) - period/2
```
to bring it back to [-180, 180] or modify it slightly to bring it back to [0, 360].
```

**Review by philippemiron — COMMENTED** [review]
I made some comments, but in general it's very good for first iteration. I would have personally do direct "distance" in the degrees by default and have a parameter to supply the distance in meters instead (but agree with you better ship this and then build on top!).

File: clouddrift/pairs.py (hunk 8 above)
Comment (milancurcic) [reply clouddrift/pairs.py:96]: This won't work because `time_distance` of 0 or `np.timedelta64(0)` evaluate as False, and these are valid values under which we want to compute time collocation.

**Comment by milancurcic** [comment]
This now returns indices and seems to work correctly in some simple tests with GLAD.

The fix for crossing antimeridian following @philippemiron's solution remains a TODO.

More tests are needed.

Do we like `collocated` as a function name instead of `chance_pair`? Likewise, we could do `collocated_pairs` instead of `chance_pairs`.

File: clouddrift/pairs.py [hunk 14]
```diff
@@ -0,0 +1,525 @@
...
+    time_distance : float, optional
+        Maximum time distance allowed for the pair to qualify as chance pair.
+        If the separation is within this distance, and the space distance
+        condition is satisfied, the pair is considered a chance pair. Default is
+        0, or no distance, i.e. the times must be exactly the same.
+
+    Returns
+    -------
+    pairs : List[Tuple[Tuple[int, int], Tuple[np.ndarray, np.ndarray, np.ndarray, np.ndarray, np.ndarray, np.ndarray]]]
+        List of tuples, each tuple being a tuple of pair indices and longitudes,
```
selipot, line 202: Do we want to say that the first tuple is the trajectory index pair?

selipot:
> Do we like `collocated` as a function name instead of `chance_pair`? Likewise, we could do `collocated_pairs` instead of `chance_pairs`.

For now I am happy/satisfied with the current names as this may be what is best known ... but we can ponder a bit more.

clouddrift/pairs.py
milancurcic: `array_like` defined like this will not appear in parsed docs. Should we be explicit instead, then?

clouddrift/pairs.py
milancurcic: How about `chance_pairs_ragged`?

selipot: it actually should be `chance_pairs_from_ragged`?

milancurcic:
Thanks for testing and the reviews!

PR #336, opened 2023-12-06
Datasets consistency

### LINKED ISSUE #287: Make time consistent in floating-point seconds since epoch across adapters and datasets
See https://github.com/Cloud-Drift/clouddrift/pull/282#issuecomment-1743233112

We should consistently across datasets return `time` in `xr.Dataset`s as floating point seconds since epoch (rather than decoding to `np.datetime64[ns]`.

Users can still decode to `np.datetime64[ns]` if they want and the functions that accept `time` (and support the same set of arithmetic operators) will continue to work.

### LINKED ISSUE #326: modify the gdp ragged array
For the GDP, let's follow consistently our *new* convention of having the following coordinates for the ragged array files: `id(traj)` and `time(obs)`.

milancurcic:
This PR:

* [x] Ensure all times are decoded by default and let the user choose not to decode (closes #287)
* [x] Removes `ids` and applies `id(traj)` as a coordinate to GDP hourly and 6-hourly (closes #326)
* [x] Changes the default value of `id_var_name` from `"ID"` to `"id"` in `ragged.subset`.
* [ ] ~Updates docs/usage.rst to reflect these changes~.
* [x] Bumps version to 0.29.0.

Adapters are not affected. If time needs to be converted at the adapters level, please use `clouddrift.datasets._to_seconds_since_epoch()`.

### DIFF
--- a/clouddrift/datasets.py
+++ b/clouddrift/datasets.py
@@ -2,22 +2,27 @@
 This module provides functions to easily access ragged array datasets. If the datasets are 
 not accessed via cloud storage platforms or are not found on the local filesystem,
 they will be downloaded from their upstream repositories and stored for later access 
-(~/.clouddrift for unix-based systems).
+(~/.clouddrift for UNIX-based systems).
 """
-
 from clouddrift import adapters
-import numpy as np
 import os
 import xarray as xr
 
 
-def gdp1h() -> xr.Dataset:
+def gdp1h(decode_times: bool = True) -> xr.Dataset:
     """Returns the latest version of the NOAA Global Drifter Program (GDP) hourly
     dataset as a ragged array Xarray dataset.
 
     The data is accessed from zarr archive hosted on a public AWS S3 bucket accessible at
     https://registry.opendata.aws/noaa-oar-hourly-gdp/. Original data source from NOAA NCEI
     is https://doi.org/10.25921/x46c-3620).
+
+    Parameters
+    ----------
+    decode_times : bool, optional
+        If True, decode the time coordinate into a datetime object. If False, the time
+        coordinate will be an int64 or float64 array of increments since the origin
+        time indicated in the units attribute. Default is True.
 
     Returns
     -------
@@ -33,7 +38,7 @@ def gdp1h() -> xr.Dataset:
     Dimensions:                (traj: 19396, obs: 197214787)
     Coordinates:
         id                     (traj) int64 ...
-        time                   (obs) float64 ...
+        time                   (obs) datetime64[ns] ...
     Dimensions without coordinates: traj, obs
     Data variables: (12/60)
         BuoyTypeManufacturer   (traj) |S20 ...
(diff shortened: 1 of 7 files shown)


philippemiron:
What we don't out an option `decode_time` and pass it to array instead?

milancurcic:
Because different datasets encode their time differently. For example, GLAD time is encoded as integer minutes since some time in 2012 (perhaps the start of deployment).

It's possible to do it with `decode_times` as well but it does need more code in a few different places.

After working some time with `np.datetime64`, I also slightly (60/40) prefer it over seconds since epoch. Seconds are more computationally efficient and make arithmetic easier with some other types, but they're not human readable, which for me is the main downside.

milancurcic:
I tend to agree. What do you think about this approach:

* Each dataset function gets a `decode_times=True` keyword parameter; we default to times as `np.datetime64[ns]`.
* `decode_times=False` converts the time to seconds since epoch (behavior introduced in this PR).

I'm only unsure of this keyword parameter name. `decode_times` is nice because it's consistent with Xarray, and we return an Xarray Dataset. However, `decode_times=False` means that the time should be whatever are the original units and reference of the dataset (e.g. minutes since X in GLAD, seconds since epoch in GDP). But I think we should be consistent in what time units we return here across datasets, i.e. always in seconds since epoch. If we do that, then, this parameter shouldn't be called `decode_times`, but perhaps `seconds_since_epoch` with a default value of False.

milancurcic:
I think the questions boil down to:

1. Default time as `np.datetime64` objects or in physical units?
2. If in physical units, should they be consistent across datasets in seconds since epoch, or should they be whatever is the source encoding of the dataset.

The source encoding of the dataset may be somewhat arbitrary and not necessarily about how data was stored in the upstream data, but how it was processed in the adapter. E.g. in case of GLAD, it's entirely due to how the datetime stamps in text files were ingested into Pandas DataFrame and then converted to Xarray Dataset.

@selipot chime in please on the 2 points above.

selipot:
> I think the questions boil down to:
> 
> 1. Default time as `np.datetime64` objects or in physical units?
> 2. If in physical units, should they be consistent across datasets in seconds since epoch, or should they be whatever is the source encoding of the dataset.
> 
> The source encoding of the dataset may be somewhat arbitrary and not necessarily about how data was stored in the upstream data, but how it was processed in the adapter. E.g. in case of GLAD, it's entirely due to how the datetime stamps in text files were ingested into Pandas DataFrame and then converted to Xarray Dataset.
> 
> @selipot chime in please on the 2 points above.

1. I prefer physical units such as seconds. I dislike time objects in python  because I never remember how to use them (but GitHub copilot helps now)
2. Being consistent across all datasets seems cumbersome to us and may turn out impractical. I am thinking of a dataset of tectonic plate drift datasets with timescales of millions of years.

For point 1 above I like the idea of having a `decode_times=True` which would return time objects (if I understand this to be an option for the functions of the `datasets` module)

selipot:
There are conflicts due to the versions. I probably created this, my apologies.

milancurcic:
OK, so it sounds like we'll go with

1. Decode times to `np.datetime64[ns]` by default, and allow the user to choose to not decode.
2. If not decoded, let datasets keep their upstream units as they are written in the file.

I think that's a fine approach to move forward with here.

milancurcic:
The version in the main branch is incorrect. This PR bumps from the current 0.28.0 to 0.29.0.

PR #338, opened 2023-12-08
Subset by rows and pruning examples in docstring

### LINKED ISSUE #196: prune taking an xarray Dataset as an input?
@milancurcic and I discussed how `prune` could perhaps take a Lagrangian dataset as an input and return a dataset. As an example take the whole hourly GDP, drop the undrogued observations and return whatever is left for both the variables of dimensions `traj` and `obs`. Thoughts @philippemiron ?

### LINKED ISSUE #305: subset by row/segment?
Could we add a functionality to `subset` or another function `subset_rows` which would subset by ... rows? One might argue that one could use the `ID` criteria but after segmenting with `segment` as an example and changing the rows this does not work anymore.

It could be
```
ds_subset = subset(ds,rows=[10,35, 45])
```

milancurcic:
Not a feature but examples in docstring.

Closes #196.
Closes #305.

After looking into how to do this more easily (see #305 for the problem), I found that it doesn't warrant new functionality, e.g. `rows` keyword parameter. We merely need to slice the `id` array by the desired rows.

### DIFF
--- a/clouddrift/ragged.py
+++ b/clouddrift/ragged.py
@@ -599,6 +599,12 @@ def subset(
 
     >>> subset(ds, {"sst": (303.15, np.inf)})
 
+    You can use the same approach to return only the trajectories that are
+    shorter than some number of observations (similar to :func:`prune` but for
+    the entire dataset):
+
+    >>> subset(ds, {"rowsize": (0, 1000)})
+
     Retrieve specific drifters from their IDs:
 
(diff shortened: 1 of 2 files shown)


philippemiron:
That's clean! And close 2 issues with a docstring might be a record. haha

philippemiron:
That's nice!

```
subset(ds, {"obs": (0, n-1)})
```
is quick to get `n` obs.

selipot (approved):

PR #340, opened 2023-12-09
fix deprecation warnings in tests

### LINKED ISSUE #263: DeprecationWarning in wavelet.py
```
../home/milan/Work/clouddrift/clouddrift/clouddrift/wavelet.py:673: DeprecationWarning: Conversion of an array with ndim > 0 to a scalar is deprecated, and will error in future. Ensure you extract a single element from your array before performing this operation. (Deprecated NumPy 1.25.)
  radian_frequency = high_ * np.ones(int(m + 1)) / r ** np.arange(0, m + 1)
```

I don't recall getting this during testing of the wavelet PR but I'm getting it now. I may be on a more recent NumPy version here.

philippemiron:
Should fix #263 and I also removed the warnings that shows up if `beta=0` in:
```
with np.errstate(divide="ignore"):  # ignore warning when beta=0
        fm = np.where(
            beta == 0,
            np.log(2) ** (1 / gamma),
            np.exp((1 / gamma) * (np.log(beta) - np.log(gamma))),
        )
```

I couldn't see the deprecation warnings on my side, but I guess we want to have `m` as in integer, and not an array.

I also went ahead and fix some other warnings.

### DIFF
--- a/clouddrift/adapters/gdp.py
+++ b/clouddrift/adapters/gdp.py
@@ -311,7 +311,7 @@ def rowsize(index: int, **kwargs) -> int:
             decode_times=False,
             concat_characters=False,
             decode_coords=False,
-        ).dims["obs"]
+        ).sizes["obs"]
     except Exception as e:
         print(
             f"Error processing {os.path.join(kwargs['tmp_path'], kwargs['filename_pattern'].format(id=index))}"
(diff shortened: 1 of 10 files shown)


philippemiron:
ds.sizes instead of ds.dims (https://github.com/pydata/xarray/issues/921).

philippemiron:
It looks like the other warnings are from `netcdf4` and `numpy`.

selipot (approved):

PR #341, opened 2023-12-10
sofar dataset

philippemiron:
Not sure if this link is being update since I couldn't find anything online linking to it except Milan's notebook and other notebooks on the same repo, https://github.com/sofarocean/spotter-data-archive/blob/main/examples/.

### DIFF
--- a/clouddrift/adapters/__init__.py
+++ b/clouddrift/adapters/__init__.py
@@ -8,8 +8,10 @@ Deployment (GLAD) dataset, and the MOSAiC sea-ice drift dataset. More adapters w
 in the future.
 """
 
+import clouddrift.adapters.andro
 import clouddrift.adapters.gdp1h
 import clouddrift.adapters.gdp6h
 import clouddrift.adapters.glad
 import clouddrift.adapters.mosaic
 import clouddrift.adapters.subsurface_floats
+import clouddrift.adapters.yomaha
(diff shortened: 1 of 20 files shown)


milancurcic:
Thank you for this PR. Does it need to be brought even with `main`? There are past unrelated commits.

For the function name, I suggest `spotters`.

I recommend holding with this PR until we get a thumbs up from upstream.

philippemiron:
I don't understand, it is showing changes from previous things that are already merge. For example, if you look at `adapters/gdp` it says there is a change from `.dims -> `.sizes`, but this is already on the main branch.

milancurcic:
Right, I can't say I understand it either :joy:

selipot:
@isabelhoughton would you like to chime in on this PR? We are ready to add a dataset accessor for the spotter dataset into clouddrift.

isabelhoughton:
Thanks for sharing - looks good to me. Not many sources point to https://github.com/sofarocean/spotter-data-archive/blob/main/examples/ yet as we're in the final steps of setting it all up. Let me know if any testing on my end would be appreciated.

philippemiron:
Quick question, what is the plan with regards to update to the archive dataset? Is it to always keep the same link or create a new one each time?

isabelhoughton:
We will create a new one each time most likely. We have not been able to sort out a DOI yet, but we would like it to be clear which version of the dataset is referred to externally.

philippemiron:
> We will create a new one each time most likely. We have not been able to sort out a DOI yet, but we would like it to be clear which version of the dataset is referred to externally.

Ok. Once this is in place, we could add logic in the code to look for the latest available archive. For now, I just hardcoded the current link.

Issue #347, opened 2023-12-18
Allow `Callable` criterion in `ragged.subset`


vadmbertr:
In my opinion, it could be useful to subset based on the mask return by a function applied over trajectories.

Possible usecases could be:
- filter in one shot trajectories based on more complex geometries than bounding boxes. For example, keep trajectories of drifters released in the Mediterranean Sea, whithout the ones of the Bay of Biscay or the Black Sea, see:
![med_sea](https://github.com/Cloud-Drift/clouddrift/assets/36510417/bce0430a-545f-4a4e-abe4-c54d58588e7c)
- filter trajectories' positions based on their time differences to the first recorded time of the trajectory (for example, keep positions every 24h after the first recorded time).

It would involved some modifications to the functions `subset` and `_var_mask`, but I think they would not break any current behavior.

What are your thoughts on this? I can propose some snippets to the mentioned usecases and possible implementations if you want.

philippemiron:
I guess you could already do this like this with a function `func()` returning a boolean for the observations.

`subset(ds, {"obs", np.arange(ds.sizes["obs"])[func()]})`

this functionality was just recently added https://github.com/Cloud-Drift/clouddrift/pull/338.

vadmbertr:
I'm not sure to understand your example.

Let's say I want to downsample all my trajectories, and retain only observations made at the same time of the day as the one of the first recorded datetime. Since the number of observations to retain will vary across trajectories, I have to first use `apply_ragged`:
```python
import clouddrift as cd

def daily_masking(
        traj_time: xr.DataArray
) -> np.ndarray:
    traj_time = traj_time.astype("float64")
    traj_time /= 1e9  # to seconds
    traj_time -= traj_time[0]  # start from 0
    mask = (traj_time % (24*60*60)) == 0  # get only obs at the same time as the first one
    rowsize = int(mask.sum())  # the number of obs per traj has to be updated
    return mask, rowsize

gdp1h = cd.datasets.gdp1h()
daily_mask, daily_rowsize = cd.ragged.apply_ragged(daily_masking, gdp1h.time, gdp1h.rowsize)
```

Then, I order to subset, either I do it "manually", or I use `subset` with the functionality introduced by https://github.com/Cloud-Drift/clouddrift/pull/338 as you mentioned:
```python
# manually
[...]

PR #350, opened 2023-12-19
Allow `Callable` criterion in `ragged.subset`

vadmbertr:
Addressing https://github.com/Cloud-Drift/clouddrift/issues/347

- update `subset` and `_mask_var` functions
- add docstring
- add one test

### DIFF
--- a/clouddrift/ragged.py
+++ b/clouddrift/ragged.py
@@ -545,22 +545,22 @@ def subset(
     obs_dim_name: str = "obs",
     full_trajectories=False,
 ) -> xr.Dataset:
-    """Subset the dataset as a function of one or many criteria. The criteria are
-    passed as a dictionary, where a variable to subset is assigned to either a
-    range (valuemin, valuemax), a list [value1, value2, valueN], a single value,
-    or a masking function applied to every trajectory using ``apply_ragged``
+    """Subset a ragged array dataset as a function of one or more criteria. 
+    The criteria are passed with a dictionary, where a dictionary key 
+    is a variable to subset and the associated dictionary value is either a range 
+    (valuemin, valuemax), a list [value1, value2, valueN], a single value, or a 
+    masking function applied to every row of the ragged array using ``apply_ragged``.
 
-    This function relies on specific names of the dataset dimensions and the
-    rowsize variables. The default expected values are listed in the Parameters
-    section, however, if your dataset uses different names for these dimensions
-    and variables, you can specify them using the optional arguments.
+    This function needs to know the names of the dimensions of the ragged array dataset 
+    (traj_dim_name and obs_dim_name), and the name of the rowsize variable (rowsize_var_name).
+    Default values are provided for these arguments (see below), but they can be changed if needed.
 
     Parameters
     ----------
     ds : xr.Dataset
-        Lagrangian dataset stored in two-dimensional or ragged array format
+        Dataset stored as ragged arrays
     criteria : dict
-        dictionary containing the variables and the ranges/values/functions to subset
+        dictionary containing the variables (as keys) and the ranges/values/functions (as values) to subset
     id_var_name : str, optional
         Name of the variable containing the ID of the trajectories (default is "id")
     rowsize_var_name : str, optional
@@ -570,7 +570,7 @@ def subset(
     obs_dim_name : str, optional
         Name of the observation dimension (default is "obs")
     full_trajectories : bool, optional
-        If True, it returns the complete trajectories where at least one observation
+        If True, it returns the complete trajectories (rows) where at least one observation
         matches the criteria, rather than just the segments where the criteria are satisfied.
         Default is False.
 
@@ -582,7 +582,8 @@ def subset(
     Examples
     --------
     Criteria are combined on any data or metadata variables part of the Dataset.
-    The following examples are based on the GDP dataset.
+    The following examples are based on NOAA GDP datasets which can be accessed with the
+    clouddrift.datasets module.
 
     Retrieve a region, like the Gulf of Mexico, using ranges of latitude and longitude:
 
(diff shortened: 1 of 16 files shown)


tests/ragged_tests.py
```diff
@@ -699,6 +699,15 @@ def test_subset_by_rows(self):
         self.assertTrue(all(ds_sub["id"] == [1, 2]))
         self.assertTrue(all(ds_sub["rowsize"] == [5, 4]))
 
+    def test_subset_callable(self):
+        func = (
+            lambda arr: ((arr - arr[0]) % 2) == 0
+        )  # test keeping obs every two time intervals
+        ds_sub = subset(self.ds, {"time": func})
```
philippemiron, line 706: I'm not sure how the test is passing, you are not passing a `dim_name` as I see `ds` has dimensions of `obs` and `traj`... Am I missing something?

Would it be possible to set the default as one of those two instead of xarray default `dim_0` when dimension names are not set.

vadmbertr: `dim_name` is passed to `_mask_var` within `subset`, so we don't need to specify it here.

For your second point, I had more or less the same question. I made `row_size` and `dim_name` arguments as optional because they are not always useful. But at the same time, `_mask_var` is for now always used with the four arguments (except in the docstring examples). Maybe default values could be removed.

philippemiron: Ok I see now. Since the docstring are written like let's keep them optional. And it is not supposed to be called by the users directly anyway.

clouddrift/ragged.py
```diff
@@ -794,6 +831,12 @@ def _mask_var(
     elif isinstance(criterion, (list, np.ndarray, xr.DataArray)):
         # select multiple values
         mask = np.isin(var, criterion)
+    elif callable(
+        criterion
+    ):  # mask directly created by applying `criterion` to each trajectory
+        mask = xr.DataArray(
+            data=apply_ragged(criterion, var, rowsize), dims=[dim_name]
```
philippemiron, line 838: Last comment, this will eventually crash if you pass a function that does not returns one boolean per element. Apply ragged would still work, but that would cause an issue with `np.logical_and` up there...

I can't think of a good way to catch this beforehand, but maybe you do?

vadmbertr: You're right. Actually, maybe in this case it crashes right away because of `dims` being specified?
One way to crash nicely could be to test the size of `apply_ragged` returned value. But I don't think it is possible to make sure beforehand that `apply_ragged` result will be compatible.

philippemiron: I guess that would work, I don't see how we can do it without executing `apply_ragged`. I think that would be good to throw an exception if the returned `mask` doesn't have the correct dimensions. But thinking about this I tried another example and modify the test.

```
func = lambda arr: arr <= 2  # keep id smaller or equal to 2
ds_sub = subset(self.ds, {"id": func})
self.assertTrue(all(ds_sub["id"] == [1, 2]))
self.assertTrue(all(ds_sub["rowsize"] == [5, 4]))
```

and it doesn't work because we are always passing `rowsize` to ragged array. It has to work for both `obs` and `traj` variables.

philippemiron: See, the latest commit. It might be possible to improve but now it works for all variables.

vadmbertr: I was focused on the `obs` case indeed... Your latest commit seems fine to me!

philippemiron (approved):

philippemiron (approved):

selipot (approved):

clouddrift/ragged.py
```diff
@@ -629,6 +630,19 @@ def subset(
 
     >>> subset(ds, {"lat": (21, 31), "lon": (-98, -78), "drogue_status": True, "sst": (303.15, np.inf), "time": (np.datetime64("2000-01-01"), np.datetime64("2020-01-31"))})
 
+    Retrieve observations every 24 hours after the first one, trajectory-wise:
+
+    >>> def daily_masking(
```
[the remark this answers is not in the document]
selipot: Ok, I changed the docstring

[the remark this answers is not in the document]
vadmbertr: Sorry @selipot, I just saw your question. Agree it is simpler to understand!

clouddrift/ragged.py
vadmbertr: I'm a bit late to the party, I was travelling yesterday and had a really quick look at the commit.
Maybe I can propose a small re-write (for consistency):
```python
        if len(var) == len(rowsize):
            mask = criterion(var)
        else:
            mask = apply_ragged(criterion, var, rowsize)

        mask = xr.DataArray(data=mask, dims=[dim_name]).astype(bool)
```

clouddrift/ragged.py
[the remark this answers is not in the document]
vadmbertr: However, I am not sure about the description of the example. It rather retrieve `every two observations of each trajectory` than `every other observation of each trajectory`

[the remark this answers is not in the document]
selipot: if 
```
func = (lambda arr: ((arr - arr[0]) % 2) == 0)
x = np.arange(0,10)
```
then
```
x[func(x)]
array([0, 2, 4, 6, 8])
```
which is *every other observation*.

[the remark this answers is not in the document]
vadmbertr: Sorry, I did not get it in the sens of one observation out of two. Never mind then!

selipot:
Closes #347

vadmbertr:
Hi!
Do you know when the PR will get merged? I have extended the use of a Callable criterion to accept a function taking several variables as arguments. I wanted to proposed the idea and the implementation, but maybe it is better to wait for this one to be merged and then create another PR?

selipot:
Hi, thank you for prodding us. We should finalize and merge this week. We just hired @kevinsantana11 to work with us and we need to clear #354 first.

I agree that another PR should be created for the new features you propose.

vadmbertr:
Great!

kevinsantana11:
Hi @vadmbertr could you rebase your branch to include the recent changes pushed to the main branch? This should retrigger the approval workflow

philippemiron:
@kevinsantana11 no need to create another PR. You can checkout that branch and rebase it.

image

kevinsantana11:
@philippemiron i did give that a try but when I tried pushing the change up I got permission denied. 

Vadim will have to configure that on his end so others can push to his branch.

philippemiron:
image

I just pushed the latest commit to the branch.

vadmbertr:
Just caching up... Thanks for merging and closing this PR!

PR #352, opened 2023-12-20
Update gdp datasets

philippemiron:

PR #354, opened 2024-01-11
Update gdp datasets (fix)

kevinsantana11:
This change fixes the issue that was causing the following [PR](https://github.com/Cloud-Drift/clouddrift/pull/352) to fail the CI builds.

The issue arises from the ERDDAP servers supporting a feature called ``    **OPeNDAP Selection contraints** which tabledap datasets use for requesting subsets of the dataset as well as the server utilizes a sequence data structure in the response. Both of these features are typically not supported in OPeNDAP libraries such as NETCDF-C which is the library that is utilized by xarray through netcdf4-python (python wrapper for the netcdf-c lib) to open netcdf files.

Below more information can be found on this issue as well as where I found the suggestion to first download the file and then open it with xarray.

reference to info on issues with opening URLs with NetCDF-C: https://erddap.aoml.noaa.gov/gdp/erddap/tabledap/documentation.html#netcdfjava

````
[OPeNDAP Libraries](https://erddap.aoml.noaa.gov/gdp/erddap/tabledap/documentation.html#opendapLibraries) - Although ERDDAP is an [OPeNDAP](https://www.opendap.org/) (external link)-compatible data server, you can't use most OPeNDAP client libraries, including [NetCDF-Java, NetCDF-C, NetCDF-Fortran, NetCDF-Perl](https://www.unidata.ucar.edu/software/netcdf/) (external link), or [Ferret](https://ferret.pmel.noaa.gov/Ferret/) (external link), to get data directly from an ERDDAP tabledap dataset because those libraries don't support the OPeNDAP Selection constraints that tabledap datasets use for requesting subsets of the dataset, nor do they support the sequence data structure in the response. (But see this [other approach](https://erddap.aoml.noaa.gov/gdp/erddap/tabledap/documentation.html#netcdfjava) that works with NetCDF libraries.) But you can use the [Pydap Client](https://erddap.aoml.noaa.gov/gdp/erddap/tabledap/documentation.html#PydapClient) or [Java-DAP2](https://www.opendap.org/deprecated-software/java-dap) (external link), because they both support Selection constraints. With both the Pydap Client and Java-DAP2, when creating the initial connection to an ERDDAP table dataset, use the tabledap dataset's base URL, e.g.,
https://coastwatch.pfeg.noaa.gov/erddap/tabledap/pmelTaoDySst

    Don't include a file extension (e.g., .nc) at the end of the dataset's name.
    Don't specify a variable or a Selection-constraint.
...
`````

reference to info on the workaround: https://erddap.aoml.noaa.gov/gdp/erddap/tabledap/documentation.html#netcdfjava

```
[NetCDF-Java, NetCDF-C, NetCDF-Fortran, and NetCDF-Perl](https://www.unidata.ucar.edu/software/netcdf/) (external link) [-](https://erddap.aoml.noaa.gov/gdp/erddap/tabledap/documentation.html#netcdfjava) Don't try to access an ERDDAP tabledap dataset URL directly with a NetCDF library or tool (by treating the tabledap dataset as an OPeNDAP dataset or by creating a URL with the .nc file extension). It won't work.

    These libraries and tools just work with OPeNDAP gridded data. Tabledap offers data as OPeNDAP sequence data, which is a different data structure.
    These libraries and tools just work with actual, static, .nc files. But an ERDDAP URL with the .nc file extension is a virtual file. It doesn't actually exist. 

Fortunately, there is a two step process that does work:

    Use the ERDDAP web pages or a program like [curl](https://erddap.aoml.noaa.gov/gdp/erddap/tabledap/documentation.html#curl) to download a .nc file with a subset of the dataset.
    Open that local .nc file (which has the data in a grid format) with one of the NetCDF libraries or tools. With NetCDF-Java for example, use:
...
```


selipot (approved):

milancurcic:
Thanks, Kevin. The explanation seems reasonable. 

The Xarray depending on Python netCDF4 depending on custom- or system-provided NetCDF-C build, especially when combined with different GitHub CI image configurations, is a moving target. It's OK to spend some effort to try and get it to work but I wouldn't lose hair over this. My suggestion to solve this is to lazy load as we have been on Linux and macOS, but use a workaround (e.g. a whole-file download) on Windows, for the time being. Eventually the platforms will catch up with this and the problem will go away.

Another alternative is to generate the 6-hourly GDP as a Zarr instead of NetCDF, which completely eliminates this issue. But this would require an extra step of updating the dataset upstream, changing the URL, and coordinating with the GDP DAC people.

kevinsantana11:
@milancurcic That makes sense, I think I'll update the change so that it follows the guidance you've provided.

We could save the effort of looking into the ERDDAP servers for later down the line. As I do believe its worth investigating and potentially leveraging the servers.

selipot:
> @milancurcic That makes sense, I think I'll update the change so that it follows the guidance you've provided.
> 
> We could save the effort of looking into the ERDDAP servers for later down the line. As I do believe its worth investigating and potentially leveraging the servers.

Thanks @milancurcic I think your suggestions are the way to go. @kevinsantana11 I do not think the current ERDDAP server provide the data in the clouddrift formats we are advocating, hence I do not think that it is worth investigating at this time.

PR #361, opened 2024-01-22
Subset callable using multiple variables

vadmbertr:
Hi!

Following PR [350](https://github.com/Cloud-Drift/clouddrift/pull/350), the function `subset` could accept a `tuple` of variables as keys for `Callable` values of the `criteria` parameter:

```python
import clouddrift as cd
import numpy as np
import xarray as xr

def mediterranean_masking(lon: xr.DataArray, lat: xr.DataArray) -> xr.DataArray:
    # Mediterranean Sea bounding box
    in_med = np.logical_and(-6.0327 <= lon, np.logical_and(lon <= 36.2173,
                                                           np.logical_and(30.2639 <= lat, lat <= 45.7833)))
    # exclude Bay of Biscay
    in_biscay = np.logical_and(lon <= -0.1462, lat >= 43.2744)
    # exclude Black Sea
    in_blacksea = np.logical_and(lon >= 27.4437, lat >= 40.9088)
    return np.logical_and(in_med, np.logical_not(np.logical_or(in_biscay, in_blacksea)))
subset(ds, {("start_lon", "start_lat"): mediterranean_masking})

gdp1h = cd.datasets.gdp1h()
gdp1h_med = cd.ragged.subset(gdp1h, {("start_lon", "start_lat"): mediterranean_masking})
```

What do you think?

Regards,
Vadim

### DIFF
--- a/clouddrift/ragged.py
+++ b/clouddrift/ragged.py
@@ -545,44 +545,45 @@ def subset(
     obs_dim_name: str = "obs",
     full_trajectories=False,
 ) -> xr.Dataset:
-    """Subset a ragged array dataset as a function of one or more criteria.
+    """Subset a ragged array xarray dataset as a function of one or more criteria.
     The criteria are passed with a dictionary, where a dictionary key
     is a variable to subset and the associated dictionary value is either a range
     (valuemin, valuemax), a list [value1, value2, valueN], a single value, or a
-    masking function applied to every row of the ragged array using ``apply_ragged``.
+    masking function applied to any variable of the dataset.
 
     This function needs to know the names of the dimensions of the ragged array dataset
     (`traj_dim_name` and `obs_dim_name`), and the name of the rowsize variable (`rowsize_var_name`).
-    Default values are provided for these arguments (see below), but they can be changed if needed.
+    Default values corresponds to the clouddrift convention ("traj", "obs", and "rowsize") but should
+    be changed as needed.
 
     Parameters
     ----------
     ds : xr.Dataset
-        Dataset stored as ragged arrays
+        Xarray dataset composed of ragged arrays.
     criteria : dict
-        dictionary containing the variables (as keys) and the ranges/values/functions (as values) to subset
+        Dictionary containing the variables (as keys) and the ranges/values/functions (as values) to subset.
     id_var_name : str, optional
-        Name of the variable containing the ID of the trajectories (default is "id")
+        Name of the variable containing the ID of the trajectories (default is "id").
     rowsize_var_name : str, optional
-        Name of the variable containing the number of observations per trajectory (default is "rowsize")
+        Name of the variable containing the number of observations per trajectory (default is "rowsize").
     traj_dim_name : str, optional
-        Name of the trajectory dimension (default is "traj")
+        Name of the trajectory dimension (default is "traj").
     obs_dim_name : str, optional
-        Name of the observation dimension (default is "obs")
+        Name of the observation dimension (default is "obs").
     full_trajectories : bool, optional
-        If True, it returns the complete trajectories (rows) where at least one observation
-        matches the criteria, rather than just the segments where the criteria are satisfied.
-        Default is False.
+        If True, the function returns complete rows (trajectories) for which the criteria
+        are matched at least once. Default is False which means that only segments matching the criteria
+        are returned when filtering along the observation dimension.
 
     Returns
     -------
     xr.Dataset
-        subset Dataset matching the criterion(a)
+        Subset xarray dataset matching the criterion(a).
 
     Examples
     --------
-    Criteria are combined on any data or metadata variables part of the Dataset.
-    The following examples are based on NOAA GDP datasets which can be accessed with the
+    Criteria are combined on any data (with dimension "obs") or metadata (with dimension "traj") variables
+    part of the Dataset. The following examples are based on NOAA GDP datasets which can be accessed with the
     ``clouddrift.datasets`` module.
 
     Retrieve a region, like the Gulf of Mexico, using ranges of latitude and longitude:
@@ -607,7 +608,7 @@ def subset(
 
     >>> subset(ds, {"rowsize": (0, 1000)})
 
-    Retrieve specific drifters from their IDs:
+    Retrieve specific drifters using their IDs:
 
     >>> subset(ds, {"id": [2578, 2582, 2583]})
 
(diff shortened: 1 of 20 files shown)


clouddrift/ragged.py
```diff
@@ -650,19 +665,26 @@ def subset(
     )
 
     for key in criteria.keys():
-        if key in ds or key in ds.dims:
-            if ds[key].dims == (traj_dim_name,):
+        if np.any(np.isin(key, ds.variables) | np.isin(key, ds.dims)):
+            if isinstance(key, tuple):
+                criterion = [ds[k] for k in key]
+                criterion_dims = criterion[0].dims
```
philippemiron, line 671: just one think here. I will try to find the time to test later, but since you do `criterion[0].dims`, is it assuming that all elements of the tuple have the same dimension? 

For example, using `datasets.gdp1h` would it work doing with a `Callable` like `def filter(start_lon, ve)` or `def filter(traj, id)`, or `def filter(traj, id, ve)` ?

tests/ragged_tests.py
```diff
@@ -714,13 +714,27 @@ def test_subset_callable(self):
         self.assertTrue(all(ds_sub["id"] == [1, 2]))
         self.assertTrue(all(ds_sub["rowsize"] == [5, 4]))
 
+        func = lambda arr1, arr2: np.logical_and(
```
philippemiron, line 717: Could you create another test function with the following.

clouddrift/ragged.py
vadmbertr: Yes that's the assumption. So indeed your examples would not work. If you agree with the idea, the documentation should probably be clear on this.

tests/ragged_tests.py
vadmbertr: You mean having the test case as a separate method? Sure!

vadmbertr (commented):
Concerning your changes, I think the condition should be a `np.all` and not a `np.any`: it is testing if **all** the variables in the `key` are either in `ds.variables` or `ds.dims`. Using `np.any` the test becomes if **any** of the variables in the `key`, which is not what we want I think.

As a quick example:
```
import numpy as np

variables = ("lon", "lat, "ve")
dims = ("traj", "id")
key = ("latitude", "lon")  # wrong key

np.all(np.isin(key, variables) | np.isin(key, dims))  # returns False, expected
np.any(np.isin(key, variables) | np.isin(key, dims))  # returns True, not expected
```

Maybe a test for this should be added.

philippemiron:
> Concerning your changes, I think the condition should be a `np.all` and not a `np.any`: it is testing if **all** the variables in the `key` are either in `ds.variables` or `ds.dims`. Using `np.any` the test becomes if **any** of the variables in the `key`, which is not what we want I think.
> 
> 
> 
> As a quick example:
> 
> ```
> 
> import numpy as np
> 
> 
> 
> variables = ("lon", "lat, "ve")
> 
> dims = ("traj", "id")
> 
> key = ("latitude", "lon")  # wrong key
> 
> 
> 
> np.all(np.isin(key, variables) | np.isin(key, dims))  # returns False, expected
> 
> np.any(np.isin(key, variables) | np.isin(key, dims))  # returns True, not expected
> 
> ```
> 
> 
> 
> Maybe a test for this should be added.

Yes, I didn't want to change the initial `np.all`. My bad.

clouddrift/ragged.py
philippemiron: I can't really think why/how one would use variable with size obs and traj in a single callable.. and we do need to know the length of the result to assign it to the proper dimension.

I think it could be better to validate that all variables of the tuple share the same dimension, and we can return an Exception if it's not case.

tests/ragged_tests.py
philippemiron: Yes, to test one specific aspect per method.

clouddrift/ragged.py
vadmbertr: Yes, it seems the way to go!

tests/ragged_tests.py
kevinsantana11: We should  consider testing the following scenarios:

* a dimension is passed in the tuple that doesn't exist (separate test) in the dataset. I believe if we try to do this we will probably get a KeyError or xarrays Dataset equivalent.
* a user passes in a callable that takes in less parameters than the size of the tuple. When unpacking the iterable it will pass more arguments than the function definition expects.
* type errors in the case where the function parameters expected type is different from its actual type.

kevinsantana11 (changes requested):

kevinsantana11 (approved):
Approving as after talking with Philippe it seems the issues of tracking key errors and the likes can be delegated to xarray.

Currently when a key is utilized in a  Dataset that doesn't contain the variable like ds['test'] the error message we get from xarray is: `KeyError: "No variable named 'test'. Variables on the dataset include ['CompositeReflectivity', 'time', 'lat', 'lon', 'pcptype_nospeckle', 'MaskedCompositeReflectivity', 'Latitude_Longitude']"`

vadmbertr:
@kevinsantana11, that's fine for me. I just add one test related to [this comment](https://github.com/Cloud-Drift/clouddrift/pull/361#issuecomment-1905546299).
I also add a small remark and a raised Exception to the docstring.

clouddrift/ragged.py
```diff
@@ -637,10 +637,32 @@ def subset(
     >>> func = (lambda arr: ((arr - arr[0]) % 2) == 0)
     >>> subset(ds, {"time": func})
 
+    The filtering function can accept several input variables
+    (note that here, filtering is done along the trajectory dimension).
```
selipot, line 641: What do you mean by "here"? In the example that follows because it pertains to the `start_lon` and `start_lat` variables which have dimensions `trajectory`? In that case that comment should probably be moved a few lines but I think it could be removed altogether.

selipot: Perhaps line 634 above could start with: "You can also filter the data by applying a function acting on any of the variables."

clouddrift/ragged.py
```diff
@@ -637,10 +637,32 @@ def subset(
...
+    >>>     in_blacksea = np.logical_and(lon >= 27.4437, lat >= 40.9088)
+    >>>     return np.logical_and(in_med, np.logical_not(np.logical_or(in_biscay, in_blacksea)))
+    >>> subset(ds, {("start_lon", "start_lat"): mediterranean_mask})
+
     Raises
     ------
     ValueError
         If one of the variable in a criterion is not found in the Dataset
+    TypeError
+        If one of the `criteria` key is a tuple while its associated value is not a `Callable` criterion
```
philippemiron, line 661:
```suggestion
    TypeError
        If one of the `criteria` key is a tuple while its associated value is not a `Callable` criterion
    TypeError
        If variables passed to the `Callable` do not share the same dimension
```

clouddrift/ragged.py
vadmbertr: > What do you mean by "here"? In the example that follows because it pertains to the `start_lon` and `start_lat` variables which have dimensions `trajectory`? In that case that comment should probably be moved a few lines but I think it could be removed altogether.

Yes exactly.

By reading the docstring, I felt that it could be useful to highlight that the `Callable` criterion either apply along the `obs` dimension (through `apply_ragged`) or along the `trajectory` dimension by directly invoking the `Callable`.

But it is definetly up to you!

philippemiron:
I think it's good to go.

philippemiron (approved):

kevinsantana11 (approved):

PR #364, opened 2024-01-26
subset docstring edit

selipot:
I think it is necessary to clarify in the docstring that the id_var_name variable is expected to have the dimension "traj" or `traj_dim_name`. We have changed several times the conventions for the dimensions of the id and traj variables.

### DIFF
--- a/clouddrift/adapters/gdp1h.py
+++ b/clouddrift/adapters/gdp1h.py
@@ -601,11 +601,11 @@ def to_raggedarray(
     )
 
     # set dynamic global attributes
-    ra.attrs_global[
-        "time_coverage_start"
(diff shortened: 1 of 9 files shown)


selipot:
This turns out necessary for me because I was reading a local version of the version 2.01 not doing what  the `datasets.gdp1h()` function does:
```
url = "https://noaa-oar-hourly-gdp-pds.s3.amazonaws.com/latest/gdp-v2.01.zarr"
ds = xr.open_dataset(url, engine="zarr", decode_times=decode_times)
ds = ds.rename_vars({"ID": "id"}).assign_coords({"id": ds.ID}).drop_vars(["ids"])
```

milancurcic (approved):

philippemiron (approved):

PR #365, opened 2024-01-28
ruff linting and formatting

kevinsantana11:
* fix linting errors (pyflakes) - Caught a couple small bugs here and there, unused variables and the likes
* sort imports (isort) - consistently sorted imports. Stdlib imports first then external dependencies then our package imports.
* format code against (pycodestyles) - Fairly similar to the formatting with black but gives us control over styling while using black we do rely a bit on the style black prefers.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -14,5 +14,5 @@ jobs:
   ruff:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - uses: chartboost/ruff-action@v1
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -7,10 +7,12 @@ on:
     branches: [ main ]
   pull_request:
     branches: [ main ]
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
 
 jobs:
-  lint:
+  ruff:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - uses: psf/black@stable
+      - uses: chartboost/ruff-action@v1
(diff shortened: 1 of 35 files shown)


.github/workflows/lint.yml
```diff
@@ -7,10 +7,12 @@ on:
...
+  workflow_dispatch:
 
 jobs:
-  lint:
+  ruff:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - uses: psf/black@stable
+      - uses: chartboost/ruff-action@v1
```
philippemiron, line 18: is there a reason why use this one instead of what seems to be the official action https://github.com/astral-sh/ruff ?

philippemiron:
and how do we have to do after committting to the repo?

- `ruff clouddrift/*py` ?

One comment on why I love black, it has almost _no_ control on the styling, so it removes all the discussion on the length of a line or whatever else, it just simply formats.

.github/workflows/lint.yml
kevinsantana11: This comes from their documentation see [this](https://github.com/astral-sh/ruff#usage)

kevinsantana11:
@philippemiron I'm a big fan of black as well, prior to this project I would utilize it on all of my python projects. It seems ruff's styling philosophy is to try and stay compatible with Blacks' styling guide while deviating a small bit between styling updates from black.

To check for formatting/linting issues we would just run `ruff clouddrift tests`. To do the auto-formatting it would be `ruff format clouddrift tests`

selipot:
So how does this work? install ruff with conda or mamba then use the commands above listed by @kevinsantana11 ?

philippemiron:
I would suggest installing it with `brew` so you don't have to install it on each environment.

PR #373, opened 2024-02-20
fix examples

philippemiron:


philippemiron:
Forgot to adjust the test. I will do this later this evening.

philippemiron:
Should be good to go now.

kevinsantana11 (approved):
changes look good to me @philippemiron

PR #375, opened 2024-02-22
🔎 Development docs revision

kevinsantana11:
Changes here include a Makefile which can be invoked to simplify certain common dev tasks (formatting, linting, static analysis tools) so the commands themselves don't need to be remembered just the make targets.

The CLI file can be invoked via `python -m clouddrift.cli --help`. Currently it has one command `gdp1h` which can generate the aggregated file given certain parameters configured through the cli inputs.


selipot:
Update the README.md to reflect these new tools.

Makefile
```diff
@@ -0,0 +1,19 @@
+all: clean check test
+check: lint mypy
+
+lint:
+	ruff clouddrift tests
```
philippemiron, line 5: I'm getting this when running ruff directly.

```
$ ruff sync/
warning: `ruff <path>` is deprecated. Use `ruff check <path>` instead.
```

philippemiron, line 5:
```suggestion
	ruff check clouddrift tests	
```

kevinsantana11:
After looking into how AwkwardArray handles some of their own pre-commit tasks it looks like they leverage pre-commit: https://pre-commit.com/. I want to look further into this tool as I think it might work better than using a Makefile as the automation can be used by both engineers and our CI from the looks of it.

philippemiron:
I think this is a good read on precommit hooks.

https://dev.to/afl_ext/are-pre-commit-git-hooks-a-good-idea-i-dont-think-so-38j6

kevinsantana11:
@philippemiron  the main benefit I saw in it was the automation of all the checks performed in our CI but locally. We can accomplish this with the Makefile but I figured if someone ever forgot to read our docs atleast the precommit hook would catch issues and save the contributor a bit of effort instead of having to wait for the CI to catch them.

I think it would be a great tool to auto run the ruff formatting and linting, mypy type checking and our test suite pre commit. I think to avoid some of the pitfalls mentioned in the post we can select a subset of the tests that run in the pre-commit process which aren't I/O heavy operations as these tend to really lengthen out the test suite execution time.

At the same time the benefit might be more effort than its worth at the moment and the Makefiles probably suffices for now in helping simplify the development workflow

philippemiron:
It's a trade of on running part of the ci pipeline locally and making it slightly harder for new developers to start.

kevinsantana11:
@philippemiron that's a fair point, I think given clouddrift is still in alpha the trade off is not worth the benefits we'll get with the automation. Lets continue then with the `Makefile` and I'll add in a `CONTRIBUTING.md` similar to what AwkwardArray has: https://github.com/scikit-hep/awkward/blob/main/CONTRIBUTING.md.

I like their use of `Conventional Commits` documented here: https://www.conventionalcommits.org/en/v1.0.0/#summary and think it would be a good idea to leverage but I'm curious about @selipot, @philippemiron and @milancurcic thoughts. My thinking is we can use this for our commit to the main branch so commits made on PRs can have any format but the PR title would use the `Conventional Commit` with a good description since that is what will be committed given we `Squash and Merge` on PRs.

This is similar to what Awkward uses: https://github.com/scikit-hep/awkward/blob/main/CONTRIBUTING.md#git-practices

philippemiron:
I like it. We do it at work and starts the title with an emoji so it's even easier to see what was changed when looking at the logs.

Just as an example:
⭐	:star:	New / changed feature	
🐛	:bug:	 Bugfix
🔒	:lock: Security fix
♻	:recycle:	Refactoring
⚡	:zap:	 Performance/memory improvements
🎨	:art:	 Formatting
📝	:pencil: Documentation
🔍	:mag:	 Writing, refactoring, fixing automated tests
🔧	:wrench: Build scripts/CI (other non-application changes)
‼	:bangbang:	Deprecation of a feature
⛔	:no_entry:	Removal of features

and there are other examples here, https://gist.github.com/parmentf/035de27d6ed1dce0b36a.

selipot:
@kevinsantana11 please keep contributing and scope at the top before install

kevinsantana11:
@selipot @philippemiron let me know what you guys think of the new updates. If you go to this branch on my repo you can see what the new readme/contributing guides will look like. One thing I also think we should consider is only using the unicode characters for the emojis so we can also see them in the commits since ':star:' is a github specific thing. A comparison of using the unicode character vs github specific directive from the `git log` output:
```
commit 5f9c2d15594b18760854cbc914f6a180645c6361 (HEAD -> dev-tools)
Author: Kevin Santana <kevinsantana11@gmail.com>
Date:   Sat Mar 23 14:34:14 2024 -0400

    ⭐ adding another thing

commit 98de2a0de21b26863dd18137335b4bffb2a10a95
Author: Kevin Santana <kevinsantana11@gmail.com>
Date:   Sat Mar 23 14:33:43 2024 -0400

    :star: adding something
```

selipot:
I don't understand in README.md the section: 

**Locally (Makefile assumes installation of micromamba for more info [visit here](https://mamba.readthedocs.io/en/latest/index.html)):**

Can we provide more explanation?

kevinsantana11:
> I don't understand in README.md the section:
> 
> **Locally (Makefile assumes installation of micromamba for more info [visit here](https://mamba.readthedocs.io/en/latest/index.html)):**
> 
> Can we provide more explanation?

this is no longer relevant so went ahead and removed it

kevinsantana11:
Last thing I want to do here is update how our docs are generated for the website. Currently there's two competing places where we have documentation and I want to see if there's a way we can consolidate these down into one place

CONTRIBUTING.md
```diff
@@ -0,0 +1,228 @@
...
+with conda:
+
+```
+conda env create -f environment.yml
+conda activate clouddrift
+```
+
+### Testing
+
+* Follow the steps in [Preparing your environment](#preparing-your-environment)
```
selipot: Should the steps of Testing be enumerated (1., 2., 3.) instead of itemized?

selipot (commented):
LOVE IT!

selipot:
@kevinsantana11 is this ready to merge?

PR #376, opened 2024-02-28
Switch to row instead of trajectory

philippemiron:


selipot (approved):
Made a couple more edits, but I didn't run the tests. However, I think it's good to go.

milancurcic:
I think this is a change in a good direction and I'd take it one step further and use "rows" for the row dimension name rather than "traj". IIRC "traj" came from some earlier convention, perhaps a CF-conventions documentation or something like that. I think it's good to use an existing convention if it exists but it should be evaluated against whether it's actually a good choice and whether it's consistent here within CloudDrift.

philippemiron:
We could have the default has `rows`, and set it to `traj` for the GDP and other datasets we are creating.
Note that I haven't look at the code to see how much work would that required.

philippemiron:
We could have the default has `rows` and set it to `traj` for the GDP and other datasets we are creating.
Note that I haven't look at the code to see how much work would that required.

milancurcic:
Yes, I suggested defaulting to "rows". Datasets can still have any dimension name they want. The change for that should only be in `ragged.subset` (and corresponding tests) I think.

philippemiron:
> Yes, I suggested defaulting to "rows". Datasets can still have any dimension name they want. The change for that should only be in `ragged.subset` (and corresponding tests) I think.

We have it hardcoded here too [raggedarray](https://github.com/Cloud-Drift/clouddrift/blob/main/clouddrift/raggedarray.py#L449).

milancurcic:
Yes, but I think raggedarray.py is only used by GDP adapters.

philippemiron:
Actually, @kevinsantana11, also uses it with the `Hurdat2` adapter.

philippemiron:
I'll resolve the conflicts later.

philippemiron:
@kevinsantana11 could you help with this? Not sure how to proceed with the changes to `raggedarray.py`. What I wanted to add is a mapping GDP_DIMS = {"rows": "traj", "obs": "obs"}, so we can use this to change the name of the dimensions and don't have to write: `"traj"` or `"obs"` in the gdp adapters.

But I now see that we have:
```
DimNames = Literal["traj", "obs"]

@dataclass
class Dim:
    name: DimNames
    size: int

class RaggedArray:
    def __init__(
        self,
        coord_dims: list[tuple[str, Dim]],
        ....
```
in the `raggedarray` module..

kevinsantana11:
@philippemiron taking a look into this now. The change I had made to the ragged-array data structure got rid of the ability to modify "traj", "obs" dimensions. I like the changes here though so I'm going to submit a PR to your repo.

philippemiron:
Thanks

kevinsantana11:
@philippemiron looks like I should have rebased and then made changes on top of yours rather than merging main into your branch. I'll need to take a look at this later today as some git surgery will be necessary. 

I think my approach will be to remove the most recent merge from main, squash all commits here into 1 commit and then rebase main. This way when I rebase, all of the changes are made on-top of the most recent commits and so when fixing the merge conflicts I'll just need to keep the current changes. I'm going to test this out separately in my own branch first before pushing into here. Just wanted to give you a heads up.

philippemiron:
I remove the conflict but there is still work to be done. We have two things that are conflicting at the moment.

I had the `name_dims = {"rows": "name_use_as_alias", "obs": "name_use_as_alias"}`, with a default to simply `name_dims = {"rows": "rows", "obs": "obs}`, and for example for GDP `name_dims = {"rows": "traj", "obs": "obs}`. 

In `raggedarray.py`, this was written in the opposite way: `name_dims = {"name_use_as_alias": "rows", "name_use_as_alias": "obs"}`. I think it makes more sense how I had it, but I might be missing something.

We still need to have something for the coordinates that maps each coordinate to a dimension. And this part could be confusing too.. We could only map to `rows` and `obs` here, but then the dimension could be renamed if we supply something to `name_dims`.

Comments?

kevinsantana11:
> I remove the conflict but there is still work to be done. We have two things causing issues.
> 
> I had the `name_dims = {"rows": "name_use_as_alias", "obs": "name_use_as_alias"}`, with a default set to `name_dims = {"rows": "rows", "obs": "obs}`, and for example for GDP `name_dims = {"rows": "traj", "obs": "obs}` to rename the `"rows"` dimension.
> 
> In `raggedarray.py`, this was written in the opposite way: `name_dims = {"name_use_as_alias": "rows", "name_use_as_alias": "obs"}`. I think it makes more sense how I had it, but I might be missing something.
> 
> We still need to have something for the coordinates that maps each coordinate to a dimension. And this part could be confusing too.. We could only map to `rows` and `obs` here, but then the dimension could be renamed if we supply something to `name_dims`.
> 
> Comments?

I mainly reversed the relationship as it made certain parts of the code a bit easier to deal with.

In relation to mapping coordinates to dimension, it seems when converting from xarray we can 'detect' the (coord, dim) mapping [ref](https://github.com/philippemiron/clouddrift/blob/351-trajectory-or-row/clouddrift/raggedarray.py#L237). The current approach is to look at the coordinates dimension and pick the last one.

When it comes to awkward arrays, since their data structure doesn't provide such information we need that to be provided from the user. I think it could become a source of error/confusion for users to have to create maps of alias -> dim and then coord -> alias where they could mistakenly enter the dim instead and it wouldn't be until runtime that they get an error for this.

We could try to catch this preemptively and throw a descriptive error

clouddrift/adapters/gdp1h.py
```diff
@@ -205,77 +205,95 @@ def preprocess(index: int, **kwargs) -> xr.Dataset:
...
+        (gdp.GDP_DIMS["traj"]),
         [False if ds.get("location_type") == "Argos" else True],
     )  # 0 for Argos, 1 for GPS
-    ds["DeployingShip"] = (("traj"), gdp.cut_str(ds.DeployingShip, 20))
-    ds["DeploymentStatus"] = (("traj"), gdp.cut_str(ds.DeploymentStatus, 20))
-    ds["BuoyTypeManufacturer"] = (("traj"), gdp.cut_str(ds.BuoyTypeManufacturer, 20))
-    ds["BuoyTypeSensorArray"] = (("traj"), gdp.cut_str(ds.BuoyTypeSensorArray, 20))
+    ds["DeployingShip"] = ((gdp.GDP_DIMS["traj"]), gdp.cut_str(ds.DeployingShip, 20))
+    ds["DeploymentStatus"] = (
+        (gdp.GDP_DIMS["traj"]),
```
philippemiron, line 219: But there is a problem here, as I said I was defining: `gdp.GDP_DIMS = {"rows": "raj," "obs": "traj"}`, so we can write adapters and put `(GDP.GDP_DIMS["rows"])` or `(GDP.GDP_DIMS["obs"])` as the dimension, and it would be renamed if needed. As we have it now, then the GDP would have "rows" as it's dimension.

clouddrift/raggedarray.py
```diff
@@ -17,67 +16,69 @@
...
-        coord_dims: list[tuple[str, Dim]],
         coords: dict,
         metadata: dict,
         data: dict,
-        attrs_global: dict = {},
-        attrs_variables: dict = {},
+        attrs_global: Optional[dict] = {},
+        attrs_variables: Optional[dict] = {},
+        name_dims: dict[str, DimNames] = {},
+        coord_dims: dict[str, str] = {}
```
philippemiron, line 31: Is this use?

clouddrift/raggedarray.py
```diff
@@ -169,81 +172,90 @@ def from_netcdf(cls, filename: str):
...
+    def from_xarray(
+        cls, ds: xr.Dataset, rows_dim_name: str = "rows", obs_dim_name: str = "obs"
+    ):
         """Populate a RaggedArray instance from an xarray Dataset instance.
 
         Parameters
         ----------
         ds : xr.Dataset
             Xarray Dataset from which to load the RaggedArray
+        dim_rows : str, optional
```
philippemiron, line 215: variable names are different in the docstring

philippemiron:
I think it's good now.

clouddrift/ragged.py
```diff
@@ -667,7 +668,7 @@ def subset(
     :func:`apply_ragged`
     """
     mask_traj = xr.DataArray(
```
selipot, line 670: Should we replace in the code the `mask_traj` variables by mask_row`? It is not needed bu t will be consistent.

selipot:
Ok, added my review with a couple of edits and a suggestion above. @kevinsantana11 please review also.

selipot:
> @selipot I'm working on adding the doctest to atleast run for our ragged module, have most of the changes done just have a couple more changes to be made.

We agreed this would be for another PR

PR #374, opened 2024-02-22
Rework generated ragged array to utilize new conventions for coordinates

kevinsantana11:
fixes #372 and reworks the coordinates to `id(traj)` and `time(obs)`

PR #378, opened 2024-02-29
⭐ Hurdat2 adapter

kevinsantana11:
This implements a new ragged array data adapter for the HURDAT2 dataset [Reference](https://www.aoml.noaa.gov/hrd/hurdat/Data_Storm.html). Dependent on the merge of: https://github.com/Cloud-Drift/clouddrift/pull/374


selipot:
@kevinsantana11 should you add a corresponding function in the `clouddrift.datasets` module? it would look like
```
def hurdat2(decode_times: bool = True) -> xr.Dataset:
    """Returns the dataset of ...

    The function will first look for the ragged-array dataset on the local
    filesystem. If it is not found, the dataset will be downloaded using the
    corresponding adapter function and stored for later access.

    The upstream data is available at https://www.aoml.noaa.gov/hrd/hurdat/hurdat2.html.
```

Otherwise I do not really know how to test this PR.

milancurcic:
As a prospective user, I'd be happy to test and review this. :)

milancurcic:
First thing before I go any further. There are many changes/additions here unrelated to HURDAT2. Was it intentional?

milancurcic:
I see, it looks like it has changes from #374 already. Let's open a clean PR for HURDAT2 only forked from the latest main.

selipot:
> I see, it looks like it has changes from #374 already. Let's open a clean PR for HURDAT2 only forked from the latest main.

@kevinsantana11 suggested otherwise since this PR depends on #374

kevinsantana11:
Changes here depend on some of the new changes made to the ragged array data structure which are introduced in #374 . While I could fork from main this would have me introduce the adapter with the old conventions and then the changes to support the new conventions would need to be made in #374 instead of here.

I would prefer to merge in #374 once I triage the failed builds and then merge this in to reduce unnecessary churn.

kevinsantana11:
> Changes here depend on some of the new changes made to the ragged array data structure which are introduced in #374 . While I could fork from main this would have me introduce the adapter with the old conventions and then the changes to support the new conventions would need to be made in #374 instead of here.
> 
> I would prefer to merge in #374 once I triage the failed builds and then merge this in to reduce unnecessary churn.

@milancurcic @philippemiron @selipot

milancurcic:
All good, I wasn't aware of the dependency. Carry on! I'll review after #374 is merged and we have a cleaner diff.

selipot:
@kevinsantana11 please move your draft example notebook outside of this PR into its own repository under the github Clouddrift project.

selipot:
See below my attempt. It runs but the `id` and `time` variables do not look right. Next using the datasets method `datasets.hurdat2()` does not seem to find the temporary folder with the existing data? 

Screenshot 2024-03-06 at 11 15 31 AM


Screenshot 2024-03-06 at 11 17 31 AM

selipot:
@kevinsantana11 shouldn't we use the `CLOUDDRIFT_PATH` like for the other datasets that are generated with adapters?

kevinsantana11:
> @kevinsantana11 shouldn't we use the `CLOUDDRIFT_PATH` like for the other datasets that are generated with adapters?

I believe here you're referring to the usage of the following (similar) code snippet:
```
<adapter_name>_TMP_PATH = os.path.join(tempfile.gettempdir(), "clouddrift", "<adapter_name>")
```

which we do utilize across all of the adapters minus the mosaic and glad dataset adapters which don't create temporary file storage for the intermediary files.

We utilize the `CLOUDDRIFT_PATH` environment variable in our datasets module instead of using the `tempfile` library for defining the file path for storing the dataset. I'm not sure if it would be better to utilize the `tempfile` library here as well or just continue with the current implementation of storing datasets either in the path defined by the CLOUDDRIFT_PATH environment variable or in the invoking users home path under `.clouddrift`.

I think keeping this logic separate makes sense as one set of temporary files are just intermediary files for creating the unified RaggedArray/ xarray Datasets and the other files are the datasets themselves. Let me know your thoughts on this @selipot

selipot:
From the page https://www.aoml.noaa.gov/hrd/hurdat/Data_Storm.html I see that there are two datasets. One for the Atlantic Basin and one for the Pacific Basin. @kevinsantana11 I think this current adapter only takes in the Atlantic basin data at https://www.aoml.noaa.gov/hrd/hurdat/hurdat2.html. 

How should we include the data from the Pacific basin at https://www.aoml.noaa.gov/hrd/hurdat/hurdat2-nepac.html ?

@milancurcic I think you were interested in this dataset. Should we split in two adapters/datasets or combine them?

milancurcic:
I didn't know East Pacific storms are in a separate HURDAT2. I would join them on the fly in the adapter and serve as a single dataset from `clouddrift.datasets`.

selipot (changes requested):
Need to fix format of id and time. Also should include the dataset from the Pacific Basin from the same page.

philippemiron:
> I didn't know East Pacific storms are in a separate HURDAT2. I would join them on the fly in the adapter and serve as a single dataset from `clouddrift.datasets`.

Maybe as an option? East/West/combined since a lot of statistics are calculated by basin.

kevinsantana11:
@philippemiron that's a great idea, I will make it so this is a parameter option users can define and configure. I think by default it will make sense for the dataset to return the combined datasets let me know what you guys think @selipot , @milancurcic.

selipot:
I think @philippemiron 's suggestion is a good one!

kevinsantana11:
@selipot updated the PR to include the feature @philippemiron had mentioned. I also modified the gdp6h and gdp1h tests to work with less mocking just to simplify the tests and make them a bit more readable.

clouddrift/adapters/hurdat2.py
```diff
@@ -0,0 +1,452 @@
...
+_ATLANTIC_BASIN_URL = "https://www.aoml.noaa.gov/hrd/hurdat/hurdat2.html"
+_NORTHEAST_PACIFIC_BASIN_URL = "https://www.aoml.noaa.gov/hrd/hurdat/hurdat2-nepac.html"
+
+_DEFAULT_FILE_PATH = os.path.join(tempfile.gettempdir(), "clouddrift", _DEFAULT_NAME)
+os.makedirs(_DEFAULT_FILE_PATH, exist_ok=True)
+
+
+class BasinOption(int, enum.Enum):
+    ATLANTIC = 1
+    NORTHEAST_PACIFIC = 2
```
philippemiron, line 26: Not sure about the `NORTHEAST_PACIFIC` name. One place they call the dataset: 
`Northeast and North Central Pacific hurricane database (HURDAT2) 1949-2022`, and just below `East Pacific Basin`. 

Also, loading the pacific dataset, the `ds.basin` variables,
```
In [22]: np.unique(ds.basin.values)
Out[22]: array(['CP', 'EP'], dtype='<U2')
```
takes two values: Central Pacific and Eastern Pacific. So maybe we just renamed this pacifc to `PACIFIC`?

clouddrift/adapters/hurdat2.py
```diff
@@ -0,0 +1,452 @@
...
+            "comments": "basin of origin.",
+        }
+    )
+    cyclone_number: int = field(
+        metadata={"long_name": "ATCF cyclone number for associated year."}
+    )
+    year: int = field(metadata={"long_name": "Year"})
+    name: str = field(metadata={"long_name": "Name", "na_values": "UNNAMED"})
+    rowsize: int = field(metadata={"long_name": "Number of best track entries"})
+    id: str = field(default_factory=lambda: uuid4().hex)
```
philippemiron, line 101: Not sure that having this id has a random string makes much sense since reloading two times the same datasets gives different ids.

```
from clouddrift.datasets import hurdat2
from clouddrift.adapters.hurdat2 import BasinOption

ds = hurdat2(option=BasinOption.NORTHEAST_PACIFIC)
ds2 = hurdat2(option=BasinOption.NORTHEAST_PACIFIC)

print(ds.id[0])
print(ds2.id[0])

>>>>
<xarray.DataArray 'id' ()> Size: 128B
array('42bdb206615f40dba48263a9ec1a3dbc', dtype='<U32')
Coordinates:
    id       <U32 128B '42bdb206615f40dba48263a9ec1a3dbc'
<xarray.DataArray 'id' ()> Size: 128B
array('8f84cae42aeb47348d1e50774020cad0', dtype='<U32')
Coordinates:
    id       <U32 128B '8f84cae42aeb47348d1e50774020cad0'
```

can't we just assign increasing (from the start date) integer for each tracks?

philippemiron:
> @selipot updated the PR to include the feature @philippemiron had mentioned. I also modified the gdp6h and gdp1h tests to work with less mocking just to simplify the tests and make them a bit more readable.

Small comments, it would also be easier to keep PR for implementing one thing. Now we have hurdat with random gdp modifications that make it more complicated to review.

selipot:
> > @selipot updated the PR to include the feature @philippemiron had mentioned. I also modified the gdp6h and gdp1h tests to work with less mocking just to simplify the tests and make them a bit more readable.
> 
> Small comment, it would also be easier to keep PR for implementing one thing. Now we have hurdat with random gdp modifications that make it more complicated to review.

I agree, this is not good form

clouddrift/adapters/hurdat2.py
```diff
@@ -0,0 +1,452 @@
...
+
+
+@dataclass
+class DataLine:
+    time: float = field(
+        metadata={"comments": "Computed property from YYY-MM-DD HH:MM in UTC"}
+    )
+    record_identifier: RecordIdentifier = field(
+        metadata={"long_name": "Record Idenfier", "comments": RecordIdentifier.__doc__}
```
philippemiron, line 110: Using .__doc__ introduces an extra change of line.

```
In [5]: ds.record_identifier
Out[5]:
<xarray.DataArray 'record_identifier' (obs: 30238)> Size: 121kB
array(['R', 'R', 'R', ..., 'R', 'R', 'R'], dtype='<U1')
Coordinates:
    time     (obs) float64 242kB -6.488e+08 -6.488e+08 ... 1.665e+09 1.665e+09
Dimensions without coordinates: obs
Attributes:
    long_name:  Record Idenfier
    comments:   \n    C – Closest approach to a coast, not followed by a land...
```

clouddrift/adapters/hurdat2.py
```diff
@@ -0,0 +1,452 @@
...
+    )
+    system_status: SystemStatus = field(
+        metadata={"long_name": "System Status", "comments": SystemStatus.__doc__}
+    )
+    lat: float = field(
+        metadata={
+            "long_name": "Latitude",
+            "units": "degrees",
```
philippemiron, line 118: We should try to use CF conventions for `standard_name`, `long_name`, `units`, etc. See [CF conventions][https://cfconventions.org/Data/cf-standard-names/current/build/cf-standard-name-table.html].

clouddrift/adapters/hurdat2.py
kevinsantana11: Fair point, let me work on this change

kevinsantana11:
> > > @selipot updated the PR to include the feature @philippemiron had mentioned. I also modified the gdp6h and gdp1h tests to work with less mocking just to simplify the tests and make them a bit more readable.
> > 
> > 
> > Small comment, it would also be easier to keep PR for implementing one thing. Now we have hurdat with random gdp modifications that make it more complicated to review.
> 
> I agree, this is not good form

understood, I'll remove these changes

clouddrift/adapters/hurdat2.py
kevinsantana11: Makes sense to me, updated.

clouddrift/adapters/hurdat2.py
kevinsantana11: updated to generate ids sequentially and assigns them chronologically based off of either the genesis data record or the earliest time we can find.

clouddrift/adapters/hurdat2.py
kevinsantana11: addressed this in new commits

clouddrift/adapters/hurdat2.py
kevinsantana11: @philippemiron went ahead and updated the metadata and the also converted some of the fields into the proper units

kevinsantana11:
@selipot this PR is ready for re-review

clouddrift/adapters/hurdat2.py
```diff
@@ -0,0 +1,498 @@
...
+        }
+    )
+    system_status: SystemStatus = field(
+        metadata={"standard_name": "System Status", "comments": SystemStatus.__doc__}
+    )
+    lat: float = field(
+        metadata={
+            "standard_name": "latitude",
+            "units": "degree_north",
+            "comment": "Latitude is positive northward; its units of degree_north (or equivalent) indicate this explicitly. In a latitude-longitude system defined with respect to a rotated North Pole, the standard name of grid_latitude should be used instead of latitude. Grid latitude is positive in the grid-northward direction, but its units should be plain degree.",
```
selipot, line 122: I understand this is from CF conventions but please remove

clouddrift/adapters/hurdat2.py
```diff
@@ -0,0 +1,498 @@
...
+        }
+    )
+    lon: float = field(
+        metadata={
+            "standard_name": "Longitude",
+            "units": "degree_east",
+            "comment": "Longitude is positive eastward; its units of degree_east (or equivalent) indicate this explicitly. In a latitude-longitude system defined with respect to a rotated North Pole, the standard name of grid_longitude should be used instead of longitude. Grid longitude is positive in the grid-eastward direction, but its units should be plain degree.",
```
selipot, line 129: please remove

clouddrift/adapters/hurdat2.py
```diff
@@ -0,0 +1,498 @@
...
+        }
+    )
+    year: int = field(metadata={"long_name": "Year"})
+    rowsize: int = field(metadata={"comment": "Number of best track entries"})
+
+
+@dataclass
+class DataLine:
+    time: float = field(
+        metadata={"comments": "Computed property from YYY-MM-DD HH:MM in UTC"}
```
selipot, line 107: the time string should be converted to a datetime object or something similar, right now I think it converts a string to a float and returns numbers that do not make sense to me

clouddrift/adapters/hurdat2.py
```diff
@@ -0,0 +1,498 @@
...
+            "standard_name": "Longitude",
+            "units": "degree_east",
+            "comment": "Longitude is positive eastward; its units of degree_east (or equivalent) indicate this explicitly. In a latitude-longitude system defined with respect to a rotated North Pole, the standard name of grid_longitude should be used instead of longitude. Grid longitude is positive in the grid-eastward direction, but its units should be plain degree.",
+        }
+    )
+    wind_speed: float = field(  # TODO: needs conversion
+        metadata={
+            "standard_name": "wind_speed",
+            "units": "m s-1",
+            "comment": "Speed is the magnitude of velocity. Wind is defined as a two-dimensional (horizontal) air velocity vector, with no vertical component. (Vertical motion in the atmosphere has the standard name upward_air_velocity.) The wind speed is the magnitude of the wind velocity.",
```
selipot, line 136: please remove

clouddrift/adapters/hurdat2.py
```diff
@@ -0,0 +1,498 @@
...
+                "max_sustained_wind_speed_radius": (
+                    ["obs"],
+                    np.array(
+                        [line.max_sustained_wind_speed_radius for line in self.data]
+                    ),
+                ),
+            },
+            coords={
+                "id": (["traj"], np.array([id_])),
+                "time": (
```
selipot, line 331: I think here you should convert to a datetime object.

clouddrift/adapters/hurdat2.py
```diff
@@ -0,0 +1,498 @@
...
+_DEFAULT_FILE_PATH = os.path.join(tempfile.gettempdir(), "clouddrift", _DEFAULT_NAME)
+os.makedirs(_DEFAULT_FILE_PATH, exist_ok=True)
+
+_METERS_IN_NAUTICAL_MILES = (
+    1825  # source: https://en.wikipedia.org/wiki/Nautical_mile#cite_note-BIPM-2
+)
+_PASCAL_PER_MILLI_BAR = 100
+
+
+class BasinOption(int, enum.Enum):
```
selipot, line 28: I do not know how to use these

clouddrift/datasets.py
```diff
@@ -12,6 +12,7 @@
 import xarray as xr
 
 from clouddrift import adapters
+from clouddrift.adapters.hurdat2 import BasinOption
```
selipot, line 15: How does one use this?

clouddrift/datasets.py
```diff
@@ -232,6 +233,53 @@ def glad(decode_times: bool = True) -> xr.Dataset:
...
+
+    The function will first look for the ragged-array dataset on the local
+    filesystem. If it is not found, the dataset will be downloaded using the
+    corresponding adapter function and stored for later access.
+
+    The upstream data is available at https://www.aoml.noaa.gov/hrd/hurdat/hurdat2.html
+
+    Parameters
+    ----------
+    option : BasinOption, default: BOTH
```
selipot, line 248: how does one use this? `ds = hurdat2(option=1)`? Here list what the option should be 1,2,3? Or "Atlantic", "Pacific","Both" (Default)?

clouddrift/datasets.py
```diff
@@ -232,6 +233,53 @@ def glad(decode_times: bool = True) -> xr.Dataset:
...
+        Basin dataset(s) to be downloaded.
```
selipot, line 249: Basin dataset(s) to be returned?

clouddrift/datasets.py
```diff
@@ -232,6 +233,53 @@ def glad(decode_times: bool = True) -> xr.Dataset:
...
+    Returns
+    -------
+    xarray.Dataset
+        HURDAT2 dataset as a ragged array
+
+    Examples TODO - EVERYTHING BELOW NEEDS TO BE UPDATED
```
selipot, line 255: yes please

selipot:
Things to work on is 

1. The formatting/conversion of the datetime string
2. The simplification/explanation on how to use the basin option
3. The writing of the docstring, thinking of a beta user
4. The check on previous data download. In my case this does not seem to work:

This
```
import clouddrift
from clouddrift.adapters import hurdat2
import clouddrift.datasets as datasets

ra = hurdat2.to_raggedarray()
ds = ra.to_xarray()
ds1 = datasets.hurdat2()
```
leads to
```
Cannot determine the file has been updated on the remote source.                               'Last-Modified' header not present.
Cannot determine the file has been updated on the remote source.                               'Last-Modified' header not present.
https://www.aoml.noaa.gov/hrd/hurdat/hurdat2-nepac.html: 3.65MB [00:00, 5.83MB/s]
https://www.aoml.noaa.gov/hrd/hurdat/hurdat2-nepac.html: 3.72MB [00:00, 5.32MB/s]
https://www.aoml.noaa.gov/hrd/hurdat/hurdat2.html: 6.61MB [00:01, 6.70MB/s]
Retrieving the number of obs: 100%|████| 3143/3143 [00:00<00:00, 3220790.98it/s]
Filling the Ragged Array: 100%|███████████| 3143/3143 [00:01<00:00, 2977.23it/s]
Cannot determine the file has been updated on the remote source.                               'Last-Modified' header not present.
Cannot determine the file has been updated on the remote source.                               'Last-Modified' header not present.
https://www.aoml.noaa.gov/hrd/hurdat/hurdat2-nepac.html: 2.82MB [00:00, 5.70MB/s]
https://www.aoml.noaa.gov/hrd/hurdat/hurdat2-nepac.html: 3.72MB [00:00, 5.90MB/s]
https://www.aoml.noaa.gov/hrd/hurdat/hurdat2.html: 6.61MB [00:00, 7.93MB/s]
Retrieving the number of obs: 100%|████| 3143/3143 [00:00<00:00, 3504172.64it/s]
Filling the Ragged Array: 100%|███████████| 3143/3143 [00:01<00:00, 3058.32it/s]
```

kevinsantana11:
> Things to work on are:
> 
>     1. The formatting/conversion of the datetime string
> 
>     2. The simplification/explanation on how to use the basin option
> 
>     3. The writing of the docstring, thinking of a beta user
> 
>     4. The check on previous data download. In my case this does not seem to work:
> 
> 
> This
> 
> ```
> import clouddrift
> from clouddrift.adapters import hurdat2
> import clouddrift.datasets as datasets
> 
> ra = hurdat2.to_raggedarray()
> ds = ra.to_xarray()
> ds1 = datasets.hurdat2()
> ```
> 
> leads to
> 
> ```
> Cannot determine the file has been updated on the remote source.                               'Last-Modified' header not present.
> Cannot determine the file has been updated on the remote source.                               'Last-Modified' header not present.
> https://www.aoml.noaa.gov/hrd/hurdat/hurdat2-nepac.html: 3.65MB [00:00, 5.83MB/s]
> https://www.aoml.noaa.gov/hrd/hurdat/hurdat2-nepac.html: 3.72MB [00:00, 5.32MB/s]
> https://www.aoml.noaa.gov/hrd/hurdat/hurdat2.html: 6.61MB [00:01, 6.70MB/s]
> Retrieving the number of obs: 100%|████| 3143/3143 [00:00<00:00, 3220790.98it/s]
> Filling the Ragged Array: 100%|███████████| 3143/3143 [00:01<00:00, 2977.23it/s]
> Cannot determine the file has been updated on the remote source.                               'Last-Modified' header not present.
> Cannot determine the file has been updated on the remote source.                               'Last-Modified' header not present.
> https://www.aoml.noaa.gov/hrd/hurdat/hurdat2-nepac.html: 2.82MB [00:00, 5.70MB/s]
> https://www.aoml.noaa.gov/hrd/hurdat/hurdat2-nepac.html: 3.72MB [00:00, 5.90MB/s]
> https://www.aoml.noaa.gov/hrd/hurdat/hurdat2.html: 6.61MB [00:00, 7.93MB/s]
> Retrieving the number of obs: 100%|████| 3143/3143 [00:00<00:00, 3504172.64it/s]
> Filling the Ragged Array: 100%|███████████| 3143/3143 [00:01<00:00, 3058.32it/s]
> ```

as we had discussed this sort of behavior is expected since the remote server doesn't give us a way to determine if the file on the server is newer than the local file. I updated the comment message here to read better and hopefully make clearer what's going on. I removed the comments for the CF conventions, added the `decode_time` bool option and updated the doc string.

Last thing missing here is some help with the ref portion of the dataset functions doc string I believe.

selipot:
@kevinsantana11 should we call the id variable `atcf_id` or `id_atcf` or simply `id` and add in the attributes the standard name and a comment that says something like: id is the atcf etc?

kevinsantana11:
> @kevinsantana11 should we call the id variable `atcf_id` or `id_atcf` or simply `id` and add in the attributes the standard name and a comment that says something like: id is the atcf etc?

Hmmm I do like using the `atcf_id` as its a bit more explicit but I'm not married to it. I think using the comment section and long name can be used for providing more details and we can use `id` for consistency across datasets. Updating to reflect this now

kevinsantana11:
@selipot this looks good to go just need approval to merge

selipot:
@kevinsantana11 it looks good, as an example
```
from clouddrift.datasets import hurdat2
from clouddrift.plotting import plot_ragged
import numpy as np
import matplotlib.pyplot as plt
import cartopy.crs as ccrs
from mpl_toolkits.axes_grid1 import make_axes_locatable
# %%
ds = hurdat2()

# %%
fig = plt.figure(figsize=(10, 5))
ax = fig.add_subplot(1, 1, 1, projection=ccrs.Mollweide())
l = plot_ragged(
    ax,
    ds.lon,
    ds.lat,
    ds.rowsize,
    colors=ds.wind_speed.values,
    transform=ccrs.PlateCarree(),
    cmap="inferno",
)
ax.set_extent([-180, 180, 0, 80], crs=ccrs.PlateCarree())
ax.coastlines()
ax.gridlines(draw_labels=True)
divider = make_axes_locatable(ax)
cax = divider.append_axes('right', size="3%", pad=0.25, axes_class=plt.Axes)
fig.colorbar(l, cax=cax)
```
gives
e378be45-f36e-4715-9e9e-bf93b599b080

Yet you will see that some wind speed are apparently negative, which is not correct. I think it is because you do not process the missing data that appears to all be `-999` for all variables. You then convert some of these missing data in your preprocessing, such as converting millibar to pascal etc. In general I think you need to replace missing data by `np.nan` before applying conversions. Otherwise you end up with misleading/erroneous data. Almost there.

selipot:
@kevinsantana11 Why is the variable `year` of dimension `traj` a 3-digit number?

selipot (approved):

selipot:
Looks good to me @kevinsantana11 Please merge.

Issue #380, opened 2024-03-12
Integrate `doctest` into our test suite


kevinsantana11:
This tool tests our doc string code to ensure any code changes that would break existing code examples can be addressed pro-actively.

kevinsantana11:
This has been distributed into several tasks one per module

PR #381, opened 2024-03-12
🔧 Integrate doctest to run on our `ragged.py` module

kevinsantana11:
Tool Documentation: https://docs.python.org/3/library/doctest.html#

Currently the integration only runs on our ragged.py module. This is related to #380


selipot:
@kevinsantana11 does this tool automatically correct the docstrings or do we do it manually? And what is the strategy? To do a PR for each module?

kevinsantana11:
> @kevinsantana11 does this tool automatically correct the docstrings or do we do it manually? And what is the strategy? To do a PR for each module?

This won't automatically correct our docstrings for us. What it will do is run the code it finds in our docstrings and detect if the code is invalid (wrong syntax, missing imports, throws an exception, etc...) or provides the wrong output (the expected output follows the code defined by the 3 angle brackets >>>). Similarly to unit tests if the doctests fail we need to go in and manually fix the code.

As far as strategy maybe we should create a project where we have one issue per module. I can then pickup a doctest issues roughly bi-weekly

I want to do this for mypy typing as well as that also need a migration strategy that's done over time.

kevinsantana11:
starts #398

PR #386, opened 2024-03-21
🔎 Refactor plot_ragged function to improve clarity and documentation

selipot:
This PR is to clarify the docstring of ``plot_ragged``. @philippemiron let me know what you think.

### DIFF
--- a/clouddrift/plotting.py
+++ b/clouddrift/plotting.py
@@ -1,5 +1,5 @@
 """
-This module provides a function to easily and efficiently plot trajectories stored in a ragged array.
+This module provides a function to easily and efficiently plot the rows of a ragged array.
 """
 
 from typing import Optional, Union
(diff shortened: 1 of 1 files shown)


clouddrift/plotting.py
```diff
@@ -41,12 +41,12 @@ def plot_ragged(
     *args : tuple
         Additional arguments to pass to ``ax.plot``.
     colors : array-like
-        Colors to use for plotting. If colors is the same shape as longitude and latitude,
-        the trajectories are splitted into segments and each segment is colored according
-        to the corresponding color value. If colors is the same shape as rowsize, the
-        trajectories are uniformly colored according to the corresponding color value.
+        Values to map on the current colormap. If ``colors`` is the shape of ``rowsize``,
+        the data points of each row are uniformly colored according to the color value for the row.
```
philippemiron, line 45: Should we removed "uniformly" here? It is colored as a function of the values that you supply, not necessarily uniform.

philippemiron (approved):

clouddrift/plotting.py
kevinsantana11: Maybe we could say:
```
Colors to use for plotting. If ``colors`` has the same shape as ``rowsize``,
each ``ob`` in a ragged array is colored according to the ``row`` it is associated with.
```

philippemiron:
There was a comment that wasn't resolved...

selipot:
Resolved now! I took a decision to move forward.

PR #387, opened 2024-04-04
🐛 use zarr archive for gdp6h dataset

kevinsantana11:
This updates the gdp6h dataset to utilize a zarr archive for retrieving the dataset as it better supports lazy loading and also helps us avoid the download bug involved with the netcdf4 lib where the byte stream downloading features times out after 100s.


selipot:
I have updated the zarr file on the bucket with the latest conventions. We still need to refactor the docstring once we get the green light from @RickLumpkin

selipot:
@kevinsantana11 I think it is good to merge?

PR #388, opened 2024-04-04
🔧 update retry config and keep completed downloads

kevinsantana11:
I believe the retry logic we had before hand was a bit to aggresive resulting again and again in failed downloads. This makes the logic a lot more forgiving.

Before hand the max retry using the description provided [here](https://tenacity.readthedocs.io/en/latest/api.html#tenacity.wait.wait_exponential_jitter)

### DIFF
--- a/clouddrift/adapters/utils.py
+++ b/clouddrift/adapters/utils.py
@@ -23,7 +23,7 @@ _standard_retry_protocol: Callable[[WrappedFn], WrappedFn] = retry(
     retry=retry_if_exception(
         lambda ex: isinstance(ex, (requests.Timeout, requests.HTTPError))
     ),
-    wait=wait_exponential_jitter(initial=1.25),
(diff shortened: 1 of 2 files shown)


selipot:
If a file is already present in the `tmp_path` it does not attempt to download again, correct?

kevinsantana11:
> If a file is already present in the `tmp_path` it does not attempt to download again, correct?

This is correct when the entire download succeeds. If one of the downloads fails we cancel all downloads and cleanup any created files in case any of them are corrupted.

I can change this logic to instead work in a way to where if the download fails we only cleanup the files that are currently in progress and keep the ones that successfully downloaded.

kevinsantana11:
@selipot this PR is ready for review. Please have a look when you get a chance

selipot:
@kevinsantana11 I have tested `adapter.gdp6.to_raggedarray` using pre-dowloaded 6hourly files as well as by redownloading everything. As a conclusion it's good to merge for me.

PR #389, opened 2024-04-10
🐛 fix mosaic dataset bug, using empty buffers

### LINKED ISSUE #379: Mosaic datasets is broken
I saw that there were no tests for the mosaic dataset. After adding:

```
    def test_mosaic(self):
        with datasets.mosaic() as ds:
            self.assertTrue(ds)
```
I see that the dataset is not built properly. I try to quickly look, but it's using Buffers and the new download class that I'm not super familiar with, so I'm just reporting here.

I thought that:
```
    buffers = [BytesIO(b"") for _ in range(len(sorted_data_urls))]
    requests = [(url, BytesIO(b""), None) for url in sorted_data_urls]
```
should be:
```
    buffers = [BytesIO() for _ in range(len(sorted_data_urls))]
    requests = [(url, buffer, None) for url, buffer in zip(sorted_data_urls, buffers)]
```
but it did not fix it.

kevinsantana11:
This fixes #379 the mosaic dataset bug where the data we were referencing to create the pandas data frame was empty.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,24 +19,27 @@ jobs:
       fail-fast: false
       matrix:
         os: ["ubuntu-latest", "macos-latest", "windows-latest"]
-        python-version: ["3.9", "3.10", "3.11", "3.12"]
+        python-version: ["3.10", "3.11", "3.12"]
     steps:
       - uses: actions/checkout@v4
       - uses: mamba-org/setup-micromamba@v1
         with:
           environment-file: environment.yml
           environment-name: clouddrift
+          create-args: >-
+            python=${{ matrix.python-version }}
+            coverage
+            matplotlib-base
+            cartopy
       - name: Run unit tests
         shell: bash -l {0}
-        run: |
-              micromamba install coverage matplotlib-base cartopy
-              coverage run -m unittest discover -s tests -p "*.py"
+        run: coverage run -m unittest discover -s tests -p "*.py"
       - name: Create coverage report
         shell: bash -l {0}
         run: |
               coverage xml
       - name: Upload coverage report to Codecov
-        if: matrix.os == 'ubuntu-latest' && matrix.python-version == '3.9'
+        if: matrix.os == 'ubuntu-latest' && matrix.python-version == '3.10'
         uses: codecov/codecov-action@v3
         with:
           file: ./coverage.xml
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -15,7 +15,7 @@ jobs:
       fail-fast: false
       matrix:
         os: ["ubuntu-latest"]
-        python-version: ["3.9"]
+        python-version: ["3.10"]
     steps:
     - uses: actions/checkout@v4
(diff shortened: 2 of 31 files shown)


philippemiron:
Weird because this is what I tried .. using`zip(urls, buffers)`.. but it was not working. I guess I has something wrong..

kevinsantana11:
> Weird because this is what I tried, using `zip(sorted_data_urls, buffers)`, but it was not working. I guess I has something wrong..

That should have worked, might have been missing the third part of the tuple which is None if the size isn't known before hand.

philippemiron:
See in the issue.

```
buffers = [BytesIO() for _ in range(len(sorted_data_urls))]
requests = [(url, buffer, None) for url, buffer in zip(sorted_data_urls, buffers)]
```

I think it's actually cleaner since we don't need the index in that case.

kevinsantana11:
> See in the issue.
> 
> ```
> buffers = [BytesIO() for _ in range(len(sorted_data_urls))]
> requests = [(url, buffer, None) for url, buffer in zip(sorted_data_urls, buffers)]
> ```
> 
> I think it's actually cleaner since we don't need the index in that case.

agreed! I want to eventually make it so the src, dst are the only things that need to be provided and then optionally the size can be passed as well.

selipot:
@kevinsantana11 I have tested this and I was able to generate the mosaic dataset by calling `clouddrift.datasets.mosaic()` and it seems fine. Yet I was getting a lot of download errors?
 
Screenshot 2024-04-10 at 4 43 00 PM

philippemiron:
What is the error on the right? We might be trying to download too many files in parallel from the server so the download sometimes failed. But with the retry mechanism in place, it eventually picks it up.

Also, in general it is more useful to paste the code instead of including screenshot.

selipot:
the code is at the top of the screenshot

philippemiron:
Again it would be faster to copy and paste (I agree this one is only ~10 characters).

I'm referring to the end of this:

```
error: HTTPSConnectionPool(host....
```

selipot:
> Again it would be faster to copy and paste (I agree this one is only ~10 characters).
> 
> But I was referring to the end of this error code:
> 
> ```
> error: HTTPSConnectionPool(host....
> ```

Interactive session in VS Code is now closed sorry. Next time I can indeed copy paste 500 lines of error messages.

philippemiron:
Just tried it and sadly couldn't reproduce...

```
>>> from clouddrift.datasets import mosaic
>>> ds = mosaic()
/Users/pmiron/.clouddrift/data/mosaic.nc not found; download from upstream repository.
Downloading data: 100%|██████████████████████████████████████████████████████████████████████████████████████████████| 216/216 [00:44<00:00,  4.91Files/s]
>>> 
```

kevinsantana11:
Below is an example of what it will look like now if a download needs to be retried, hopefully this is clearer @selipot:

Screenshot from 2024-04-11 00-25-35

kevinsantana11:
@selipot please review when you get a chance. This looks good to me

PR #391, opened 2024-04-10
🔧 python 3.10 upgrade

### LINKED ISSUE #335: python 3.10 upgrade
Going from 3.9 to 3.10, there were significant convenience updates to Python's typing syntax:

* Can use built-in types (`list`, `dict`, `tuple`...) for type hints rather than importing them from `typing` (`List`, `Dict`, `Tuple`)
* Using `Optional` not necessary
* `|` instead of union

For example, in Python 3.9 (our current code)

```python
from typing import Tuple, List, Optional, Union
import numpy as np
import xarray as xr

def f(
    x: Union[List, np.array, xr.DataArray],
    n: Optional[int] = 0
) -> Tuple[int, float]:
    ... 
```

becomes

```python
import numpy as np
import xarray as xr

def f(
    x: list | np.array | xr.DataArray,
    n: int = 0
) -> tuple[int, float]:
    ... 
```

The advantages are developer convenience, more concise syntax, and requiring users to use a more recent Python version.

A disadvantage is requiring users to use a more recent Python version.

kevinsantana11:
upgrade to python 3.10. Many of the typing changes arose from using the `pyupgrade --py310-plus` upgrade command. fixes #335

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,7 @@ jobs:
       fail-fast: false
       matrix:
         os: ["ubuntu-latest", "macos-latest", "windows-latest"]
-        python-version: ["3.9", "3.10", "3.11", "3.12"]
+        python-version: ["3.10", "3.11", "3.12"]
     steps:
       - uses: actions/checkout@v4
       - uses: mamba-org/setup-micromamba@v1
@@ -36,7 +36,7 @@ jobs:
         run: |
               coverage xml
       - name: Upload coverage report to Codecov
-        if: matrix.os == 'ubuntu-latest' && matrix.python-version == '3.9'
+        if: matrix.os == 'ubuntu-latest' && matrix.python-version == '3.10'
         uses: codecov/codecov-action@v3
         with:
           file: ./coverage.xml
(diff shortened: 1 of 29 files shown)


philippemiron:
Looks good!

I found one more thing to modify here:

https://github.com/Cloud-Drift/clouddrift/blob/main/.github/workflows/ci.yml#L22

kevinsantana11:
@philippemiron I see there's another PR with changes involving micro-mamba running our tests on 3.12 instead of 3.9. I pulled in one of the commits from that branch so that those changes can all be committed together unless you think its better to commit these separately

philippemiron:
Fine either way but now it's already in this branch so I guess you can close the other one.

philippemiron:
But I think in general we should not merge commits from others PR.

- lose attributions
- lose track of what has been changed since we do squash and merge.

kevinsantana11:
> I think in general we should not merge commits from others PR.
> 
>     * lose attributions for the history
> 
>     * lose attributions for the release
> 
>     * lose track of what has been changed since we do squash and merge
> 
> 
> I don't really care since I've collaborated to the project already, but that could different for a new collaborator for example.

That's fair, should I remove the previous commit? Unsure if maybe we need to merge the #392 first and then this PR. If so I can create a commit reverting it.

philippemiron:
I think if we merge the other one first, we won't have to do anything here.

PR #412, opened 2024-04-14
🔧  show progress for downloads configurable for testing

kevinsantana11:
I added a global var that makes it easy to configure whether any output progress is shown atleast for downloads. we will still see progress outputs for filling the ragged array. To also silence these I can do something similar but wanted to get some feedback on this approach since we should be careful with any usage/relience on global vars.

### DIFF
--- a/clouddrift/adapters/utils.py
+++ b/clouddrift/adapters/utils.py
@@ -17,7 +17,7 @@ from tenacity import (
 )
 from tqdm import tqdm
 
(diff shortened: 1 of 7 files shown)


philippemiron:
Looks good. Could we include it in raggedarray.py

- number_of_observations
- allocate

also?

kevinsantana11:
> Looks good. Could we include it in raggedarray.py
> 
>     * number_of_observations
> 
>     * allocate
> 
> 
> also?

output's alot better now :D

PR #414, opened 2024-04-16
🔧 use pytest for test execution

kevinsantana11:
this sort of relates to: #412 where the tests would display tons of text during test execution particularly when running in github actions. This would make debugging tests kind of tedious as you cant jump to the exception and having to scroll down really far to see the error message is kind of a pain.

While we've now reduced the noise it has  become a bit difficult to understand the progress of the test suite execution. 

To address this I thought we could leverage [pytest](https://docs.pytest.org/en/8.0.x/contents.html#further-topics) as sort of a drop in replacement for running tests instead of using `unittest`. 

Some benefits that we get from leveraging this tool is that the test execution progress is clearly shown, and the libraries cli interface provides some rich features like the `-lf` option that runs only the last failed tests or `-nf` option that runs new tests first.


kevinsantana11:
Check out some of the unittests and doctests to see what the output looks like

philippemiron:
I wish I had started the test in `pytest` directly instead of using `unittest`..

kevinsantana11:
> I wish I had started the tests in `pytest` directly instead of using `unittest`..
> 
> At least running them with pytest now, we get better looking logs.

I really like the setup we have now. We use unittest for organizing and mocking in our tests and then we leverage `pytest` as the runner improving output and error reasoning.

Going to merge this in.

philippemiron:
I agree it's nice. But I don't know if you used pytest before? I'm using it at work and in the end it's much simpler to design test, and you don't have those annoying unittest classes to deal with.

PR #418, opened 2024-04-18
🔧 fix deprecation warnings (version, pandas read_csv)

philippemiron:
Very minor things that I saw from the new cleaned logs 🙌.


philippemiron:
Also, is it possible that the rust action only runs the `lint` step and not also `format`? I ran it manually and it changed a few files (as you can see from this PR..).

kevinsantana11:
> Also, is it possible that the rust action only runs the `lint` step and not also `format`? I ran it manually and it changed a few files (as you can see from this PR..).

hmm unsure I follow but I think we can merge this in and continue the conversation in an separate issue.

PR #424, opened 2024-04-28
🔧 add workflow to paper branch

kevinsantana11:
paper branch also needs the workflow inorder to run


philippemiron:
Was this needed? Seems like it already ran fine.

PR #429, opened 2024-04-30
doi fix in paper.bib

selipot:
removed doi.org prefix in two doi fields


selipot:
Maybe we don't want this PR to be merged

PR #430, opened 2024-05-01
🐛 label reference

philippemiron:


philippemiron:
I think this workflow is not working correctly, it's only fetching the branch so recompiling was is there already..

philippemiron:
Ready to 🚢.

selipot (approved):

PR #454, opened 2024-06-06
🔎 add default url and path to download function

philippemiron:
This will make it easier to use the `download` functions by itself. I was trying to fix the previous Notebook examples and didn't want to have to set a URL and a PATH by default.


selipot:
I think so, sure

PR #457, opened 2024-06-10
🔎 fix docs: gdp usage example

### LINKED ISSUE #449: 🔎 The example in the documentation is broken
I encountered errors when running the code from the documentation here: https://clouddrift.org/usage.html.
There's a typo in the variable name (id in the dataset is lowercase, but it's retrieved as `.ID`).
There might be other errors as well.

In my experience, keeping a static documentation in sync with the code can be very challenging.
I suggest either running the code and producing the output when you compile the documentation or using doctest to ensure the hardcoded outputs match the current code.

I'll give it another go and provide feedback on the content once the usage code is fixed.

JOSS: https://github.com/openjournals/joss-reviews/issues/6742

philippemiron:
Fix #449. Note that we will be integrating slowly doctest to spot those errors in the future.


philippemiron:
Will rebase to the main.

PR #458, opened 2024-06-10
 🔎 fix usage documentation

### LINKED ISSUE #449: 🔎 The example in the documentation is broken

philippemiron:
Fix #449. I agree with the reviewer that examples like this are prone to get out of sync with the library. Hopefully, this is something that could be addressed with `doctest`.


kevinsantana11:
We can probably use ellipses in place of fixed values for things that are subject to commonly change like the size of variables (in bytes)

philippemiron:
Yeah, not sure what is the best strategy here to be honest. When we do `ds` and it's an `xr.Dataset`, it will completely change every time there are new files.

kevinsantana11 (approved):
go ahead and merge when you can

PR #479, opened 2024-06-24
🔍 reference latest DOI

kevinsantana11:
Instead of referencing the DOI related to a release of clouddrift we should use the latest DOI which references the latest release of clouddrift, whatever version number that may be.


selipot (approved):
lgtm please merge.

PR #489, opened 2024-07-02
🐛 fix the andro datasets

philippemiron:
There is a new "Quality controlled" datasets available (https://www.seanoe.org/data/00360/47077/) and the Data Release from 2022 is not available anymore. So we were querying an empty page.


philippemiron:
Note that I forgot to update the docstring.

kevinsantana11:
This change looks good to me. Only thing I can see that would also help us catch any future changes here with our CI would be to add some assertion in the test for the andro dataset function in regards to the data contained (maybe we can check variable lengths > 0). I can add this in tomorrow just to bolster our testing.

philippemiron:
I agree adding some type of test could be good. But the dataset was return with garbage with the other url.. so not sure if just len(vars) > 0 would be enough.

philippemiron:
Let's merge and release a new version?

selipot:
Sure

PR #491, opened 2024-07-15
🔎 type comparison with isinstance

philippemiron:
Small refactoring to use `isinstance()` instead of `type() is`.


kevinsantana11 (approved):
looks good to me

PR #494, opened 2024-07-19
🐛 fix locationtype bug in dataset

### LINKED ISSUE #492: 🐛 all `location_type` are  `True` for gdp1h 2.01
```
import clouddrift as cd
ds1 = cd.datasets.gdp1h()
 all(ds1["location_type"] == True)
```
incorrectly return `True` yet
```
ds0 = xr.open_dataset('https://noaa-oar-hourly-gdp-pds.s3.amazonaws.com/latest/gdp_v2.00.zarr',engine='zarr')
all(ds1["location_type"] == True)
```
returns `False` as expected.

This means that the `location_type variable` appears to be wrong in 2.01. This needs fixing asap.

kevinsantana11:
fixes #492, where the location type was found to always be `true`

### DIFF
--- a/clouddrift/adapters/gdp1h.py
+++ b/clouddrift/adapters/gdp1h.py
@@ -20,7 +20,7 @@ import clouddrift.adapters.gdp as gdp
 from clouddrift.adapters.utils import download_with_progress
 from clouddrift.raggedarray import RaggedArray
 
-GDP_VERSION = "2.01"
+GDP_VERSION = "2.01.1"
 
 
 GDP_DATA_URL = "https://www.aoml.noaa.gov/ftp/pub/phod/buoydata/hourly_product/v2.01"
@@ -212,7 +212,7 @@ def preprocess(index: int, **kwargs) -> xr.Dataset:
     # convert attributes to variable
     ds["location_type"] = (
         ("traj"),
-        [False if ds.get("location_type") == "Argos" else True],
+        [False if ds.attrs.get("location_type") == "Argos" else True],
     )  # 0 for Argos, 1 for GPS
     ds["DeployingShip"] = (("traj"), gdp.cut_str(ds.DeployingShip, 20))
     ds["DeploymentStatus"] = (
--- a/clouddrift/adapters/gdp1h.py
+++ b/clouddrift/adapters/gdp1h.py
@@ -139,6 +139,7 @@ def preprocess(index: int, **kwargs) -> xr.Dataset:
         fp,
         decode_times=False,
         decode_coords=False,
+        engine = "netcdf4",
     )
 
     # parse the date with custom function
--- a/clouddrift/adapters/gdp1h.py
+++ b/clouddrift/adapters/gdp1h.py
@@ -238,8 +238,8 @@ def preprocess(index: int, **kwargs) -> xr.Dataset:
         ("traj"),
         gdp.cut_str(ds.attrs.get("PurchaserFunding", ""), 20),
     )
-    ds["SensorUpgrade"] = (("traj"), gdp.cut_str(ds.attrs.get(".SensorUpgrade"), 20))
-    ds["Transmissions"] = (("traj"), gdp.cut_str(ds.attrs.get("Transmissions"), 20))
+    ds["SensorUpgrade"] = (("traj"), gdp.cut_str(ds.attrs.get("SensorUpgrade", ""), 20))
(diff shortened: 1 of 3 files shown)


philippemiron:
I'm not sure how you fix it, this PR shows no changes. I would suggest we use:
```
ds.attrs.get("attributes")
```
for all of the attributes conversion. That way, if an attribute is not available, it is automatically converted to None.

kevinsantana11:
@philippemiron sorry for the confusion, I created the PR as a place holder for now while I debug the issue. I'm trying to understand where this attribute originates and it seems so far that it is available in some of the drifter data files but not all. I'm currently analyzing the data to see which drifter files have the attribute and which ones don't. I will also include the changes you mentioned in regards to the conversion.

philippemiron:
I think the main issue is that we are doing `ds.get("location_type")` and it should be `ds.attrs.get("location_type")`.

selipot:
Trying now to regenerate the hourly dataset, bumping version to 2.01.1 but lots of failed requests to download

selipot:
The same bug fix cloud be implemented for `clouddrift.adapters.gdp6h()` yet it looks like the attributes of the 6 hourly netcdf files do *not* contain the `location_type` variable. This might be deduced from the string `BuoyTypeSensorArray` which should contain the letter `G` but it is not foolproof. I think we should discuss with the DAC.

selipot:
I get the following error when filling out the array after download:
```

File /opt/homebrew/Caskroom/mambaforge/base/envs/clouddrift/lib/python3.12/site-packages/xarray/backends/api.py:552, in open_dataset(filename_or_obj, engine, chunks, cache, decode_cf, mask_and_scale, decode_times, decode_timedelta, use_cftime, concat_characters, decode_coords, drop_variables, inline_array, chunked_array_type, from_array_kwargs, backend_kwargs, **kwargs)
    549     kwargs.update(backend_kwargs)
    551 if engine is None:
--> 552     engine = plugins.guess_engine(filename_or_obj)
    554 if from_array_kwargs is None:
    555     from_array_kwargs = {}

File /opt/homebrew/Caskroom/mambaforge/base/envs/clouddrift/lib/python3.12/site-packages/xarray/backends/plugins.py:197, in guess_engine(store_spec)
    189 else:
    190     error_msg = (
    191         "found the following matches with the input file in xarray's IO "
    192         f"backends: {compatible_engines}. But their dependencies may not be installed, see:\n"
    193         "https://docs.xarray.dev/en/stable/user-guide/io.html \n"
    194         "https://docs.xarray.dev/en/stable/getting-started-guide/installing.html"
    195     )
--> 197 raise ValueError(error_msg)

ValueError: did not find a match in any of xarray's currently installed IO backends ['netcdf4', 'h5netcdf', 'scipy', 'zarr']. Consider explicitly selecting one of the installed engines via the ``engine`` parameter, or installing additional IO dependencies, see:
https://docs.xarray.dev/en/stable/getting-started-guide/installing.html
https://docs.xarray.dev/en/stable/user-guide/io.html
```

selipot:
I finally managed to make things work by adding `engine="netcdf"` but I am not sure it is the real reason why it works now

kevinsantana11:
@selipot Philippe pushed the change that fixes this and I added a test ensuring the values are correct. The engine issue is separate.

selipot:
> @selipot Philippe pushed the change that fixes this and I added a test ensuring the values are correct. The engine issue is separate.

Really, what was it?

kevinsantana11:
The issue here was related to the following line: 
```
[False if ds.get("location_type") == "Argos" else True],
```
[source](https://github.com/Cloud-Drift/clouddrift/blob/main/clouddrift/adapters/gdp1h.py#L215)

`ds.get("location_type")` would return None for each drifter since the actual value for the `location_type` was correctly retrieved using `ds.attrs.get("location_type")` [as mentioned by Philippe](https://github.com/Cloud-Drift/clouddrift/pull/494#issuecomment-2240819428).

This is likely a result of some xarray API change since this code hasn't changed for a while and I believe the previous version of the aggregate dataset (v2.00) was generated successfully with the correct `location_type` values.

selipot:
Yes, I pushed that commit following @philippemiron 's comment. But could not complete the re-generation of the entire array until this morning

kevinsantana11:
@selipot got it, I'm also working on another PR that will surround all of our dependency interfacing code (requests to external servers) with some robust retry logic.

PR #496, opened 2024-07-21
🧹 Update random number generator in gdp1h and gdp6h adapters

selipot:
The code changes update the random number generator in the `gdp1h.py` and `gdp6h.py` adapters. The previous usage of `np.random.RandomState` has been replaced with `np.random.Generator(np.random.MT19937)`. This change ensures consistent and reproducible random number generation.

### DIFF
--- a/clouddrift/adapters/gdp1h.py
+++ b/clouddrift/adapters/gdp1h.py
@@ -103,7 +103,7 @@ def download(
                 f"Retrieving all listed trajectories because {n_random_id} is larger than the {len(filelist)} listed trajectories."
             )
         else:
-            rng = np.random.RandomState(42)
+            rng = np.random.Generator(np.random.MT19937(42))
             filelist = sorted(rng.choice(filelist, n_random_id, replace=False))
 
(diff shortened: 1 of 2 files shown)


kevinsantana11:
I think this fits with the :chore kind of commit but prefer if we use an emoji (like: 🧹) and update our CONTRIBUTING.md accordingly.

philippemiron:
It looks like this is the new suggested way to do it. But I don't think `np.random.RandomState(42)` was incorrect.

Using:
```
python -c "import numpy as np; rng = np.random.RandomState(42); print(rng.random(10))"
```
or 
```
python -c "import numpy as np; rng = np.random.Generator(np.random.MT19937(42)); print(rng.random(10))"
```
both returns the same _random_ numbers.

selipot:
I agree it returns the same but i read that RandomState will be deprecated. Not a big deal we don't have to change.

kevinsantana11 (approved):
I think its good to get out ahead of any deprecation efforts, lets merge this in

PR #497, opened 2024-07-21
🧹 Update type comparison in ragged.py

selipot:
The code changes in ragged.py update the type comparison for the `tolerance` variable. Instead of using the `type()` function, it now uses the `isinstance()` function to check if `tolerance` is an instance of `pd.Timedelta`. This change improves the accuracy and reliability of the comparison.

Note: This commit message follows the established convention of using the "chore" type for non-functional changes that don't affect the code's behavior.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -61,6 +61,7 @@ Common emojis to use are as follow:
 ❗ Deprecation of a feature
 ⛔ Removal of feature
 🐛 Bugfix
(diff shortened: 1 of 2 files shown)


philippemiron:
Fyi Refactoring is at two spots.

kevinsantana11:
@philippemiron thanks for the heads up

kevinsantana11 (approved):
looks good to me, merging

PR #502, opened 2024-07-24
⭐️ new functions `index_to_row` and `rowsize_to_rowvector`

### LINKED ISSUE #501: ⭐ derive new function `index_to_row`
**Describe the solution you'd like**
Define a new function which considering a ragged array xarray dataset `ds` with row size variable `rowsize` would return the row index associated with a observation index
`row_index = index_to_row(index,ds["rowsize"])`

**Additional context**
Sort of inverting `rowsize_to_index` but not really

selipot:
New functions `index_to_row` and `rowsize_to_rowvector` to easily find the rows of observations. This fixes #501.

I think these are useful if one tries to find the row of a given observation. As an example, let's say you notice a weird observation and you want to identify the trajectory (or row) associated with that observation. 

I am open to name changes and algorithm changes. Please suggest!

### DIFF
--- a/clouddrift/ragged.py
+++ b/clouddrift/ragged.py
@@ -855,14 +855,14 @@ def unpack(
 
 
 def index_to_row(
-    index: int | list[int],
+    index: int | list[int] | np.ndarray,
     rowsize: list[int] | np.ndarray | xr.DataArray,
 ) -> list:
     """Obtain a list of row indices from a list of observation indices of a ragged array.
 
     Parameters
     ----------
-    index : int or list
+    index : int or list or np.ndarray
         A integer observation index or a list of observation indices of a ragged array.
     rowsize : list or np.ndarray or xr.DataArray
         A sequence of row sizes of a ragged array.
@@ -900,7 +900,7 @@ def index_to_row(
     rowsize_index = rowsize_to_index(rowsize)
 
     # test that no index is out of bounds
-    if any((i < rowsize_index[0]) | (i >= rowsize_index[-1]) for i in index_list):
+    if any([(i < rowsize_index[0]) | (i >= rowsize_index[-1]) for i in index_list]):
         raise ValueError("Input index out of bounds based on input rowsize")
 
     return (np.searchsorted(rowsize_index, index_list, side="right") - 1).tolist()
--- a/clouddrift/ragged.py
+++ b/clouddrift/ragged.py
@@ -911,27 +911,33 @@ def index_to_row(
     To obtain the row index of observation 5 within a ragged array of three consecutive
     rows of sizes 2, 4, and 3:
     >>> index_to_row(5, [2, 4, 3])
-    1
+    [1]
 
     To obtain the row indices of observations 0, 2, and 4 within a ragged array of three consecutive
     rows of sizes 2, 4, and 3:
     >>> index_to_row([0, 2, 4], [2, 4, 3])
-    [0, 1, 2]
+    [0, 1, 1]
 
     """
     # if index is an integer, convert it to a list
-    if isinstance(index, int):
+    if isinstance(index, int) | isinstance(index, float):
         index_list = [index]
     else:
         index_list = index
 
-    # if index is not a list of integers, raise an error
-    if not all(isinstance(i, int) for i in index_list):
+    # if index is not a list of integers or integer-likes, raise an error
+    if not all(
+        isinstance(i, int) or (isinstance(i, float) and i % 1 == 0) for i in index_list
+    ):
         raise ValueError("The index must be an integer or a list of integers.")
 
-    rowvector_flattened = rowsize_to_rowvector(rowsize)
+    rowsize_index = rowsize_to_index(rowsize)
 
-    return [rowvector_flattened[i] for i in index_list]
+    # test that no index is out of bounds
+    if any((i < rowsize_index[0]) | (i >= rowsize_index[-1]) for i in index_list):
+        raise ValueError("Input index out of bounds based on input rowsize")
+
+    return (np.searchsorted(rowsize_index, index_list, side="right") - 1).tolist()
 
 
 def _mask_var(
--- a/clouddrift/ragged.py
+++ b/clouddrift/ragged.py
@@ -854,40 +854,6 @@ def unpack(
     return [unpacked[i] for i in rows]
 
 
-def rowsize_to_rowvector(
-    rowsize: list[int] | np.ndarray | xr.DataArray,
-) -> list:
-    """Obtain a list of repeated row indices from a list of row sizes of a ragged array.
-
-    Parameters
-    ----------
-    rowsize : list or np.ndarray or xr.DataArray
-        A sequence of row sizes greater than zero.
-
-    Returns
-    -------
-    list
-        A list of repeated row indices.
-
-    Examples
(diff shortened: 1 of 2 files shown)


philippemiron:
Hi Shane,

I think this could be a bit simpler. I have tested quickly, and I think this should work:

```
def index_to_row(
    index: int | list[int], rowsize: list[int] | np.ndarray | xr.DataArray
):
    if isinstance(index, int):
        index_list = [index]
    else:
        index_list = index

    rowsize_index = rowsize_to_index(rowsize)

    return [np.searchsorted(rowsize_index, i, side="right") - 1 for i in index]
```

philippemiron:
Actually it doesn't even need the loop.

```
def index_to_row(
    index: int | list[int], rowsize: list[int] | np.ndarray | xr.DataArray
):
    rowsize_index = rowsize_to_index(rowsize)
    return np.searchsorted(rowsize_index, index, side="right") - 1
```

selipot:
@philippemiron  I think the other function `rowsize_to_rowvector` is still useful, what do you think? Further this could be used to transform a variable with dimension `row` to the corresponding variable with dimension `obs`.

philippemiron:
You can use `np.repeat` for this if needed:

```
rowsize = [2, 5, 3]
index = list(range(3))
np.repeat(index, rowsize)
```

selipot:
> You can use `np.repeat` for this if needed:
> 
> ```
> rowsize = [2, 5, 3]
> index = list(range(3))
> np.repeat(index, rowsize)
> ```

So maybe not worth a function ...

philippemiron:
That's also what I was thinking haha.

clouddrift/ragged.py
```diff
@@ -854,6 +854,58 @@ def unpack(
     return [unpacked[i] for i in rows]
 
 
+def index_to_row(
+    index: int | list[int],
```
philippemiron, line 858:
```suggestion
    index: int | list[int] | np.ndarray,
```

clouddrift/ragged.py
```diff
@@ -854,6 +854,58 @@ def unpack(
...
+    rowsize: list[int] | np.ndarray | xr.DataArray,
+) -> list:
+    """Obtain a list of row indices from a list of observation indices of a ragged array.
+
+    Parameters
+    ----------
+    index : int or list
```
philippemiron, line 865:
```suggestion
    index : int or list or np.ndarray
```

clouddrift/ragged.py
```diff
@@ -854,6 +854,58 @@ def unpack(
...
+    # if index is not a list of integers or integer-likes, raise an error
+    if not all(
+        isinstance(i, int) or (isinstance(i, float) and i % 1 == 0) for i in index_list
+    ):
+        raise ValueError("The index must be an integer or a list of integers.")
+
+    rowsize_index = rowsize_to_index(rowsize)
+
+    # test that no index is out of bounds
+    if any((i < rowsize_index[0]) | (i >= rowsize_index[-1]) for i in index_list):
```
philippemiron, line 903: I would think you need a `[]` here.. but I guess it's working?

```suggestion
    if any([(i < rowsize_index[0]) | (i >= rowsize_index[-1]) for i in index_list]):
```

selipot:
good to merge?

kevinsantana11:
Still need to review the code, I will take a look at this around noon

clouddrift/ragged.py
```diff
@@ -854,6 +854,58 @@ def unpack(
...
+    [1]
+
+    To obtain the row indices of observations 0, 2, and 4 within a ragged array of three consecutive
+    rows of sizes 2, 4, and 3:
+    >>> index_to_row([0, 2, 4], [2, 4, 3])
+    [0, 1, 1]
+
+    """
+    # if index is an integer, convert it to a list
+    if isinstance(index, int) | isinstance(index, float):
```
kevinsantana11, line 889: We likely want this check to only appy to values that are of type int. The right side of the or operator should be removed.

kevinsantana11:
We should maybe use a different name here for the function, like `get_row_indexes` or `row_indexes` with parameters like `obs_index` and `rowsize`

clouddrift/ragged.py
```diff
@@ -854,6 +854,58 @@ def unpack(
...
+        index_list = [index]
+    else:
+        index_list = index
+
+    # if index is not a list of integers or integer-likes, raise an error
+    if not all(
+        isinstance(i, int) or (isinstance(i, float) and i % 1 == 0) for i in index_list
```
kevinsantana11, line 896: The float type check should be removed along with the divisibility by 1 check.

clouddrift/ragged.py
kevinsantana11, line 889: Have we tested handling a DataArray or should we include a case here that would handle grabbing the array from the `.data` accessor?

selipot: can we allow to pass a float pretending to be an integer?

clouddrift/ragged.py
selipot: same reply as above

clouddrift/ragged.py
kevinsantana11: not sure I follow the question, is this regarding the static type checking or whether we can pass a float like 3.18 but instead interpret it as the integer 318?

selipot: pass 5.0 for 5

selipot:
> We should maybe use a different name here for the function, like `get_row_indexes` or `row_indexes` with parameters like `obs_index` and `rowsize`

When we create a new datasets we have consistently used `obs` and `traj` for dimensions when a ragged array Lagrangian dataset is presented as an xarray dataset, yet we decided to keep the default row dimension name as `row` instead of `traj`. Keeping with the naming convention `A_to_B` then it should be `obs_to_row` or `obs_index_to_row`.

clouddrift/ragged.py
kevinsantana11: I don't think we should allow the use of floats if the intent of the function is to find the row_idx from the obs_idx where we know obs_idx is an integer. 

If the intent is instead to pass in a data var say like: `sst` and find the associated row_idx of the sst observation (determining the sst observations obs_index first but encapsulating the logic) then it makes sense to support this type but we would also need to update the logic.

I think my main confusion here is figuring out the case where an obs_idx should be a float rather than an integer.

selipot: I am not sure, but I would easily see myself generating a vector of indices which contains float instead of integer

clouddrift/ragged.py
```diff
@@ -854,6 +854,62 @@ def unpack(
     return [unpacked[i] for i in rows]
 
 
+def obs_index_to_row(
+    index: int | list[int] | np.ndarray,
```
philippemiron, line 858: Missing xr.DataArray here

philippemiron (approved):

kevinsantana11 (approved):

kevinsantana11:
@selipot once you fix the ruff issue, please merge

kevinsantana11:
Hi @selipot, could we align this PR with what's documented in the [contributing docs](https://github.com/Cloud-Drift/clouddrift/blob/main/CONTRIBUTING.md#git-practices) before merging?

PR #504, opened 2024-08-04
🧹 fix wrong None arguments

philippemiron:
Hi @kevinsantana11 / @selipot, I tried to fix `mypy` for all modules and failed (to say the least). I did this PR as a first step, we have a few arguments that were:
```
x: int | None = 0
```
which should simply be:
```
x: int = 0
```

To make typing a bit more easy for us, I think we should discuss a strategy. As of right now, we have a lot of functions that try to accept all types:
```
list | np.ndarray | pd.Series | xr.DataArray
```
but this is not constant across `clouddrift`, so for example, another function will have:
```
np.ndarray[float] | pd.Series | xr.DataArray
```
so if we pass a list somewhere `mypy` will complain.

Similarly, we are accepting `list` at a few places alongside other `array_like` structures and do operations like (if `x` is `array_like`):
```
x / CONSTANT
x.shape
x.size
```
which are not valid if `isinstance(x, list)`.

My main questions are:
- do we really want to support all the types?
- what do other libraries do in that situation? Or someone knows what's the best practices here?

One solution could be to defined a `clouddrift_array_like_type` once, and then reused it.. Something we do in `pairs.py` with:
```
array_like = Union[list, np.ndarray, pd.Series, xr.DataArray]
```

Maybe we can copy this to a discussion?

### DIFF
--- a/clouddrift/kinematics.py
+++ b/clouddrift/kinematics.py
@@ -65,8 +65,8 @@ def inertial_oscillation_from_position(
     latitude: np.ndarray,
     relative_bandwidth: float | None = None,
     wavelet_duration: float | None = None,
-    time_step: float | None = 3600.0,
-    relative_vorticity: float | np.ndarray | None = 0.0,
+    time_step: float = 3600.0,
+    relative_vorticity: float | np.ndarray = 0.0,
 ) -> np.ndarray:
     """Extract inertial oscillations from consecutive geographical positions.
 
@@ -324,9 +324,9 @@ def position_from_velocity(
     time: np.ndarray,
     x_origin: float,
     y_origin: float,
-    coord_system: str | None = "spherical",
-    integration_scheme: str | None = "forward",
-    time_axis: int | None = -1,
+    coord_system: str = "spherical",
+    integration_scheme: str = "forward",
+    time_axis: int = -1,
 ) -> tuple[np.ndarray, np.ndarray]:
     """Compute positions from arrays of velocities and time and a pair of origin
     coordinates.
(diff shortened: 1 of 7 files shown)


philippemiron:
We could define a clouddrift_type.py module with:

```
from typing import Union
import numpy as np
import pandas as pd
import xarray as xr

ArrayLike = Union[list, np.ndarray, pd.Series, xr.DataArray]

# If we want to support list everywhere
# also works with float, which are converted to `np.array(42)` and gives access to `.shape` and `.size`
def ensure_array(data: Union[list, np.ndarray, pd.Series, xr.DataArray]) -> np.ndarray:
    if isinstance(data, list):
        return np.array(data)
    elif isinstance(data, (np.ndarray, pd.Series, xr.DataArray)):
        return data
    else:
        raise TypeError(f"Unsupported data type: {type(data)}")
```

Then in functions, we could:
```
def method_a(x: ArrayLike) -> ArrayLike:
    data = ensure_array(data)
    ...
    ...
    return data   
```

kevinsantana11:
> We could define a clouddrift_type.py module with:
> 
> ```
> from typing import Union
> import numpy as np
> import pandas as pd
> import xarray as xr
> 
> ArrayLike = Union[list, np.ndarray, pd.Series, xr.DataArray]
> 
> # If we want to support list everywhere
> # also works with float, which are converted to `np.array(42)` and gives access to `.shape` and `.size`
> def ensure_array(data: ArrayLike) -> np.ndarray:
>     if isinstance(data, list):
>         return np.array(data)
>     elif isinstance(data, (np.ndarray, pd.Series, xr.DataArray)):
>         return data
>     else:
>         raise TypeError(f"Unsupported data type: {type(data)}")
> ```
> 
> Then in functions, we could:
> 
> ```
> def method_a(x: ArrayLike) -> ArrayLike:
>     data = ensure_array(data)
>     ...
>     ...
>     return data   
> ```

I think this is a great idea, it could simplify alot of our type annotations

kevinsantana11 (approved):
These changes look good to me. I went ahead and created the separate discussion to talk about standardizing out array like type support.

https://github.com/Cloud-Drift/clouddrift/discussions/507

PR #510, opened 2024-08-15
🔧 fix typing issues from checking untyped defs, fixes #509

### LINKED ISSUE #509: Fix untyped definitions and enable option to enforce typing throughout the library

kevinsantana11:
One thing we should enable is the feature to disallow untyped definitions. This option was previously turned off but I've enabled it and fixed all of the issues listed (100+)

I still want to add some more changes here regarding standardizing on the array types we want to support as well as leveraging generics more with bounds to achieve some of the more dynamic behavior of the library.

This also follows the guidance from mypy to turn these feature on as soon as possible: [article](https://mypy.readthedocs.io/en/stable/existing_code.html#introduce-stricter-options)

### DIFF
--- a/clouddrift/ragged.py
+++ b/clouddrift/ragged.py
@@ -460,8 +460,8 @@ def rowsize_to_index(rowsize: cd_typing.ArrayTypes) -> np.ndarray:
 def segment(
     x: cd_typing.ArrayTypes,
     tolerance: float | np.timedelta64 | timedelta | pd.Timedelta,
-    rowsize: np_typing.NDArray[np.int_] | None = None,
-) -> np_typing.NDArray[np.int_]:
+    rowsize: np_typing.NDArray[np.int64] | None = None,
+) -> np_typing.NDArray[np.int64]:
     """Divide an array into segments based on a tolerance value.
 
     Parameters
@@ -797,7 +797,7 @@ def subset(
 def unpack(
     ragged_array: cd_typing.ArrayTypes,
     rowsize: cd_typing.ArrayTypes,
-    rows: int | np.int_ | Iterable[int] | None = None,
+    rows: int | np.int64 | Iterable[int] | None = None,
     axis: int = 0,
 ) -> list[np_typing.NDArray[typing.Any]]:
     """Unpack a ragged array into a list of regular arrays.
--- a/clouddrift/ragged.py
+++ b/clouddrift/ragged.py
@@ -6,6 +6,7 @@ import warnings
 from collections.abc import Callable, Iterable
 from concurrent import futures
 from datetime import timedelta
+from typing import Any
 
 import numpy as np
 import pandas as pd
@@ -16,12 +17,12 @@ def apply_ragged(
     func: callable,
     arrays: list[np.ndarray | xr.DataArray] | np.ndarray | xr.DataArray,
     rowsize: list[int] | np.ndarray[int] | xr.DataArray,
-    *args: tuple,
+    *args: Any,
     rows: int | Iterable[int] = None,
     axis: int = 0,
     executor: futures.Executor = futures.ThreadPoolExecutor(max_workers=None),
-    **kwargs: dict,
-) -> tuple[np.ndarray] | np.ndarray:
+    **kwargs: Any,
+) -> tuple[np.ndarray, np.ndarray] | np.ndarray:
     """Apply a function to a ragged array.
 
     The function ``func`` will be applied to each contiguous row of ``arrays`` as
@@ -450,9 +451,9 @@ def rowsize_to_index(rowsize: list | np.ndarray | xr.DataArray) -> np.ndarray:
 
 
 def segment(
-    x: np.ndarray,
+    x: list | np.ndarray | xr.DataArray | pd.Series,
     tolerance: float | np.timedelta64 | timedelta | pd.Timedelta,
-    rowsize: np.ndarray[int] = None,
+    rowsize: np.ndarray[int] | None = None,
 ) -> np.ndarray[int]:
     """Divide an array into segments based on a tolerance value.
 
@@ -789,7 +790,7 @@ def subset(
 def unpack(
     ragged_array: np.ndarray,
     rowsize: np.ndarray[int],
-    rows: int | Iterable[int] = None,
+    rows: int | np.int_ | Iterable[int] | None = None,
     axis: int = 0,
 ) -> list[np.ndarray]:
     """Unpack a ragged array into a list of regular arrays.
--- a/clouddrift/ragged.py
+++ b/clouddrift/ragged.py
@@ -9,11 +9,11 @@ from concurrent import futures
 from datetime import timedelta
 
 import numpy as np
-import numpy.typing as np_typing
 import pandas as pd
 import xarray as xr
+from numpy.typing import NDArray
 
-import clouddrift.typing as cd_typing
+from clouddrift.typing import ArrayTypes
 
 _ArrayOutput = typing.TypeVar(
     "_ArrayOutput", bound=tuple[np.ndarray, np.ndarray] | np.ndarray
@@ -22,8 +22,8 @@ _ArrayOutput = typing.TypeVar(
 
 def apply_ragged(
     func: Callable[..., _ArrayOutput],
-    arrays: cd_typing.ArrayTypes,
-    rowsize: cd_typing.ArrayTypes,
+    arrays: ArrayTypes,
+    rowsize: ArrayTypes,
     *args: typing.Any,
     rows: int | Iterable[int] = None,
     axis: int = 0,
@@ -431,7 +431,7 @@ def regular_to_ragged(
     return array[valid], np.sum(valid, axis=1)
 
 
-def rowsize_to_index(rowsize: cd_typing.ArrayTypes) -> np.ndarray:
+def rowsize_to_index(rowsize: ArrayTypes) -> np.ndarray:
     """Convert a list of row sizes to a list of indices.
 
     This function is typically used to obtain the indices of data rows organized
@@ -458,10 +458,10 @@ def rowsize_to_index(rowsize: cd_typing.ArrayTypes) -> np.ndarray:
 
 
 def segment(
-    x: cd_typing.ArrayTypes,
+    x: ArrayTypes,
     tolerance: float | np.timedelta64 | timedelta | pd.Timedelta,
-    rowsize: np_typing.NDArray[np.int64] | None = None,
-) -> np_typing.NDArray[np.int64]:
+    rowsize: NDArray[np.int64] | None = None,
+) -> NDArray[np.int64]:
     """Divide an array into segments based on a tolerance value.
 
     Parameters
@@ -795,11 +795,11 @@ def subset(
 
 
 def unpack(
-    ragged_array: cd_typing.ArrayTypes,
-    rowsize: cd_typing.ArrayTypes,
+    ragged_array: ArrayTypes,
+    rowsize: ArrayTypes,
     rows: int | np.int64 | Iterable[int] | None = None,
     axis: int = 0,
-) -> list[np_typing.NDArray[typing.Any]]:
+) -> list[NDArray[typing.Any]]:
     """Unpack a ragged array into a list of regular arrays.
 
     Unpacking a ``np.ndarray`` ragged array is about 2 orders of magnitude
--- a/clouddrift/ragged.py
+++ b/clouddrift/ragged.py
@@ -13,6 +13,8 @@ import pandas as pd
 import xarray as xr
 
 _Out = t.TypeVar("_Out", bound=tuple[np.ndarray, np.ndarray] | np.ndarray)
+
+
 def apply_ragged(
     func: Callable[..., _Out],
     arrays: list[np.ndarray | xr.DataArray] | np.ndarray | xr.DataArray,
--- a/clouddrift/ragged.py
+++ b/clouddrift/ragged.py
@@ -2,27 +2,27 @@
 Transformational and inquiry functions for ragged arrays.
 """
 
+import typing as t
 import warnings
 from collections.abc import Callable, Iterable
 from concurrent import futures
 from datetime import timedelta
-from typing import Any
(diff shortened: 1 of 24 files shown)


clouddrift/wavelet.py
```diff
@@ -344,9 +344,9 @@ def wavelet_transform(
 
 def morse_wavelet(
     length: int,
-    gamma: float,
-    beta: float,
-    radian_frequency: np.ndarray,
+    gamma: float | np.ndarray,
```
selipot, line 347: Gamma and beta parameters cannot be np.ndarray the way the function is written

clouddrift/sphere.py
```diff
@@ -660,13 +664,12 @@ def cartesian_to_spherical(
     return lon, lat
 
 
+T = TypeVar("T", bound=float | np.ndarray)
```
selipot, line 667: I am not finding this alias very readable. Why does this help?

kevinsantana11: This isn't an alias, its a generic type variable: https://mypy.readthedocs.io/en/stable/generics.html

kevinsantana11:
This code also fixes a bug surfaced by the new  type strictness.

The bug itself was an issue where the download logic would never check if the upstream source was updated or the same resource when compared to the local data-store always resulting in files being downloaded.

Our test code should have caught this but a mismatch in type checking introduced a bit of an oversight.

Increase the type strictness should prevent code like where we call functions with the wrong types.

Had this been type checked I would have updated to use the right type and realized the function would no longer check for the local modification time since it only handled buffers.

clouddrift/ragged.py
```diff
@@ -787,9 +790,9 @@ def subset(
 
 
 def unpack(
-    ragged_array: np.ndarray,
-    rowsize: np.ndarray[int],
-    rows: int | Iterable[int] = None,
+    ragged_array: np.ndarray | xr.DataArray,
+    rowsize: np.ndarray[int] | xr.DataArray,
+    rows: int | np.int_ | Iterable[int] | None = None,
```
selipot, line 795: is it necessary to have both `int` and `np.int_?`

clouddrift/signal.py
```diff
@@ -10,7 +10,7 @@ def analytic_signal(
     x: np.ndarray | xr.DataArray,
     boundary: str = "mirror",
     time_axis: int = -1,
-) -> np.ndarray | tuple[np.ndarray, np.ndarray]:
+) -> np.ndarray:
```
selipot, line 13: The output CAN be a tuple depending on the input. Why remove it?

clouddrift/sphere.py
selipot: Sorry, the whole typing thing is confusing to me

clouddrift/wavelet.py
```diff
@@ -15,22 +15,52 @@
...
 """
 
+import typing as t
+
 import numpy as np
 from scipy.special import gamma as _gamma
 from scipy.special import gammaln as _lgamma
 
 
+@t.overload
```
selipot, line 25: What is this for? @t.overload then a repeat of the definition?

clouddrift/wavelet.py
```diff
@@ -15,22 +15,52 @@
...
+) -> np.ndarray: ...
+
+
 def morse_wavelet_transform(
     x: np.ndarray,
     gamma: float,
     beta: float,
     radian_frequency: np.ndarray,
-    complex: bool = False,
+    complex: bool,
```
selipot, line 58: a boolean is False by default? Why remove the default?

clouddrift/wavelet.py
```diff
@@ -185,18 +215,9 @@ def morse_wavelet_transform(
...
-        # real case
-        wtx = wavelet_transform(x, wavelet, boundary=boundary, time_axis=time_axis)
-
-    else:
-        raise ValueError(
-            "`complex` optional argument must be boolean 'True' or 'False'"
-        )
-
-    return wtx
+    return wavelet_transform(x, wavelet, boundary=boundary, time_axis=time_axis)
```
selipot, line 220: ok i understand how it is the same logic but to me it is less clear.

clouddrift/wavelet.py
```diff
@@ -346,7 +367,7 @@ def morse_wavelet(
     length: int,
     gamma: float,
     beta: float,
-    radian_frequency: np.ndarray,
+    radian_frequency: float | np.ndarray,
```
selipot, line 370: I am pretty sure the code needs the radian frequency to be an array, not a float, even with a single element.

tests/wavelet_tests.py
```diff
@@ -22,7 +22,7 @@ def test_morse_wavelet_transform_real(self):
         length = 1023
         radian_frequency = 2 * np.pi / np.logspace(np.log10(10), np.log10(100), 50)
         x = np.random.random(length)
-        wtx = morse_wavelet_transform(x, 3, 10, radian_frequency)
+        wtx = morse_wavelet_transform(x, 3, 10, radian_frequency, False)
```
selipot, line 25: Why can't we have the complex=False a default for this function. You seem to have it changed to a complusory argument?

selipot (commented):
Here are my comments. You will see that I am confused in many places, we need to discuss.

milancurcic:
This is generally not necessary, but can be helpful for developers (e.g. if you want to require mypy to pass with certain settings on).

I didn't look at the full diff, but from the description it seems that the changes are only to the type hints. Python ignores type hints, so the code itself is unchanged in functionality.

So, two questions to ask are:

1. Are these hints helpful for the contributors?
2. Are these hints helpful for the end user?

IMO:

1. They can be in moderation, if I can understand the types. But if you develop a type system only to satisfy a strict type checking tool, e.g. mypy, it can easily escalate to a type system nightmare. There are some type definitions here that I don't understand on first look. Doesn't mean that I couldn't understand them if I studied them, but they're complex enough that they're not obvious.
2. Probably not; the users will mostly read the docstrings.

clouddrift/ragged.py
```diff
@@ -11,17 +12,19 @@
 import pandas as pd
 import xarray as xr
 
+_Out = t.TypeVar("_Out", bound=tuple[np.ndarray, np.ndarray] | np.ndarray)
```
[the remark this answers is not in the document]
kevinsantana11: I understand that the type `_Out` makes reading the function signature more difficult. I wish we could use the [direct syntax support for generics](https://docs.python.org/3/whatsnew/3.12.html#whatsnew312-pep695) that was released in python 3.12 but given were currently supporting python 3.10+ so we'll have to wait a bit before being able to leverage that syntax for generics.

The purpose of using this generic type variable is to annotate that the output type of both the `func` passed to `apply_ragged` and `apply_ragged` itself are both the same. As an example if the output type of the `func` variable is a `tuple[ndarray, ndarray]` then the output type of `apply_ragged` will also be a `tuple[ndarray, ndarray]`.

Maybe I'm misunderstanding how this function works but the output type of the function passed to apply_ragged (`func`) and `apply_ragged` itself should match right?

Regarding t.Any: I used it in places where that was the implicit type and no more specific type could be determined like np.int or something similar. Examples are type annotations such as `list` which would implicitly be interpret as `list[Any]` by the type checker. In places where I could determine a more specific type I would use that but there were cases where I couldn't and so I would just annotate it with `t.Any`.

My hope with the type annotations and stricter type checking is never to make users life harder but to try and improve the libraries robustness to change.

We could take a different approach to typing similar to numpy where they leverage  `.pyi` or [stub files](https://typing.readthedocs.io/en/latest/spec/distributing.html#stub-files) to define types like this  

A quick code search on the usage of TypeVar is different per project please see:
* [scipy usage](https://github.com/search?q=repo%3Ascipy%2Fscipy%20TypeVar&type=code)
* [numpy usage](https://github.com/search?q=repo%3Anumpy%2Fnumpy+TypeVar&type=code)
* [xarray usage](https://github.com/search?q=repo%3Apydata%2Fxarray%20TypeVar&type=code)
* [matplotlib usage](https://github.com/search?q=repo%3Amatplotlib%2Fmatplotlib%20TypeVar&type=code)
* [pandas usage](https://github.com/search?q=repo%3Apandas-dev%2Fpandas%20TypeVar&type=code)

Taking a look at the above I see moderate usage throughout the different projects. Some use it to some extent other projects have little use.

clouddrift/ragged.py
kevinsantana11: I can check on this

clouddrift/signal.py
kevinsantana11: Got it, will update it back.

clouddrift/wavelet.py
kevinsantana11: These likely fit better in a stub file but the purpose of it is two fold.
1. Document that the output type is directly dependent on the the value of the `complex` parameter
2. Helps the type checker understand the functions output type is dependent on the `complex` parameters value.

Without this, the type checker will think function calls made to `morse_wavelet_transform` can return either type (tuple or a single ndarray) and so will warn us to write code to check the types and handle the two scenarios separately.

clouddrift/wavelet.py
kevinsantana11: I can revert this change.

clouddrift/wavelet.py
kevinsantana11: Got it, I will correct this.

tests/wavelet_tests.py
kevinsantana11: I will revert to have this default put back

kevinsantana11:
@selipot this is ready for re-review

clouddrift/sphere.py
philippemiron: I read more about `TypeVar` and actually like it. If I understand correctly, it assigns the type to `T` dynamically, corrct?

philippemiron:
What's the advantage of using `from numpy.typing import NDArray` instead of `np.ndarray` ?

philippemiron:
and I would suggest also doing :

```
from clouddrift.typing import ArrayTypes
```

to remove some verbosity in the docstring.

clouddrift/sphere.py
```diff
@@ -660,13 +664,13 @@ def cartesian_to_spherical(
...
+
+
 def cartesian_to_tangentplane(
-    u: float | np.ndarray,
-    v: float | np.ndarray,
-    w: float | np.ndarray,
-    longitude: float | np.ndarray,
-    latitude: float | np.ndarray,
-) -> tuple[float, float] | tuple[np.ndarray, np.ndarray]:
+    u: T, v: T, w: T, longitude: V, latitude: V
```
philippemiron, line 672: but I'm confused then what is the puporse of the `ArrayTypes` from `clouddrift.typing` if we use `TypeVar` here.

pyproject.toml
```diff
@@ -70,6 +70,22 @@ select = ["E4", "E7", "E9", "F", "I"]
 [tool.mypy]
```
philippemiron, line 70: but we still have the section `[[tool.mypy.overrides]]` ?

tests/wavelet_tests.py
philippemiron: I think this was forgotten.

kevinsantana11:
> What's the advantage of using `from numpy.typing import NDArray` instead of `np.ndarray` ?
> 
> And if it's really required, I would suggest doing it uniformly. I see some places you have:
> 
> ```
> from numpy.typing import NDArray
> import numpy.typing as np_typing
> ```
> 
> and I think we are only using NDArray in the code from this module.

The main reason I swapped the `ndarray` class out with the `NDArray` one is because it makes the type annotations less verbose. Using `np.ndarray` would be preferable if not for the extra verbosity required when using it as an annotation. It's one less import and more familiar to users but what made me consider using `NDArray` was that it only required we a dtype as a a type variable. To provide a fully concrete type using the `ndarray` class, a type for the shape must also be provided and the dtype (np.int64, np.float64, etc...) must be provided as a type variable of the `np.dtype` type. An example of what I mean in code can be found below:

### Example

**Using np.ndarray**
```python
import numpy as np
from typing import Any

def foo() -> np.ndarray[Any, np.dtype[np.int64]]:
    ...
```

**Using np.typing.NDArray**
```python
from numpy.typing import NDArray

def foo() -> NDArray[np.int64]
    ...
```

pyproject.toml
kevinsantana11: Yes, this is to harden the rules for the modules we've already enabled typing for.

clouddrift/sphere.py
kevinsantana11: Here we define bounds to the type variable so while the type is "variable" we can define some bounds to constrain it a bit

pyproject.toml
kevinsantana11: We will still need to work through the modules that still aren't being type checked

clouddrift/sphere.py
KevinShuman: I am curious as to why we have both T and V here when they seem to have the same definition.

clouddrift/sphere.py
```diff
@@ -725,12 +729,12 @@ def cartesian_to_tangentplane(
     return u_projected, v_projected
 
 
+T = TypeVar("T", bound=float | np.ndarray)
```
KevinShuman, line 732: Why is this defined a second time? I see it is also defined in line 667.

clouddrift/ragged.py
```diff
@@ -10,18 +11,25 @@
...
 
 def apply_ragged(
-    func: callable,
-    arrays: list[np.ndarray | xr.DataArray] | np.ndarray | xr.DataArray,
-    rowsize: list[int] | np.ndarray[int] | xr.DataArray,
-    *args: tuple,
+    func: Callable[..., _ArrayOutput],
+    arrays: ArrayTypes,
+    rowsize: ArrayTypes,
+    *args: typing.Any,
```
KevinShuman, line 27: It might help to keep things consistent to also update the docstrings here.

kevinsantana11:
closing as these changes are fairly stale.

PR #522, opened 2024-09-30
🔍 gdp6h adapter ragged array metadata update

kevinsantana11:
Updates the code to generate the correct metadata for the generated ragged array.


kevinsantana11:
@selipot we should try to generate the new zarr archive to update it to S3 so we can also include the updates for the dataset module

selipot:
Working on that but I think many 6 hourly individual netcdf files are repeated in one of the FTP folders

PR #526, opened 2024-10-03
🧹 Refactor boundary conditions in analytic_signal and morse_wavelet_transform functions

selipot:
Changing the default boundary conditions from `mirror` to `periodic` as I believe this handles better signals with non-zero means and is closer to expectations.


kevinsantana11 (approved):
This change looks good to me

Issue #534, opened 2024-12-17
🧹 MyPy flagging depreciated np.chararray


KevinShuman:
Description:
When running MyPy, whether locally or via the test actions, the use of np.chararray in clouddrift/adapters/gdp/__init__.py is being flagged as unrecognized. From what I have gathered, np.chararray has been depreciated in numpy, and we should adjust the cut_str() function to work without it.

1. Clone main
2. create environment with MyPy as suggested in the [contributing page](https://github.com/Cloud-Drift/clouddrift/blob/main/CONTRIBUTING.md)
3. Run the MyPy commands under the "Automatic formatting and linting" section

Expected behavior:
All tests MyPy tests pass.

Desktop:
 - Operating System: Windows, MacOS, Linux
 - `clouddrift` library version v0.43.0

PR #536, opened 2024-12-17
🧹 Updates cut_str() to avoid MyPy flagging depreciated np.chararray Issue #534

KevinShuman:
See issue #534 to see what is being address.

I replace the np.chararray's in the cut_str() function with a numpy array with a fixed length.


kevinsantana11:
Changes look good, feel free to merge

Issue #539, opened 2025-02-04
🐛Outdated Link to Andro Dataset


KevinShuman:
Description:
The link to the Andro dataset is out of date and new version of the dataset needs to be linked.

How to Reproduce It:
1. Import clouddrift: `import clouddrift as cd`
2. Run `cd.datasets.andro() `

Expected behavior:
The data loads.

Desktop:
 - Operating System: Ubuntu
 - `clouddrift` library version: 0.43

KevinShuman:
All was fixed in a pull request #540 .

PR #540, opened 2025-02-04
Updating Link to Andro Library #539

KevinShuman:
This is to resolve issue #539. 

It is as simple as updating the link in the adapter.


KevinShuman:
Looks like I forgot to update a version variable!

kevinsantana11 (approved):
changes look good

Issue #516, opened 2024-08-28
Migrate Jonathan Lilly jLab ridge analysis function to detect eddies


kevinsantana11:

kevinsantana11:
https://www.mathworks.com/matlabcentral/fileexchange/52885-jlab-a-matlab-toolbox-for-data-analysis

PR #544, opened 2025-02-27
⭐ Ridge Analysis Functions for Eddy Detection Issue #516

KevinShuman:
Porting the ridge analysis functions written by [Jonathan Lilly](https://jmlilly.net/code.html) in MatLab to Python to be used in clouddrift.


selipot:
@jonathanlilly we are starting to port your ridge analysis for eddy detection to Clouddrift. I hope you can review and advise.

selipot:
Some comments about the notebook and on improving some of the figures:

- let's calculate the wavelet transform on the analytic and anti-analytic signals, so let's do instead the following, even if it does not change the results much:

```
from clouddrift.signal import analytic_signal
xy_p, xy_n = analytic_signal(x_t+1j*y_t)
wavelet_plus = morse_wavelet_transform(xy_p, gamma, beta, freqs, boundary="mirror")
wavelet_minus = morse_wavelet_transform(xy_n, gamma, beta, freqs, boundary="mirror")
```

- Because the center frequencies of the wavelet as returned by `morse_logspace_freq` are logarithmically spaced, i think using 
```
plt.imshow(np.log10(np.abs(wavelet_plus)), aspect='auto', extent=[t[0], t[-1], freqs[-1], freqs[0]], cmap='jet', vmin=-1, vmax=3.5)
```
creates a wrong representation of the frequency axis because of `extent = ...`. See my suggestion to use `pcolormesh` below.

- Let's scale properly the time and frequency axis to days and cycles per day so maybe defines `freqs_scaled = freqs / (2 * np.pi) / dt` and `t_scaled = t * dt` with `dt = 1/24` for plotting purposes. So I would plot something like:
```
fig, axs = plt.subplots(2, 2, figsize=(12, 6), dpi=150)
X, Y = np.meshgrid(t/24, np.log10(freqs_scaled))
pcm = axs[0,0].pcolormesh(X, Y, np.log10(np.abs(wavelet_plus)), shading='auto', cmap='jet')
axs[0,0].set_title("Morse Wavelet Transform - Plus Transform")
fig.colorbar(pcm, ax=axs[0,0], label='Log10 Power')
axs[0,0].set_ylabel("Log10 Frequency (cpd)")
axs[0,0].set_xlabel("Time (days)")
...
```

For the final figures, I think that plotting `np.log10` of the wavelet amplitude is probably a better idea but I am having trouble subsequently scaling the range of values to plot/contour via the function `wavelet_contourf`.

selipot (commented):
I think it works. @KevinShuman what would be the next steps in your opinion for an eddy detection function using the ridge function?

selipot:
@philippemiron would you be able to test this new module? I am sharing an [example notebook](https://gist.github.com/selipot/770242658023241e4d3c6cca68993e1e) in my Gists.

philippemiron:
Sorry I missed this! I'll check it yes.  @selipot can you share the file `path = "matern_displacement.txt"`?

selipot:
> Sorry I missed this! I'll check it yes. @selipot can you share the file `path = "matern_displacement.txt"`?

[matern_displacement.txt](https://github.com/user-attachments/files/24547708/matern_displacement.txt)

philippemiron:
Looks good! I just have a few comments on the Notebook.

- There is a lot of code, we should maybe include comments to highlight which functions are from `clouddrift`. Or maybe just `import clouddrift` and then we would see `clouddrift.ridges.ridge_analysis` for example.
- It's been a while since we discussed this, but I think I understand the analytical example. In the GDP example (screenshot below), what is the physical meaning of time-overlapping ridges?
- This is not used right?
```
# peak at the some SST data for this drifter
ds.sst1[slice(traj_idx[j], traj_idx[j + 1])]
```

image

selipot:
> Looks good! I just have a few comments on the Notebook.
> 
> * There is a lot of code, we should maybe include comments to highlight which functions are from `clouddrift`. Or maybe just `import clouddrift` and then we would see `clouddrift.ridges.ridge_analysis` for example.

I agree.

> * It's been a while since we discussed this, but I think I understand the analytical example. In the GDP example (screenshot below), what is the physical meaning of time-overlapping ridges?

I think that maybe we could remove the analytical example for this notebook. In the GDP example there can be time-overlapping ridges because these can corresponds to oscillations at distinct frequencies. Just like when a drifter is embedded in a mesoscale eddy and also undergoes inertial oscillations, which is exactly what is happening here between days 200 and ~300. This would be clearer if the ridge analysis is done on cyclonic and anticyclonic motions instead of x and y.

> * This is not used right?
> 
> ```
> # peak at the some SST data for this drifter
> ds.sst1[slice(traj_idx[j], traj_idx[j + 1])]
> ```

Indeed, I did not see this used in the notebook

philippemiron:
I think I would keep the analytical one personally, but maybe you could add this comment/explanation in the notebook at the GDP section.

selipot:
Thanks for this port, @KevinShuman — it is the foundation of everything that follows.

I have continued the work in #610 , which **includes this branch unchanged in its history**, so your commits and authorship are preserved rather than reimplemented. It does not need this pull request to be merged first,
and once it lands the diff here becomes empty, so I am closing #544 as superseded.

What changed on top of your branch, in case it is useful to see:

- `ridge_analysis` now takes a bivariate transform and a `mask`, which is what expresses the
  one-sided joint ridges of Lilly and Pérez-Brunius (2021), and returns the transform sampled
  along each ridge, Eq. (54).
- Three defects found while testing against jLab: the multivariate instantaneous frequency was
  differentiated across components rather than along time; ridge chaining had no continuity
  constraint, so ridges could jump between unrelated oscillations; and the ridge length
  integrated the scale frequency where Eq. (60) uses the instantaneous frequency. Each has a
  regression test.
- The Copilot comment on this pull request is addressed, and the per-ridge arrays are now keyed
  by name rather than position, which is what allowed the length defect in the first place.
- There is now a fixture-based comparison against jLab: with the fixes, the ridge sets match
  `ridgewalk` exactly on identical input.

The notebook comments from @philippemiron and me are not addressed in the new pull request
either; they belong in a `*-get-started` repository under the organization, since `docs/` here
renders no notebooks.

Issue #558, opened 2025-05-23
🧹 Printing temp directory location for GDP1h data


KevinShuman:
While the gdp6h adapter prints out the temporary directory, the gdp1h does not, and we want it to.

PR #559, opened 2025-05-23
Changes logger statement to print statement Issue #558

KevinShuman:
This PR address the want to have a print statement in the gdp1h adapter that prints the temporary directory.


selipot:
Looks like it's doing what it needs to do!

PR #567, opened 2025-06-28
⭐ added various functions to the binning module

### LINKED ISSUE #565: ⭐ binning improvements
To track the future steps for the binning function / module.

- [ ] extending to more statistics than just the current mean [(see, twodstats.m)](https://github.com/jonathanlilly/jLab/blob/master/jStats/twodstats.m)
- [ ] automatic handling of datetime in coordinates/variables
- [ ] think about how we can abstract the "histogram" part so we could use other methods like polynomial fitting

philippemiron:
Created this since I have worked on extending the binning module (closes #565).

- [x] extending to more statistics than just the current mean [(see, twodstats.m)](https://github.com/jonathanlilly/jLab/blob/master/jStats/twodstats.m)
- [x] automatic handling of datetime in coordinates/variables
- [x] add new tests for the other functions
- [x] ~~think about how we can abstract the "histogram" part so we could use other methods like polynomial fitting~~ (another PR)
- [x] fix anonymous function name and name collisons

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "clouddrift"
-version = "0.45.0"
+version = "0.46.0"
 
 authors = [
   { name="Shane Elipot", email="selipot@miami.edu" },


clouddrift/binning.py
```diff
@@ -0,0 +1,519 @@
...
+    bins : int or lists, optional
+        Number of bins or bin edges per dimension. It can be:
+        - An int: same number of bins for all dimensions (default: 10),
+        - A list of ints or arrays: one per dimension, specifying either bin count or bin edges,
+        - None: defaults to 10 bins per dimension.
+    bins_range : list of tuples, optional
+        Outer bin limits for each dimension.
+    statistics : str or list of str, optional
+        Statistics to compute for each bin (default: "mean"). Can be a string or a list of
+        variables to compute. Supported values are 'count', 'sum', 'mean', 'median', 'std', 'min', 'max'.
```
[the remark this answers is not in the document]
philippemiron: Correct, that's a mistake.

clouddrift/binning.py
```diff
@@ -33,6 +288,12 @@ def binned_statistics(
         - None: defaults to 10 bins per dimension.
     bins_range : list of tuples, optional
         Outer bin limits for each dimension.
+    statistics : str or list of str, Callable[[np.ndarray], float] or list[Callable[[np.ndarray], float]]
+        Statistics to compute for each bin. It can be:
+        - a string, supported values: 'count', 'sum', 'mean', 'std', 'min', 'max', (default: "count"),
+        - a list of strings from the same supported values, each specifying a statistic to compute,
+        - a custom function as a callable that take a 1D array of values and return a single value,
```
selipot, line 295: Could the custom function takes more than one argument. As an example, to calculate the covariance matrix of the data inputs?

philippemiron: Yes, but you have to do some tricks.


Either:

```
from functools import partial

def my_func(a, b):
    return a + b

# Create a new callable with 'a' set to 5
partial_func = partial(my_func, 5)

def caller(func):
    return func(10)

result = caller(partial_func)  # Calls my_func(5, 10)
```

or with a lambda:

```

def caller(func):
    return func(10)

result = caller(lambda x: my_func(5, x))  # Also calls my_func(5, 10)

```

philippemiron: I added tests yesterday, see for example to use `np.percentile`. 

```
ds = binned_statistics(
    coords=self.coords_1d,
    data=self.values_1d,
    bins=3,
    statistics=[lambda x: np.percentile(x, q=40), "mean"],
)
```

philippemiron:
> We should probably do a release once you’re done with this PR. Do you think you could create a binning get started notebook that illustrates the various functionalities at the same time? Something simple using various datasets, but that creates a plot we could add to the newsletter promoting the release.

Yes, I can do this too, or maybe we include binning in one of the existing "dataset" getting-started guides? We also have the `clouddrift-examples` repo..

clouddrift/binning.py
selipot: > I think I didn't understand the question correctly. Are you asking if it's possible to use multiple data variables in the same function?
> 
> If so, right now it isn't, but I'm thinking we could do something like this:
> 
> ```
> ds = binned_statistics(
>     coords=self.coords_1d,
>     data=[self.values_1d, self.values_1d],
>     bins=10,
>     statistics=([0, 1], lambda data: np.sqrt(np.mean(data[0]**2 + data[1]**2))),
> )
> ```
> 
> where the tuple would define the index of the data variables needed, and the function. It does add a bit of complexity to call the function, but would make it versatile...

Yes exactly, to get the KE as an example.

clouddrift/binning.py
```diff
@@ -1,18 +1,271 @@
...
+        bin_counts = _binned_count(flat_idx, n_bins)
+
+    if bin_mean is None:
+        bin_mean = _binned_mean(flat_idx, n_bins, values, bin_counts)
+
+    bin_sumsq = np.bincount(flat_idx, weights=values**2, minlength=n_bins)
+    bin_mean_sq = np.divide(
+        bin_sumsq, bin_counts, out=np.full(n_bins, 0.0), where=bin_counts > 0
+    )
+    variance = np.maximum(bin_mean_sq - bin_mean**2, 0)
```
selipot, line 155: That's the variance? Does it work for complex quantities? in which case the variance is `np.mean(np.conj(xp)*xp)` where `xp = flat_idx - bin_mean_sq`, I think.

selipot:
> > We should probably do a release once you’re done with this PR. Do you think you could create a binning get started notebook that illustrates the various functionalities at the same time? Something simple using various datasets, but that creates a plot we could add to the newsletter promoting the release.
> 
> Yes, I can do this too, or maybe we include binning in one of the existing "dataset" getting-started guides? We also have the `clouddrift-examples` repo.. What do you think?

Whatever you think is best. I like the idea of a notebook for a new module though.

clouddrift/binning.py
```diff
@@ -131,40 +425,70 @@ def binned_statistics(
...
-        # and statistic when data is provided
-        if var.size:
-            ds[name] = xr.DataArray(
-                mean.reshape(edges_sz),
+            if zeros_to_nan and np.any(binned_stats == 0):
+                binned_stats = np.where(binned_stats == 0, np.nan, binned_stats)
+
+            # and variable to the Dataset
+            variable_suffix = (
+                _get_function_name(statistic) if callable(statistic) else statistic
```
selipot, line 488: passing the suffix "anon_lambda_func" is a bit much. Why "anon"? Maybe the suffix could be "stat0", "stat1", etc based on the number of lambda functions passed in the `statistics` input argument?

selipot:
I tried all of this with my local version of the 6hourly ragged datatset and it works great. I made a comment/suggestion inline for the name of the outputs when using a lambda function.

```
import xarray as xr
import numpy as np

# filein = '/Users/selipot/Data/awss3/latest/gdp-v2.01.1.zarr'
# ds = xr.open_zarr(filein, consolidated=True)

ds = xr.open_zarr('/Users/selipot/Data/drifters/process/gdp6h_ragged_jul24.zarr')

# %%
from clouddrift.binning import binned_statistics
from scipy.stats import skew

coords = [ds.lon, ds.lat]
variables = [ds.ve, ds.vn, ds.temp]

ds_binned = binned_statistics(
    coords=coords,
    data=variables,
    bins=[360,180],
    bins_range=[(-180, 180), (-90, 90)],
    dim_names=["lon_bin", "lat_bin"],
    output_names=["ve", "vn", "temp"],
    statistics=["mean", "std", "count"],
    zeros_to_nan=True,
)

ds_median = binned_statistics(
    coords=coords,
    data=variables,
    bins=[360,180],
    bins_range=[(-180, 180), (-90, 90)],
    dim_names=["lon_bin", "lat_bin"],
    output_names=["ve", "vn", "temp"],
    statistics= lambda x: np.percentile(x, q=50),
    zeros_to_nan=True,
)

ds_skew = binned_statistics(
    coords=coords,
    data=variables,
    bins=[360,180],
    bins_range=[(-180, 180), (-90, 90)],
    dim_names=["lon_bin", "lat_bin"],
    output_names=["ve", "vn", "temp"],
    statistics= lambda x: skew(x, axis=None, nan_policy='omit'),
    zeros_to_nan=True,
)

```

clouddrift/binning.py
philippemiron: Anon is for Anonymous lambda. I could try to figure out a clever way to add a counter. Right now the function name was retrieved only when adding the results to the dataset.

philippemiron:
The code is ugly, but it is working. I don't think I'll have more time over the weekend, so I will go back to try to clean this up next week.

The idea is that for multivariable `Callable`, you have to pass a `tuple`,  `(output_name, Callable)`, where the `output_name` is the variable resulting from a combination of any `data = variables`. The `Callable` will receive the complete list, so for example in your case where you pass `variables = [ds.ve, ds.vn, ds.temp]`, if you want to calculate KE from `ve` and `vn`, the `lambda` will reference `data[0]` (`ve`) and `data[1]` (`vn`).


This should work:

```
ds_binned = binned_statistics(
    coords=coords,
    data=variables,
    bins=[360, 180],
    bins_range=[(-180, 180), (-90, 90)],
    dim_names=["lon_bin", "lat_bin"],
    output_names=["ve", "vn", "temp"],
    statistics=[
        "mean",
        "std",
        "count",
        "median",
        lambda x: skew(x, axis=None, nan_policy="omit"),
        ("ke", lambda data: np.sqrt(np.mean(data[0] ** 2 + data[1] ** 2))),
    ],
    zeros_to_nan=True,
)

```

clouddrift/binning.py
selipot: We should not have to do a different version for real and complex. When the input is real, the formula I gave above is still correct. We might need to make sure the returned result is always real. But is the variance really `np.maximum(bin_mean_sq - bin_mean**2, 0)`? I don't get it.

clouddrift/binning.py
```diff
@@ -1,19 +1,315 @@
...
 def binned_statistics(
     coords: np.ndarray | list[np.ndarray],
     data: np.ndarray | list[np.ndarray] | None = None,
     bins: int | list = DEFAULT_BINS_NUMBER,
     bins_range: list | None = None,
     dim_names: list[str] | None = None,
     output_names: list[str] | None = None,
+    statistics: str | list | Callable[[np.ndarray], float] = "count",
     zeros_to_nan: bool = False,
+    fill_value: float = np.nan,
```
selipot, line 312: What is this for? Not described in docstrings.

clouddrift/binning.py
```diff
@@ -33,12 +329,21 @@ def binned_statistics(
         - None: defaults to 10 bins per dimension.
     bins_range : list of tuples, optional
         Outer bin limits for each dimension.
+    statistics : str or list of str, Callable[[np.ndarray], float] or list[Callable[[np.ndarray], float]]
+        Statistics to compute for each bin. It can be:
+        - a string, supported values: 'count', 'sum', 'mean', 'median', 'std', 'min', 'max', (default: "count"),
+        - a custom function as a callable that take a 1D array of values and return a single value.
```
selipot, line 335: a custom function as a callable for univariate statistics that take a 1D array of values and return a single value.

clouddrift/binning.py
```diff
@@ -33,12 +329,21 @@ def binned_statistics(
...
+          The callable is applied at each variable.
+        - a tuple of (var_name, callable) for multivariate statistics. 'var_name' is used to identify the resulting variable.
+          In this case, the callable will receive the list of arrays provided in `data`. For example, to calculate kinetic energy,
+          you can pass `data = [u, v]` and  `statistics=("ke", lambda data: np.sqrt(np.mean(data[0] ** 2 + data[1] ** 2)))`.
+        - a list containing any combination of the above, e.g., ['mean', np.nanmax, ('ke', np.sqrt(np.mean(data[0] ** 2 + data[1] ** 2)))].
```
selipot, line 340: Do you mean:  - a list containing any combination of the above, e.g., ['mean', np.nanmax, ('ke', lambda data: np.sqrt(np.mean(data[0] ** 2 + data[1] ** 2)))].

clouddrift/binning.py
```diff
@@ -33,12 +329,21 @@ def binned_statistics(
         - None: defaults to 10 bins per dimension.
     bins_range : list of tuples, optional
         Outer bin limits for each dimension.
+    statistics : str or list of str, Callable[[np.ndarray], float] or list[Callable[[np.ndarray], float]]
+        Statistics to compute for each bin. It can be:
+        - a string, supported values: 'count', 'sum', 'mean', 'median', 'std', 'min', 'max', (default: "count"),
+        - a custom function as a callable that take a 1D array of values and return a single value.
+          The callable is applied at each variable.
+        - a tuple of (var_name, callable) for multivariate statistics. 'var_name' is used to identify the resulting variable.
```
selipot, line 337: For consistency, should you use (output_name, callable) instead of (var_name, callable)?

clouddrift/binning.py
```diff
@@ -83,6 +388,56 @@ def binned_statistics(
             for r, c in zip(bins_range, coords)
         ]
 
+    # validate statistics parameter
+    ordered_statistics = ["count", "sum", "mean", "median", "std", "min", "max"]
+    if isinstance(statistics, (str, tuple)) or callable(statistics):
+        statistics = [statistics]
+    elif not isinstance(statistics, list):
+        raise ValueError(
+            "'statistics' must be a string, list of strings, callable, or a list of callables. "
```
selipot, line 397: use backticks around statistics

clouddrift/binning.py
```diff
@@ -66,7 +371,7 @@ def binned_statistics(
     # set default bins and bins range
     if isinstance(bins, (list, tuple)):
         if len(bins) != len(coords):
-            raise ValueError("bins must match the dimensions of the coordinates")
+            raise ValueError("'bins' must match the dimensions of the coordinates")
```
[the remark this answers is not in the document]
philippemiron: I thought we used 'vars' in warning/error messages. I will validate.

clouddrift/binning.py
philippemiron: I don't think it's exactly the variance... There is a bincount that reduce everything to the sum per bin.

clouddrift/binning.py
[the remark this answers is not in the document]
philippemiron: For example here: https://github.com/Cloud-Drift/clouddrift/blob/main/clouddrift/kinematics.py. There are single and double backtick in doctring but nothing in the assertion error messages. But I see you did it like this in `transfer.py`.

clouddrift/binning.py
philippemiron: removed

clouddrift/binning.py
philippemiron: Fixed.

clouddrift/binning.py
selipot: Oh I think I get it. I had to ask copilot to explain to me: This line calculates the variance for each bin using the formula variance = E[x²] - (E[x])², where E[x²] is the mean of the squared values (bin_mean_sq) and E[x] is the mean (bin_mean). This is a standard approach for computing variance from summary statistics.

The use of np.maximum(..., 0) ensures that the variance is never negative.

philippemiron:
> @philippemiron I am trying to add support for complex variables. I implemented something with the aid of Copilot. This can be rewritten or reverted easily I believe? Next steps:
> 
> 
> 
> - [ ] add specific test for complex inputs?
> 
> - [ ] add a Raise Error in the wrapper function when the input is complex for min, max, and median? I added Raise Errors in the `_binned_min` and `_binned_max` but it would make it more efficient to raise an error earlier. For the 'median' statistic it would work but it would disregard the complex parts which I think is not a behavior which is desirable.
> 
> 
> 
> 

I took a quick look and it looks good! Yeah, I'm not sure what would be the ideal behavior for `median`, `min`, `max`. I still have to do the time conversion...

tests/binning_test.py
```diff
@@ -267,23 +277,353 @@ def test_rename_variables(self):
...
+        self.assertFalse(np.iscomplexobj(ds["binned_0_std"].values))
+        self.assertTrue(np.issubdtype(ds["binned_0_std"].dtype, np.floating))
+        # Count should be real and integer
+        self.assertTrue(np.issubdtype(ds["binned_0_count"].dtype, np.integer))
+        # Check that the sum is the sum of values in each bin
+        self.assertAlmostEqual(np.sum(ds["binned_0_sum"].values), np.sum(values))
+        # Check that the mean is the mean of values in each bin
+        mask0 = (coords >= -0.2) & (coords < 0.9)
+        mask1 = (coords >= 0.9) & (coords < 2.0)
+        mask2 = (coords >= 2.0) & (coords <= 3.1)
```
selipot, line 541: It seems illogical to me that a data value located exactly on the right edge of the last bin would be included in the calculation of the given statistic since all the other right edges are excluded in the other intervals.

philippemiron: I see the point, but if you have 10 data points and don't put any `bins_range`, it seems illogical that you would end up with 9 binned data points by default.

philippemiron: Not sure how they do it.. I will take a look at the code later. But see this example,

```
import numpy as np
from scipy import stats

# 1D array
arr = [1, 2, 9.25, 20, 34]

ret = stats.binned_statistic(arr, np.arange(5), statistic="count", bins=4)
# median
print(f"\nbinned_statistic for count : {ret}\n")

for e in ret.bin_edges:
    print(e)
```

```

binned_statistic for count : BinnedStatisticResult(statistic=array([2., 1., 1., 1.]), bin_edges=array([ 1.  ,  9.25, 17.5 , 25.75, 34.  ]), binnumber=array([1, 1, 2, 3, 4]))

1.0
9.25
17.5
25.75
34.0
```
where the third point (9.25) is in the second bin [9.25, 17.5], but the last point (34) is included in the last bin [25.75, 34].

philippemiron: See this function:

https://github.com/scipy/scipy/blob/main/scipy/stats/_binned_statistic.py#L766

```
    # Using `digitize`, values that fall on an edge are put in the right bin.
    # For the rightmost bin, we want values equal to the right
    # edge to be counted in the last bin, and not as an outlier.
```

philippemiron: I moved the last bins by eps, what they do is they keep the edges value with:

```
bins_range = [
        (np.nanmin(c), np.nanmax(c)) for c in coords
    ]
```

but then have an extra step to search for points that would be directly on `np.nanmax(c)` to include them on the last bin.

philippemiron:
@selipot for the datetime automatic conversion: 
- does it make sense to have `datetime` in the `data` too?
- or do we only support datetime in the `coords`? This would allow to do temporal binning.

The first one is more complex, for example: if convert datetime to float of `s` since 1970, do a `mean`, you can convert back to datetime. If you do `std` instead, then the results can't directly be converted back to datetime, you have to also somehow store the mean value.

philippemiron:
Sorry, I didn't realize this. Will try fixing right now.

clouddrift/binning.py
```diff
@@ -26,21 +479,29 @@ def binned_statistics(
...
         Data values associated with the Lagrangian coordinates in coords.
         Can be a single array or a list of arrays for multiple variables.
+        Complex values are supported for the supported statistics except for 'min', 'max', and 'median'.
     bins : int or lists, optional
         Number of bins or bin edges per dimension. It can be:
         - An int: same number of bins for all dimensions (default: 10),
         - A list of ints or arrays: one per dimension, specifying either bin count or bin edges,
         - None: defaults to 10 bins per dimension.
     bins_range : list of tuples, optional
         Outer bin limits for each dimension.
```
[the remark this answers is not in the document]
philippemiron: You can provide range for a subset. I need to add tests to this actually. If you do this: [[-90, 90], None], it *should* apply the range only on the first variable and take `min/max` for the second one.

clouddrift/binning.py
```diff
@@ -1,19 +1,472 @@
...
+        1D array of data values
+    bin_counts : array-like, optional
+        Precomputed counts per bin. If None, it will be computed using `_binned_count`.
+    bin_sum : array-like, optional
+        Precomputed sum per bin. If None, it will be computed using `_binned_sum`.
+
+    Returns
+    -------
+    result : array-like
+        1D array of length n_bins with the mean per bin
```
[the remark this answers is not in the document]
philippemiron: Good question :D, I see I have not been consistent with that.

selipot:
> Hi @selipot,
> 
> In the end I don't think anything is wrong, just maybe confusing. If you use the `tuple` in `statistics`, this is for the multivariate statistics. In that situation, the `x` will be a list containing all the `data` variables. So you would have to do something like this:
> 
> ```
> coords = [ds.lon, ds.lat]
> variables = [ds.ve, ds.vn, ds.temp]
> 
> ds_stats = binned_statistics(
>     coords=coords,
>     data=variables,
>     bins=[360,180],
>     bins_range=[(-180, 180), (-90, 90)],
>     dim_names=["lon", "lat"],
>     output_names=["ve", "vn", "temp"],
>     statistics=['mean',
>         ('ve_q25', lambda x: np.percentile(x[0], q=2.5)),
>         ('ve_q975', lambda x: np.percentile(x[0], q=97.5)),
>         ('ve_skew', lambda x: skew(x[0], axis=None, nan_policy='omit')),
>         ('vn_q25', lambda x: np.percentile(x[1], q=2.5)),
>         ('vn_q975', lambda x: np.percentile(x[1], q=97.5)),
>         ('vn_skew', lambda x: skew(x[1], axis=None, nan_policy='omit')),
>         ('temp_q25', lambda x: np.percentile(x[2], q=2.5)),
>         ('temp_q975', lambda x: np.percentile(x[2], q=97.5)),
>         ('temp_skew', lambda x: skew(x[2], axis=None, nan_policy='omit')),
>     ],
> )
> ```
> 
> but here if you want to apply the same function to all variables, you just pass a `Callable` instead of a `tuple(name, Callable)`.
> 
> ```
> coords = [ds.lon, ds.lat]
> variables = [ds.ve, ds.vn, ds.temp]
> 
> ds_stats = binned_statistics(
>     coords=coords,
>     data=variables,
>     bins=[360,180],
>     bins_range=[(-180, 180), (-90, 90)],
>     dim_names=["lon", "lat"],
>     output_names=["ve", "vn", "temp"],
>     statistics=['mean',
>         lambda x: np.percentile(x[0], q=2.5),
>         lambda x: np.percentile(x[0], q=97.5),
>         lambda x: skew(x[0], axis=None, nan_policy='omit')
>     ],
> )
> ```
> 
> because it is an anonymous function, the variable names will be `ve_stat_0`, `vn_stat_0`, `ve_stat_1`, etc.
> 
> Let me know if you have suggestions on how to make this clearer.

Ok, in the second option you propose, I do not think the Callable should be `lambda x: np.percentile(x[0], q=2.5)` but should be `lambda x: np.percentile(x, q=2.5)`. Is this correct? Otherwise you are passing only the first element to the computation?

I don't want to ask tooooo much but I think the output I originally expected would be nicer :) or makes more sense, i.e. get `ve_q25`, `vn_q25`, etc.

PR #573, opened 2025-08-13
⭐️ updated datasets.gdp6h() docstring

samouertani:
I've updated the docstring for the gdp6h() function, to reflect the new December 2024 update

### DIFF
--- a/clouddrift/datasets.py
+++ b/clouddrift/datasets.py
@@ -93,8 +93,8 @@ def gdp6h(decode_times: bool = True) -> xr.Dataset:
     Oceanographic and Meteorological Laboratory (AOML) accessible at
     https://www.aoml.noaa.gov/phod/gdp/index.php.
 
-    This returns the July 2024 version of the dataset including data from 1979-02-15:00:00:00Z
-    to 2024-08-16:12:00:00Z.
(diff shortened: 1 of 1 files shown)


philippemiron:
@samouertani,

I see that by loading the file that was uploaded:

```
url = "https://noaa-oar-hourly-gdp-pds.s3.amazonaws.com/experimental/gdp6h_ragged_dec24.zarr"
ds = xr.open_dataset(url, decode_times=False, engine="zarr")
ds.history

> 'version December 2024. Metadata from dirall.dat and deplog.dat'
```
you modified the date properly. But I think we also have to modify the following [line](https://github.com/Cloud-Drift/clouddrift/blob/main/clouddrift/adapters/gdp/gdp6h.py#L21C1-L21C26) so the correct date is applied if someone runs the code again.

selipot:
In addition, I am noticing that in the example in the docstring, the property `time_coverage_end` shows:
`time_coverage_end:    2024-08-16:12:00:00Z`. Should it not show something like `2024-12-xx`?

In terms of the PR, I also notice that you used the "main" branch of your forked clouddrift repo. Typically one would create a new branch under your fork (origin repo) that you then request to merge to the main of the upstream repo.

philippemiron:
Closing since other opened PR.

PR #577, opened 2025-08-20
🐛 fix url for glad dataset

philippemiron:
- update glad GRIIDC URL
- add header to mimic browsers
- increase the timeout


philippemiron:
Good to go!

selipot (approved):
I see the modifications, and they seem to do the trick! Thank you for spending the time to fix this!

PR #590, opened 2026-04-07
🐛 add warning for short data length in inertial_oscillation_from_position and implement corresponding tests

selipot:
the function `inertial_oscillation from_position` would fail if the length of the data was too short to capture the inertial frequency. A new test verify that the Rayleigh frequency implied by the length of the time series is smaller than the range defined by 95% of the lowest inertial frequency and 105% of the highest inertial frequency defined by the range of latitudes. If not the function should return a warning and output NaN values without producing an error.


clouddrift/kinematics.py
```diff
@@ -76,6 +78,10 @@ def inertial_oscillation_from_position(
     potentially shifted in frequency by a measure of relative vorticity. The result is a pair
     of zonal and meridional relative displacements in meters.
 
+    If the length of the data is too short to capture the inertial frequency, as implied by the
+    time step and range of latitudes of the data, the function will raise a warning and return arrays
+    of NaNs.
```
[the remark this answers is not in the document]
selipot: @copilot apply changes based on [this feedback](https://github.com/Cloud-Drift/clouddrift/pull/590#discussion_r3047521022)

clouddrift/kinematics.py
```diff
@@ -186,6 +231,26 @@ def inertial_oscillation_from_position(
...
+    # as implied by the time step and range of latitudes of the data; if not,
+    # issue a warning and return arrays of NaNs because the wavelet transform will
+    # not be able to capture the inertial oscillations
+    positive_cor_freq = cor_freq[cor_freq > 0]
+    if data_length < 2 * np.pi / (time_step * np.min(positive_cor_freq) * 0.95):
+        warnings.warn(
+            "The length of the data is too short to capture the inertial frequency",
+            UserWarning,
+        )
+        return np.full_like(latitude, np.nan, dtype=float), np.full_like(
```
philippemiron, line 245: I'm wondering if it's not better to return an exception instead of returning an array of nan in that case. Wha t do you think?

selipot: @philippemiron what does this mean concretely? I did this because using `apply_ragged` with this function basically broke when one of the trajectories was too short. I implemented this so that the batch processing of `apply_ragged` worked and returned NaN for the too short trajectories. Hope this makes sense.

philippemiron: I didn't think about wrapping this function in `apply_ragged`! All good, I understand why returning an array of `nan` now.

philippemiron:
@selipot I merged the main branch, so it should fix the tests now.

selipot:
Thanks but oops @philippemiron I was trying to do the same thing at the same time, updating `ANDRO_VERSION = "2026-01"` but to `ANDRO_VERSION = "2026-04"`. Now I think I need to merge the main again in my PR.

philippemiron:
I think they only update the dataset once a year. That's why it was like this. But if they write the version as `2026-04`, we can update it.

PR #591, opened 2026-04-17
update ANDRO_URL to point to the latest dataset version

selipot:


clouddrift/adapters/andro.py
```diff
@@ -29,7 +29,7 @@
 from clouddrift.adapters.utils import download_with_progress
 
 # order of the URLs is important
-ANDRO_URL = "https://www.seanoe.org/data/00360/47077/data/116520.dat"
+ANDRO_URL = "https://www.seanoe.org/data/00360/47077/data/127690.zip"
 ANDRO_TMP_PATH = os.path.join(tempfile.gettempdir(), "clouddrift", "andro")
 ANDRO_VERSION = "2025-01"
```
philippemiron, line 34: this should be update to `2026-01`.

philippemiron: and this `https://github.com/selipot/clouddrift/blob/f032568a729266096cc76acabd2b41a2dc814507/clouddrift/datasets.py#L739`, also.

selipot: argh, went too fast

PR #597, opened 2026-04-19
✨ Add `raw` and `qc1` level to the CARTHE glad dataset

philippemiron:


selipot:
Thank you for doing this, that's exciting. A couple of points.

- Copilot has a point to not change `glad()` to `carthe_glad()`  because it breaks backward compatibility and the notebook. I personally prefer for the name of the dataset to be the name of the experiment which was GLAD. 
- Should we add in the docstrings more details about the qc level? What do they mean etc.
- The doi and url in the docstrings (https://doi.org/10.7266/N7VD6WC8) only leads to one version of the data. Is there another doi/link for the deifferent qc/level dataset?

philippemiron:
- I saw copilot's comment but my idea was to make it a bit easier to find group's experiments, since I'll be adding laser, gomecc3, and splash after also. Not sure what's the best strategy.

- I'll add more details and publication/doi (if there are) in the docstring.

selipot:
> * I saw copilot's comment but my idea was to make it a bit easier to find group's experiments, since I'll be adding laser, gomecc3, and splash after also. Not sure what's the best strategy.
> * I'll add more details and publication/doi (if there are) in the docstring.

CARTHE seems to be more of a research group (http://carthe.org) than an experiment, right? Keeping with our convention so far I would therefore use GLAD which is the name of the experiment, even if it was funded through CARTHE. LASER was another experiment etc. Unless I misunderstand?

philippemiron:
Yes a research group. All the experiments I enumerated above are from the same group.

I wanted to have some types of structure:

carthe
- glad
- laser
- splash

but it's fine we can keep like this.

Issue #595, opened 2026-04-17
⭐ Cape Basin CARTHE drifter datasets and LASER Experiment dataset


selipot:
Add the following datasets to the clouddrift library as suggested by @guillaumenovelli

Cape Basin CARTHE drifter dataset: https://doi.org/10.25921/9m27-m532 also available from zenodo https://zenodo.org/records/14902851 This is similar to GLAD but the raw data is every 10 minutes, and the clean data is every 30 minutes.

The LASER experiment (in the GoM, very similar to GLAD but with1000 drifters) CARTHE drifter dataset: https://data.griidc.org/data/R4.x265.237:0001

PR #601, opened 2026-05-09
✨ Add LASER dataset adapter and corresponding dataset and test functions

selipot:
New PR to add the LASER dataset. This addresses issue #595 https://github.com/Cloud-Drift/clouddrift/issues/595


selipot:
Only some data adapters are failing because of upstream issues I believe. So merging!

PR #603, opened 2026-05-24
⭐️ Refactor adapters to return RaggedArray instances

selipot:
- Updated `subsurface_floats`, `yomaha`, `glad`, and `mosaic` adapters to convert datasets to RaggedArray instead of xarray Dataset.
- Modified the `to_xarray` methods to return RaggedArray instances and adjusted the associated tests accordingly.
- Added new unit tests for `glad`, `mosaic`, `subsurface_floats`, and `yomaha` to ensure correct functionality of the new RaggedArray implementation.
- Updated dataset retrieval functions in `datasets.py` to accommodate the changes in adapter return types.

### DIFF
--- a/clouddrift/datasets.py
+++ b/clouddrift/datasets.py
@@ -565,7 +565,9 @@ def mosaic(decode_times: bool = True) -> xr.Dataset:
         rowsize                     (traj) int64 ...
     """
     return _dataset_filecache(
-        "mosaic.nc", decode_times, lambda: adapters.mosaic.to_raggedarray().to_xarray()
+        "mosaic.nc",
+        decode_times,
+        lambda: adapters.mosaic.to_raggedarray().to_xarray(),
     )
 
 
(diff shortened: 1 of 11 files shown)


clouddrift/datasets.py
```diff
@@ -766,7 +772,9 @@ def yomaha(decode_times: bool = True) -> xr.Dataset:
     assessed  from trajectories of Argo floats at parking level and at the sea
     surface. IPRC Technical Note, 4(2), 1-16.
     """
-    return _dataset_filecache("yomaha.nc", decode_times, adapters.yomaha.to_xarray)
+    return _dataset_filecache(
+        "yomaha.nc", decode_times, lambda: adapters.yomaha.to_raggedarray().to_xarray()
```
philippemiron, line 776: I think here, if you remove the last `()` you wouldn't have to have a `lambda:`.

philippemiron (approved):
Only small fixes in dataset to remove the need of the lambda.

clouddrift/datasets.py
selipot: copilot doesn't think so

PR #608, opened 2026-06-09
🔧 fix pkg following conda issues

philippemiron:
Modification after updating the `conda` feedstock (https://github.com/conda-forge/clouddrift-feedstock/pull/55).

- add `license-file` to pypi package
- `conda` would not resolved the new packages for python 3.10. I think we are good dropping support and going with `python >= 3.11` (it's already 4 years old).
- update the environment in the repo
- update documentation


selipot (approved):
Not sure I have the expertise to review. I understand it fixes some CI stuff and now requires python 3.11 and higher. Seems fine to me!

